PR #6, opened 2023-02-20
Expose additional api's for consumption

birkskyum:
In order to hide the internal structure of this package, this exposes the functions needed in maplibre-gl-js, and bumps the verison with a prerelease

### DIFF
--- a/src/style-spec.ts
+++ b/src/style-spec.ts
@@ -89,7 +89,7 @@ import {IMercatorCoordinate, ICanonicalTileID, ILngLat, ILngLatLike} from './til
 import EvaluationContext from './expression/evaluation_context';
 import {FormattedType, NullType, Type, toString} from './expression/types';
 
-import interpolates, {number, color, array, padding, interpolateFactory} from './util/interpolate';
+import interpolates, {interpolateFactory} from './util/interpolate';
 import expressions from './expression/definitions';
 import Interpolate from './expression/definitions/interpolate';
 import type {InterpolationType} from './expression/definitions/interpolate';
--- a/src/style-spec.ts
+++ b/src/style-spec.ts
@@ -69,7 +69,7 @@ import latest from './reference/latest';
 import format from './format';
 import migrate from './migrate';
 import derefLayers from './deref';
-import diff, {operations as diffOperations} from './diff';
+import diff, {operations} from './diff';
 import ValidationError from './error/validation_error';
 import ParsingError from './error/parsing_error';
 import {FeatureState, StyleExpression, isExpression, createExpression, createPropertyExpression, normalizePropertyExpression, ZoomConstantExpression, ZoomDependentExpression, StylePropertyFunction, Feature, GlobalProperties, SourceExpression, CompositeExpression, StylePropertyExpression} from './expression';
@@ -85,7 +85,7 @@ import {eachSource, eachLayer, eachProperty} from './visit';
 import ResolvedImage from './expression/types/resolved_image';
 import validate from './validate_style';
 import {supportsPropertyExpression} from './util/properties';
-import {number} from './util/interpolate';
+import {range} from './util/interpolate';
 import {IMercatorCoordinate, ICanonicalTileID, ILngLat, ILngLatLike} from './tiles_and_coordinates';
 import EvaluationContext from './expression/evaluation_context';
 import {FormattedType, NullType, Type, toString} from './expression/types';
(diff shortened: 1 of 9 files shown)


src/style-spec.ts
```diff
@@ -69,20 +69,40 @@ import latest from './reference/latest';
...
 import Padding from './util/padding';
+import Formatted, {FormattedSection} from './expression/types/formatted';
 import {createFunction, isFunction} from './function';
 import convertFunction from './function/convert';
 import {eachSource, eachLayer, eachProperty} from './visit';
-
+import ResolvedImage from './expression/types/resolved_image';
 import validate from './validate_style';
+import {supportsPropertyExpression} from './util/properties';
+import {number} from './util/interpolate';
```
HarelM, line 88: I think `number` is a reserved word...?
Can we call this class/method differently?

src/style-spec.ts
```diff
@@ -69,20 +69,40 @@ import latest from './reference/latest';
...
+import {IMercatorCoordinate, ICanonicalTileID, ILngLat, ILngLatLike} from './tiles_and_coordinates';
+import EvaluationContext from './expression/evaluation_context';
+import {FormattedType, NullType, Type, toString} from './expression/types';
+
+import * as interpolate from './util/interpolate';
```
HarelM, line 93: Can we change this to something more specific? I don't like using `import * as...`...

HarelM:
Note that there are some jsdocs warnings that needs to be fixed according to github.
The fact that we now know which classes/methods are exported in order to be used in maplibre-gl-js is great as we'll probably need those if we ever get to change the implementation of the style-spec from this v8 to json-scheme or anything else.
Great progress!

src/style-spec.ts
birkskyum: It's used like this - not clear to me how to resolve it... 

```ts
const interp: ((a: T, b: T, t: number) => T) = (interpolate as any)[this.specification.type];
```

src/style-spec.ts
birkskyum: I have renamed it to range()

src/style-spec.ts
```diff
@@ -106,19 +126,63 @@ const visit = {eachSource, eachLayer, eachProperty};
 
 export {
     v8,
+    number,
+    interpolate,
+    Interpolate,
+    InterpolationType,
     latest,
     format,
+    emptyStyle,
```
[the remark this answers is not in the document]
birkskyum: I've sorted in types / functions / const etc. for now

src/style-spec.ts
```diff
@@ -69,20 +69,40 @@ import latest from './reference/latest';
 import format from './format';
 import migrate from './migrate';
 import derefLayers from './deref';
-import diff from './diff';
+import diff, {operations as diffOperations} from './diff';
```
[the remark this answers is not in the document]
birkskyum: okay, the operations is exported as it instead, and and then the "as" can be applied in maplibre-gl-js if needed.

src/style-spec.ts
birkskyum: Reminds me of the other "generic" tricky case which also doesn't have a good solution yet:

https://github.com/maplibre/maplibre-gl-js/blob/045f1ad208f646c80058312887c92608bdda2dbd/build/generate-style-code.ts#L194-L200

birkskyum: Good idea. I've implemented a simple factory like that

src/style-spec.ts
```diff
@@ -69,31 +69,59 @@ import latest from './reference/latest';
...
+import groupByLayout from './group_by_layout';
+import emptyStyle from './empty';
+import validateStyleMin from './validate_style.min';
+import Step from './expression/definitions/step';
+import {typeOf} from './expression/values';
+import FormatExpression from './expression/definitions/format';
+import Literal from './expression/definitions/literal';
+import CompoundExpression from './expression/compound_expression';
+
+const interpolateFactory = (interpolationType: 'range'|'color'|'array'|'padding') => {
```
HarelM, line 106: Move this to the interpolate file please...

birkskyum:
I believe everything but this situation in generate-style-code is in place

https://github.com/maplibre/maplibre-gl-js/blob/045f1ad208f646c80058312887c92608bdda2dbd/build/generate-style-code.ts#L194-L200

HarelM (approved):

PR #7, opened 2023-02-22
Minify style spec

birkskyum:
The minification of the style spec should happen when the style-spec build is made, and not inside of maplibre-gl-js.

This takes ~ 100 kb off the bundle, from ~ 370 kb to ~ 270 kb for the .mjs


HarelM (approved):
LGTM. please check that the removed values/fields are not used in maplibre-gl-js code base

birkskyum:
Yes, if anything they are probably used in the docs generation

PR #11, opened 2023-02-23
node-resolve, preferBuiltins: true,

birkskyum:
Regarding the 'url' module, preferBuiltins has to be true


HarelM (approved):
Can you remove this line and see if this passes the build?
https://github.com/maplibre/maplibre-gl-style-spec/blob/14cf309d8eac5c86c843942a76c4f1878cf5071f/src/migrate/v8.ts#L2
use `new URL(...)` instead maybe?

HarelM (approved):

PR #17, opened 2023-02-23
Fix imports in binaries

birkskyum:
Allow the binaries to make imports in the exact files

### DIFF
--- a/bin/gl-style-format.js
+++ b/bin/gl-style-format.js
@@ -2,7 +2,7 @@
 
 import fs from 'fs';
 import minimist from 'minimist';
-import {format} from '@maplibre/maplibre-gl-style-spec';
+import format from '@maplibre/maplibre-gl-style-spec/tsc/src/format';
 const argv = minimist(process.argv.slice(2));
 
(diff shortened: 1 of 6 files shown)


bin/gl-style-format.js
```diff
@@ -2,7 +2,7 @@
 
 import fs from 'fs';
 import minimist from 'minimist';
-import {format} from '@maplibre/maplibre-gl-style-spec';
+import format from '@maplibre/maplibre-gl-style-spec/tsc/src/format';
```
HarelM, line 5: This assumes there's a tsc output folder in rollup config, this isn't a great assumption.
I would consider restructuring these files into a single file, written in typescript and exported using rollup as an alternative entry point for this module.

birkskyum: Good idea, I tried converting them to ts previously, because the static typing is helpful, but the repo just isn't made for it. I need a break for this task even though it's quite far - feel free to continue.

HarelM: I've opened a ticket to track this.

PR #19, opened 2023-02-23
separate-rollup-for-binaries

### LINKED ISSUE #18: Migrate bin scripts to use typescript
current bin scripts are in javascript, but they are importing typescript files.
This isn't ideal as the whole repo is using typescript and these scripts should too.
It might be also easier to create a single entry point script with the relevant command line arguments to support all these scripts.
This will probably require another rollup target for this.

birkskyum:
Fixes #18 , converting binaries to typescript and give them separate rollup bundles

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@maplibre/maplibre-gl-style-spec",
   "description": "a specification for maplibre gl styles",
-  "version": "18.0.1-pre.12",
+  "version": "18.0.1-pre.13",
   "author": "MapLibre",
   "keywords": [
     "mapbox",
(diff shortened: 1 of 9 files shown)


rollup.config.gl-style-format.ts
```diff
@@ -0,0 +1,44 @@
+import replace from '@rollup/plugin-replace';
```
HarelM, line 1: Consider using a single config to all 3 scripts and a single build command in package.json file.

package.json
```diff
@@ -18,6 +18,10 @@
   "type": "module",
   "scripts": {
     "build-style-spec": "rollup --configPlugin @rollup/plugin-typescript -c rollup.config.style-spec.ts",
+    "build-gl-style-format": "rollup --configPlugin @rollup/plugin-typescript -c rollup.config.gl-style-format.ts",
```
HarelM, line 21: build-bin to all 3 scripts?

HarelM: Also add it to the CI to make sure it doesn't break.

PR #23, opened 2023-02-24
Add style spec docs

birkskyum:
Part of
https://github.com/maplibre/maplibre-gl-js/issues/2149
https://github.com/maplibre/maplibre-gl-js-docs/issues/400

This copies the files "as is" as suggested in https://github.com/maplibre/maplibre-gl-js/issues/2149, and is thus not a refactor. A refactor can be done either in maplibre-gl-js-docs before the transfer, or after the merge, but it's confusing to do it at the same time as git can't show the alterations to the files then.


HarelM:
It would be great if you could, as part of this PR, write some of your knowledge regarding how all this works in terms of steps, technologies, etc.
Either as a readme in this repo or in the initial comment of this PR.
So I could
A. Understand how this works
B. Be able to better review it

HarelM:
I know this is still a draft, but I've seen some files that are not accurate or i don't think should be added.
Let's try and bring the minimal number of files that will allow the creation of the docs and not copy the entire docs repo here.
I might be stating the obvious, but I want to make sure we are on the same page...

birkskyum:
Sure. I'll try to make some refactoring to the existing repo first, and then do the split.

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Build style spec
+      run: |
+        npm run generate-style-spec
+        npm run generate-typings
+        npm run compile
+
+    - name: Build docs
+      run: |
+        cd ./docs
```
HarelM, line 28: Why do we need to `cd` into a folder and start `npm ci` and all that from that folder and not the main folder?

docs/batfish.config.js
```diff
@@ -0,0 +1,98 @@
+const webpack = require('webpack');
+const mapboxAssembly = require('@mapbox/mbx-assembly');
+const path = require('path');
+const {
+    buildNavigation,
+    buildFilters
+} = require('@mapbox/dr-ui/helpers/batfish/index.js');
+
+const addPages = [
+    // {
```
HarelM, line 10: Do we need to keep this comment?

docs/batfish.config.js
```diff
@@ -0,0 +1,98 @@
+const webpack = require('webpack');
```
HarelM, line 1: Can we use `import` instead of `require`?

docs/batfish.config.js
```diff
@@ -0,0 +1,98 @@
...
+const mapboxAssembly = require('@mapbox/mbx-assembly');
```
HarelM, line 2: Do we have to keep this dependency? Can we hard code some of the setting here instead?

docs/docs/bin/create-image.js
```diff
@@ -0,0 +1,101 @@
+// eslint-disable-line
```
HarelM, line 1: docs/docs/bin seems like a strange folder structure...

.github/workflows/docs-deploy.yml
birkskyum: Because I think it's unwise to pollute the style-spec package with the many many deps etc form the docs page.

HarelM:
That's a lot of files to review... It might take me a while...

docs/docs/bin/create-image.js
```diff
@@ -0,0 +1,101 @@
...
+import path from 'path';
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+import pack from 'maplibre-gl/package.json' assert { type: 'json' };
+import puppeteer from 'puppeteer';
```
HarelM, line 6: We need to remove this dependency, this tool is basically deprecated.
We also should consider migrating this file to typescript.

docs/docs/bin/create-image.js
```diff
@@ -0,0 +1,101 @@
...
+
+    // eslint-disable-next-line xss/no-mixed-html
+    await page.setContent(html);
+
+    // Wait for map to load, then wait two more seconds for images, etc. to load.
+    await page
+        .waitForFunction('map.loaded()')
+        .then(async () => {
+            // Wait for 5 seconds on 3d model examples, since this takes longer to load.
+            const waitTime = exampleName.includes('3d-model') ? 5000 : 1500;
```
HarelM, line 56: This is a very specific if, I would love to have it removed somehow...
Is this script used in this repo? Isn't this only for the examples images which are not part of the style-spec docs?

docs/docs/bin/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
HarelM, line 1: Is there a way to remove this file and have all this repo use `type: module`?

docs/docs/components/__tests__/expression-metadata.test.js
```diff
@@ -0,0 +1,24 @@
+import { types } from '../expression-metadata';
+import ref from '../../../../tsc/src/reference/latest';
+
+// help us catch that every type in expression-metadata.js exists in the style-spec
+// otherwise the site will not build
+describe('expression-metadata.js type exists in maplibre-gl-js-style-spec', () => {
```
HarelM, line 6: I'm not sure I understand what this test is actually testing.
It seems that this should be part of the regular tests of the style-spec and should not be here...

docs/docs/components/__tests__/expression-metadata.test.js
```diff
@@ -0,0 +1,24 @@
...
+    for (const name in types) {
+        it(name, () => {
+            expect(ref['expression_name'].values[name]).not.toBeUndefined();
+        });
+    }
+});
+
+// help us catch expressions we need to add to expression-metadata.js during version bumps
```
HarelM, line 14: I'm sure there are better ways to do it instead of catching it in a test - i.e. better automation...

birkskyum:
Keep in mind that this PR, related to https://github.com/maplibre/maplibre-gl-js/issues/2149 has the scope of 

> As a starting point I would advise to simply move parts of this repo as is (without any code changes)

Refactoring the whole thing is quite a scope creep

docs/docs/components/application-wrapper.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
```
HarelM, line 2: This can be easily solved by typescript I guess instead of using this `prop-types` dependency.
But I'm not sure I understand what all this react code is... Can you help me understand?

docs/docs/components/appropriate-image.js
```diff
@@ -0,0 +1,12 @@
+import { scopeAppropriateImage } from '@mapbox/appropriate-images-react';
+import imageConfig from '../img/dist/image.config.json'; // eslint-disable-line
```
HarelM, line 2: We are importing a json here, shouldn't we import with assert keyword?

docs/docs/components/expression-metadata.js
```diff
@@ -0,0 +1,754 @@
...
+        (ctx, [v]) => typeToString(typeOf(v.evaluate(ctx)))
+    ],
+    'to-rgba': [
+        array(NumberType, 4),
+        [ColorType],
+        (ctx, [v]) => {
+            return v.evaluate(ctx).toArray();
+        }
+    ],
+    rgb: [ColorType, [NumberType, NumberType, NumberType], rgba],
```
HarelM, line 82: This part (all the expressions here) should be generated, otherwise there's a need to change this file every time we change the expressions.

docs/docs/components/page-shell.js
```diff
@@ -0,0 +1,129 @@
...
+import { styleSpecNavigation } from '../data/style-spec-navigation';
+
+import AppropriateImage from './appropriate-image';
+import Browser from '@mapbox/dr-ui/browser';
+import classnames from 'classnames';
+import { version as styleSpecVersion } from '../../../package.json';
+
+const redirectStyleSpec = require('../util/style-spec-redirect');
+
+class PageShell extends React.Component {
```
HarelM, line 25: I'm not sure this should be here, this looks like the definition of the main react page in the docs.
If we want to make it modular, we can have a look at the direction of micro front ends, or simply move to a page containing the style-spec definitions without all the other parts of the docs.
Bottom line, this is probably duplicated between here and the other docs repo and there's a need to better restructure this...

docs/docs/components/prism_highlight.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { highlightJsx } from '@mapbox/dr-ui/highlight/jsx';
```
HarelM, line 2: Probably not part of this PR but using a project that writes the following is shooting ourselves in the foot:
>This project is for internal Mapbox usage. The code is open source and we appreciate bug reports; but we will only consider feature requests and pull requests from Mapbox developers.

docs/docs/components/style-spec/item.js
```diff
@@ -0,0 +1,368 @@
...
+            case 'transition':
+                return (
+                    <span>
+                        {' '}
+                        <a href="/maplibre-gl-style-spec/style-spec/transition/">
+                            transition
+                        </a>
+                    </span>
+                );
+            case 'sources':
```
HarelM, line 35: Also just a poor decision to replicate the style spec here in the UI... This means that `terrain` and `padding` is probably not here...

docs/docs/constants.json
```diff
@@ -0,0 +1,8 @@
+{
+  "SITE": "MapLibre GL",
+  "BASEURL": "/maplibre-gl-style-spec",
+  "FORWARD_EVENT_WEBHOOK": {
+    "staging": "https://evj5gwoa8j.execute-api.us-east-1.amazonaws.com/hookshot/webhook",
```
HarelM, line 5: What are these links??

docs/docs/data/expressions-related.json
```diff
@@ -0,0 +1,220 @@
+{
```
HarelM, line 1: Is this needed for the style-spec docs?
Can we somehow incorporate this in the v8.json file?

docs/docs/data/style-spec-navigation.js
```diff
@@ -0,0 +1,239 @@
...
+                    level,
+                    ...(icon && { icon: icon })
+                });
+            }
+
+            return arr;
+        }, []);
+}
+
+export const styleSpecNavigation = [
```
HarelM, line 49: Can this be generated as well?

docs/docs/documentation.yml
```diff
@@ -0,0 +1 @@
+toc:
```
HarelM, line 1: hu?

docs/docs/pages/style-spec/expressions.md
```diff
@@ -0,0 +1,211 @@
...
+<!-- This h1 is an exception as the following headings need to be an h2 for scrollspy to work on the sidebar. See site.css for override styles for these headings. -->
+
+
+## Types
+
+The expressions in this section are for testing for and converting between different data types like strings, numbers, and boolean values.
+
+Often, such tests and conversions are unnecessary, but they may be necessary in some expressions where the type of a certain sub-expression is ambiguous. They can also be useful in cases where your feature data has inconsistent types; for example, you could use `to-number` to make sure that values like `"1.5"` (instead of `1.5`) are treated as numeric values.
+
+{{<ExpressionReference group='Types' />}}
```
HarelM, line 169: Mixing markdown and react components seems like an odd choice.
I would prefer using a single technology instead of this mixture.
Preferably, that the part that are "dynamic" are generated and not partially programmed using code, syntax, and libraries that are not documented...
As an initial phase, I would recommend removing these react components and simply hardcode everything into the markdown.
After that we can decide how to proceed.

docs/docs/pages/style-spec/other.md
```diff
@@ -0,0 +1,385 @@
...
+    - "import Note from '@mapbox/dr-ui/note';"
+---
+
+<!--copyeditor disable basic-->
+<!--copyeditor disable js-->
+<!--copyeditor disable ios-->
+<!--copyeditor disable macos-->
+<!--copyeditor disable all-of-->
+
+Some style properties are no longer the preferred method of accomplishing a particular styling goal. While they are still supported, they will eventually be removed from the MapLibre Style Specification and it is not recommended to use them in new styles. The following is provided as reference for users who need to continue maintaining legacy styles while transitioning to preferred style properties.
```
HarelM, line 26: I'm not sure we fully stand behind this remark.
So, I'm not sure how valuable this page is... I would consider removing it.

docs/docs/pages/style-spec/root.md
```diff
@@ -0,0 +1,36 @@
...
+    - "import ref from '../../../../tsc/src/reference/latest';"
+---
+
+Root level properties of a MapLibre style specify the map's layers, tile sources and other resources, and default values for the initial camera position when not specified elsewhere.
+
+```json
+{
+    "version": {{ref.$version}},
+    "name": "Mapbox Streets",
+    "sprite": "mapbox://sprites/mapbox/streets-v{{ref.$version}}",
```
HarelM, line 22: We should probably change that to `https://...` as we don't support mapbox protocol anymore.

docs/docs/pages/style-spec/root.md
```diff
@@ -0,0 +1,36 @@
...
+description: Root level properties of a MapLibre style specify the map's layers, tile sources and other resources, and default values for the initial camera position when not specified elsewhere.
+contentType: specification
+order: 2
+layout: page
+hideFeedback: true
+products:
+- MapLibre Style Specification
+prependJs:
+    - "import Items from '../../components/style-spec/items';"
+    - "import ref from '../../../../tsc/src/reference/latest';"
```
HarelM, line 13: Can the docs import from the src but not from the tsc folder or this is a problem due to how the docs are deployed?

docs/docs/pages/style-spec/sources.md
```diff
@@ -0,0 +1,319 @@
...
+    "coordinates": [
+        [-122.51596391201019, 37.56238816766053],
+        [-122.51467645168304, 37.56410183312965],
+        [-122.51309394836426, 37.563391708549425],
+        [-122.51423120498657, 37.56161849366671]
+    ]
+}
+```
+
+<!--
```
HarelM, line 302: Seems like there's a generated code that's coming from the v8.json file, can't we move all this docs into it and avoid this duplication of docs and effort?

docs/local.dic
```diff
@@ -0,0 +1,32 @@
+multipolyline
```
HarelM, line 1: Where is this used? lint?

docs/package.json
```diff
@@ -0,0 +1,96 @@
+{
+  "name": "maplibre-gl-style-spec-docs",
+  "description": "MapLibre GL JS documentation website",
+  "license": "BSD-3-Clause",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/maplibre/maplibre-gl-style-spec.git"
+  },
+  "devDependencies": {
```
HarelM, line 9: just... too... many... :-(((

docs/remove-assert.js
```diff
@@ -0,0 +1,15 @@
+var fs = require('fs');
```
HarelM, line 1: This is needed for batfish, right?

docs/vendor/docs-page-shell/page-shell-styles.css
```diff
@@ -0,0 +1,733 @@
+/* ========================================================================== */
+/* shell-prefixed mapbox-assembly + assembly v0.16.0 */
+/* ========================================================================== */
+
+.shell-wrapper ul,
```
HarelM, line 5: holly sh*t.... this css is huge.
We got to use a library that comes with a pre-built css...

HarelM:
I've added tons of comments. :-(
I'm not sure what should be the right direction here.
We can decide to take all this huge codebase and use it as is, but I think it will take longer to move it here, make it work inside the "regular" docs pages and then modify it to a new system.
Instead, I think we should try and define this new docs system, just for the style-spec at this point and use that as a proof of concept to see that it can facilitate our spec docs.
Once this works (and it should be a lot simpler than what we currently have with react, and routing, and special components, and tons of css that we'll never be able to maintain) and we can generate a simple docs site (out of markdown or I don't know what) we should deploy it, even if it looks different and might not have all the functionality it used to have.
Once this is up and running we can see how to extend it to the other parts of the documentation.
I don't see the "single look and feel" for all the docs as a hard requirement. While the UX might not be supper in terms of moving from one part of the docs to the other, if the maintenance effort is low, and we can easily change the spec, code, whatever without doing a rework to update the docs the end user will benefit from up-to-date and accurate docs, which I think is a better UX.
Let me know what you think.

birkskyum:
I think most of the comments here are meaningful but more related to general maintenance of the docs, than related to the split.

I also think a different docs system would be worth investing in compared to a big rework within the batfish environment, especially for the style-spec, and this is likely easier to approach after splitting the docs as done in this PR.

HarelM:
I tend to think it will take you longer to make this work after the split inside the docs website as it is right now due to all the existing hacks and quirks.
Instead, moving to a simple docs like github pages default or jekyll might prove a lot simpler and straight forward and prevent a lot of grief.
IDK, your call, you know the code base better and which of the options is shorter.

birkskyum:
Getting the page up "as is" after the split is not so time consuming, but reworking it would likely still not be worth it. What it would bring though is the ability to locally make a change to style-spec and see it reflected in the docs instead of having to go through npm.

HarelM:
Please convert the comments that you are not planning to address to issues in this repo, this way we can track the tech debt we are creating.
I think you are too optimistic about the DNS, but I'd be happy to be wrong...
A lot of the spec docs are already in markdown, so for me it seems like the right direction to take this opportunity to break out of the current system, especially since I feel like a few files were brought here which don't feel like they should be here and are duplicated from the other docs repo, which will cause pain if someone would like to change things in the near future...
To be honest, I don't feel great about this PR as I feel it creates a larger tech debt, but again, I might be wrong.
Having said that, if you need me to approve it, I'll do it.

birkskyum:
Let's cancel this then and invest the energy in making entirely new docs for the style-spec. Maybe even something like astro could make sense here

HarelM:
I've done some reading last night, what I found which seems the most straight forward is using jekyll or kramdown with static content hosted in a github folder in this repo.
This way you have the docs both in the repo and in github pages.
It's easy to edit, preview etc.
I've looked at the lodash docs for example, and they seem to use the same.
We can also go for hugo like the maplibre.org website, but I'm not sure it's really needed and I found it a bit complicated.
We can create a small script that generates md files from the v8.json file (and maybe checks the output, or not, I haven't fully thought about it).
One this is in place, you have everything you need handy with little to no dependencies...
Feel free to ping me on slack if you want to do a brainstorm session.

HarelM:
As a side note, and this PR might not be the right place for these comments, but here goes anyway:
I've looked at the following options (no actual code, just reading about them):
1. Jekyll
2. Hugo
3. Astro

In the current documentation site, there is a left sidebar and a right sidebar with table of contents - one for all the docs and the other for the current page which highlights in the TOC as you scroll the page, which is useful.
I've found some comments about a jekyll theme that can support this (maybe): Minimal mistakes and this blog post:
https://mmistakes.github.io/minimal-mistakes/layout-table-of-contents-sticky/
And maybe some other comments about it for "just-the-docs" theme that should be able to support this (maybe).
This is what I found for hugo:
https://ma.ttias.be/adding-a-sticky-table-of-contents-in-hugo-to-posts/
I did not find something similar for Astro, although this is probably possible with html and css it seems that this echosystem is a bit immature...

PR #93, opened 2023-04-06
Remove tsc compilation of individual files

birkskyum:
Due to the old docs, we had to be able to reach each file as a .js, and tsc compiling everything to a /tsc folder was the workaround.

Now we just use the typescript or the bundle instead.


HarelM (approved):

HarelM:
There's a need to remove this command from the package.json as well...?

birkskyum:
@louwers , the native repo doesn't use the /tsc .js files does it? Noticed the gl js submodule was finally removed.

louwers:
Nope.

PR #94, opened 2023-04-06
Add CSS Color 4 syntax support, color interpolation changes

kajkal:
- CSS Color 4 support
- Color interpolation changes, [see live demo](https://kajkal.github.io/maplibre-gl-style-spec-color-demo/)
    - fix interpolation in hcl and lab color spaces
    - include colors with alpha=0 in color interpolation calculations

See more in discussion #70 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -1,0 +1,29 @@
+import Color from './src/util/color';
+
+expect.extend({
+
+    toMatchColor(received: unknown, expectedSerialized: string, numDigits = 5) {
+        const [r, g, b, a] = expectedSerialized.match(/^rgb\(([\d.]+)% ([\d.]+)% ([\d.]+)% \/ ([\d.]+)\)$/).slice(1).map(Number);
+        const expected = expect.objectContaining({
+            r: expect.closeTo(r / 100 * a, numDigits),
+            g: expect.closeTo(g / 100 * a, numDigits),
+            b: expect.closeTo(b / 100 * a, numDigits),
+            a: expect.closeTo(a, 4),
+        });
+        const pass = (received instanceof Color) && this.equals(received, expected);
+        return {
+            pass,
+            message: () => `${this.utils.matcherHint('toMatchColor', this.utils.printReceived(received), expectedSerialized)}\n\n${this.utils.diff(expected, received)}`,
+        };
+    },
+
+    closeToNumberArray(received: unknown, expectedArray: number[], numDigits = 5) {
+        const expected = expectedArray.map(n => isNaN(n) ? n : expect.closeTo(n, numDigits));
+        const pass = this.equals(received, expected);
+        return {
+            pass,
+            message: () => `${this.utils.matcherHint('closeToNumberArray', this.utils.printReceived(received), this.utils.printExpected(expectedArray))}\n\n${this.utils.diff(expected, received)}`,
+        };
+    },
+
+});
--- a/src/expression/definitions/interpolate.ts
+++ b/src/expression/definitions/interpolate.ts
@@ -1,9 +1,8 @@
 import UnitBezier from '@mapbox/unitbezier';
 
-import * as interpolate from '../../util/interpolate';
+import interpolate from '../../util/interpolate';
 import {toString, NumberType, ColorType} from '../types';
 import {findStopLessThanOrEqualTo} from '../stops';
-import {hcl, lab} from '../../util/color_spaces';
 
 import type {Stops} from '../stops';
 import type {Expression} from '../expression';
@@ -174,12 +173,13 @@ class Interpolate implements Expression {
         const outputLower = outputs[index].evaluate(ctx);
         const outputUpper = outputs[index + 1].evaluate(ctx);
 
-        if (this.operator === 'interpolate') {
-            return ((interpolate[this.type.kind.toLowerCase()] as any))(outputLower, outputUpper, t); // eslint-disable-line import/namespace
-        } else if (this.operator === 'interpolate-hcl') {
-            return hcl.reverse(hcl.interpolate(hcl.forward(outputLower), hcl.forward(outputUpper), t));
-        } else {
-            return lab.reverse(lab.interpolate(lab.forward(outputLower), lab.forward(outputUpper), t));
+        switch (this.operator) {
+            case 'interpolate':
+                return interpolate[this.type.kind.toLowerCase()](outputLower, outputUpper, t);
+            case 'interpolate-hcl':
+                return interpolate.color(outputLower, outputUpper, t, 'hcl');
+            case 'interpolate-lab':
+                return interpolate.color(outputLower, outputUpper, t, 'lab');
         }
     }
 
--- a/src/function/index.ts
+++ b/src/function/index.ts
@@ -1,9 +1,7 @@
-
-import * as colorSpaces from '../util/color_spaces';
 import Color from '../util/color';
 import extend from '../util/extend';
 import getType from '../util/get_type';
-import * as interpolate from '../util/interpolate';
+import interpolate, {isSupportedInterpolationColorSpace} from '../util/interpolate';
 import Interpolate from '../expression/definitions/interpolate';
 import Formatted from '../expression/types/formatted';
 import ResolvedImage from '../expression/types/resolved_image';
@@ -44,8 +42,8 @@ export function createFunction(parameters, propertySpec) {
         }
     }
 
-    if (parameters.colorSpace && parameters.colorSpace !== 'rgb' && !colorSpaces[parameters.colorSpace]) { // eslint-disable-line import/namespace
-        throw new Error(`Unknown color space: ${parameters.colorSpace}`);
+    if (parameters.colorSpace && !isSupportedInterpolationColorSpace(parameters.colorSpace)) {
+        throw new Error(`Unknown color space: "${parameters.colorSpace}"`);
     }
 
     let innerFun;
@@ -176,12 +174,7 @@ function evaluateExponentialFunction(parameters, propertySpec, input) {
 
     const outputLower = parameters.stops[index][1];
     const outputUpper = parameters.stops[index + 1][1];
-    let interp = interpolate[propertySpec.type] || identityFunction; // eslint-disable-line import/namespace
-
-    if (parameters.colorSpace && parameters.colorSpace !== 'rgb') {
-        const colorspace = colorSpaces[parameters.colorSpace]; // eslint-disable-line import/namespace
-        interp = (a, b) => colorspace.reverse(colorspace.interpolate(colorspace.forward(a), colorspace.forward(b), t));
-    }
+    const i11nFn = interpolate[propertySpec.type] || identityFunction;
 
     if (typeof outputLower.evaluate === 'function') {
         return {
@@ -192,12 +185,12 @@ function evaluateExponentialFunction(parameters, propertySpec, input) {
                 if (evaluatedLower === undefined || evaluatedUpper === undefined) {
                     return undefined;
                 }
-                return interp(evaluatedLower, evaluatedUpper, t);
+                return i11nFn(evaluatedLower, evaluatedUpper, t, parameters.colorSpace);
             }
         };
     }
 
-    return interp(outputLower, outputUpper, t);
+    return i11nFn(outputLower, outputUpper, t, parameters.colorSpace);
 }
 
 function evaluateIdentityFunction(parameters, propertySpec, input) {
--- a/src/util/color.ts
+++ b/src/util/color.ts
@@ -1,27 +1,29 @@
-import {parseCSSColor} from 'csscolorparser';
+import colorString from 'color-string';
+import {hcl, hsl, lab, HCLColor, LABColor, RGBColor} from './color_spaces';
 
 /**
- * An RGBA color value. Create instances from color strings using the static
- * method `Color.parse`. The constructor accepts RGB channel values in the range
- * `[0, 1]`, premultiplied by A.
- *
- * @param {number} r The red channel.
- * @param {number} g The green channel.
- * @param {number} b The blue channel.
- * @param {number} a The alpha channel.
+ * Color representation used by WebGL.
+ * Defined in sRGB color space and pre-blended with alpha.
  * @private
  */
 class Color {
-    r: number;
-    g: number;
-    b: number;
-    a: number;
 
-    constructor(r: number, g: number, b: number, a: number = 1) {
+    readonly r: number;
+    readonly g: number;
+    readonly b: number;
+    readonly a: number;
+
+    /**
+     * @param r Red component premultiplied by `alpha` 0..1
+     * @param g Green component premultiplied by `alpha` 0..1
+     * @param b Blue component premultiplied by `alpha` 0..1
+     * @param [alpha=1] Alpha component 0..1
+     */
+    constructor(r: number, g: number, b: number, alpha: number = 1) {
         this.r = r;
         this.g = g;
         this.b = b;
-        this.a = a;
+        this.a = alpha;
     }
 
     static black: Color;
@@ -30,59 +32,90 @@ class Color {
     static red: Color;
 
     /**
-     * Parses valid CSS color strings and returns a `Color` instance.
-     * @param input A valid CSS color string.
+     * Parses CSS color strings and converts colors to sRGB color space if needed.
+     * Officially supported color formats:
+     * - keyword, e.g. 'aquamarine' or 'steelblue'
+     * - hex (with 3, 4, 6 or 8 digits), e.g. '#f0f' or '#e9bebea9'
+     * - rgb and rgba, e.g. 'rgb(0,240,120)' or 'rgba(0%,94%,47%,0.1)' or 'rgb(0 240 120 / .3)'
+     * - hsl and hsla, e.g. 'hsl(0,0%,83%)' or 'hsla(0,0%,83%,.5)' or 'hsl(0 0% 83% / 20%)'
+     *
+     * @param input CSS color string to parse.
      * @returns A `Color` instance, or `undefined` if the input is not a valid color string.
      */
-    static parse(input?: string | Color | null): Color | void {
-        if (!input) {
-            return undefined;
+    static parse(input: String | string | undefined | null): Color | undefined {
+        if (Object.prototype.toString.call(input) !== '[object String]') {
+            return;
         }
 
-        if (input instanceof Color) {
-            return input;
+        const parsedColor = parseCssColor(input.toLowerCase());
+        if (parsedColor) {
+            const [r, g, b, alpha] = parsedColor;
+            const color = new Color(r * alpha, g * alpha, b * alpha, alpha);
+            if (!alpha) {
+                // alpha = 0 erases completely rgb channels. This behavior is not desirable
+                // if this particular color is later used in color interpolation.
+                // Because of that, a reference to original color is saved.
+                color.overwriteGetter('rgb', parsedColor);
+            }
+            return color;
         }
-
-        if (typeof input !== 'string') {
-            return undefined;
-        }
-
-        const rgba = parseCSSColor(input);
-        if (!rgba) {
-            return undefined;
-        }
-
-        return new Color(
-            rgba[0] / 255 * rgba[3],
-            rgba[1] / 255 * rgba[3],
-            rgba[2] / 255 * rgba[3],
-            rgba[3]
-        );
     }
 
     /**
-     * Returns an RGBA string representing the color value.
+     * Used in color interpolation.
      *
-     * @returns An RGBA string.
-     * @example
-     * var purple = new Color.parse('purple');
-     * purple.toString; // = "rgba(128,0,128,1)"
-     * var translucentGreen = new Color.parse('rgba(26, 207, 26, .73)');
-     * translucentGreen.toString(); // = "rgba(26,207,26,0.73)"
+     * @returns Gien color, with reversed alpha blending, in sRGB color space.
      */
-    toString(): string {
-        const [r, g, b, a] = this.toArray();
-        return `rgba(${Math.round(r)},${Math.round(g)},${Math.round(b)},${a})`;
+    get rgb(): RGBColor {
+        const {r, g, b, a} = this;
+        return this.overwriteGetter('rgb', [r / a, g / a, b / a, a]);
     }
 
-    toArray(): [number, number, number, number] {
-        const {r, g, b, a} = this;
-        return a === 0 ? [0, 0, 0, 0] : [
-            r * 255 / a,
-            g * 255 / a,
-            b * 255 / a,
-            a
-        ];
+    /**
+     * Used in color interpolation.
+     *
+     * @returns Gien color, with reversed alpha blending, in HCL color space.
+     */
+    get hcl(): HCLColor {
+        return this.overwriteGetter('hcl', hcl.fromRgb(this.rgb));
+    }
+
+    /**
+     * Used in color interpolation.
+     *
+     * @returns Gien color, with reversed alpha blending, in LAB color space.
+     */
+    get lab(): LABColor {
+        return this.overwriteGetter('lab', lab.fromRgb(this.rgb));
+    }
+
+    /**
+     * Lazy getter pattern. When getter is called for the first time lazy value
+     * is calculated and then overwrites getter function in given object instance.
+     *
+     * @param getterKey Getter key
+     * @param lazyValue Lazy value
+     * @private
+     */
+    private overwriteGetter<T>(getterKey: string, lazyValue: T): T {
+        Object.defineProperty(this, getterKey, {value: lazyValue});
+        return lazyValue;
+    }
+
+}
+
+function parseCssColor(colorToParse: string): RGBColor | undefined {
+    const parsingResult = colorString.get(colorToParse);
+    if (parsingResult) {
+        switch (parsingResult.model) {
+            case 'rgb': {
+                const [r, g, b, alpha] = parsingResult.value;
+                return [r / 255, g / 255, b / 255, alpha];
+            }
+            case 'hsl': {
+                return hsl.toRgb(parsingResult.value);
+            }
+        }
     }
 }
 
--- a/src/util/color_spaces.ts
+++ b/src/util/color_spaces.ts
@@ -1,20 +1,34 @@
-import Color from './color';
+/**
+ * @param r Red component 0..1
+ * @param g Green component 0..1
+ * @param b Blue component 0..1
+ * @param alpha Alpha component 0..1
+ */
+export type RGBColor = [r: number, g: number, b: number, alpha: number];
 
-import {number as interpolateNumber} from './interpolate';
+/**
+ * @param h Hue as degrees 0..360
+ * @param s Saturation as percentage 0..100
+ * @param l Lightness as percentage 0..100
+ * @param alpha Alpha component 0..1
+ */
+export type HSLColor = [h: number, s: number, l: number, alpha: number];
 
-type LABColor = {
-    l: number;
-    a: number;
-    b: number;
-    alpha: number;
-};
+/**
+ * @param h Hue as degrees 0..360
+ * @param c Chroma 0..~230
+ * @param l Lightness as percentage 0..100
+ * @param alpha Alpha component 0..1
+ */
+export type HCLColor = [h: number, c: number, l: number, alpha: number];
 
-type HCLColor = {
-    h: number;
-    c: number;
-    l: number;
-    alpha: number;
-};
+/**
+ * @param l Lightness as percentage 0..100
+ * @param a A axis value -125..125
+ * @param b B axis value -125..125
+ * @param alpha Alpha component 0..1
+ */
+export type LABColor = [l: number, a: number, b: number, alpha: number];
 
 // Constants
 const Xn = 0.950470, // D65 standard referent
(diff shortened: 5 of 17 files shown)


src/function/index.ts
```diff
@@ -176,12 +174,7 @@ function evaluateExponentialFunction(parameters, propertySpec, input) {
 
     const outputLower = parameters.stops[index][1];
     const outputUpper = parameters.stops[index + 1][1];
-    let interp = interpolate[propertySpec.type] || identityFunction; // eslint-disable-line import/namespace
-
-    if (parameters.colorSpace && parameters.colorSpace !== 'rgb') {
-        const colorspace = colorSpaces[parameters.colorSpace]; // eslint-disable-line import/namespace
-        interp = (a, b) => colorspace.reverse(colorspace.interpolate(colorspace.forward(a), colorspace.forward(b), t));
-    }
+    const i11nFn = interpolate[propertySpec.type] || identityFunction;
```
HarelM, line 177: Please change to full method name, without abbreviations.

src/util/color.ts
```diff
@@ -1,27 +1,29 @@
...
  * @private
  */
 class Color {
-    r: number;
-    g: number;
-    b: number;
-    a: number;
 
-    constructor(r: number, g: number, b: number, a: number = 1) {
+    readonly r: number;
```
HarelM, line 11: This can be declared right in the constructor, no need to repeat these 4 variables.

src/util/color.ts
```diff
@@ -30,59 +32,88 @@ class Color {
...
+     * - hsl and hsla, e.g. 'hsl(0,0%,83%)' or 'hsla(0,0%,83%,.5)' or 'hsl(0 0% 83% / 20%)'
+     *
+     * @param input CSS color string to parse.
      * @returns A `Color` instance, or `undefined` if the input is not a valid color string.
      */
-    static parse(input?: string | Color | null): Color | void {
-        if (!input) {
-            return undefined;
+    static parse(input: String | string | undefined | null): Color | undefined {
+        if (Object.prototype.toString.call(input) !== '[object String]') {
```
HarelM, line 46: Why call to string? seems a bit expensive, why not use the original `input instanceof Color?

src/util/interpolate.ts
```diff
@@ -10,41 +29,67 @@ export const interpolateFactory = (interpolationType: 'number'|'color'|'array'|'
...
 }
 
-export function color(from: Color, to: Color, t: number) {
-    return new Color(
-        number(from.r, to.r, t),
-        number(from.g, to.g, t),
-        number(from.b, to.b, t),
-        number(from.a, to.a, t)
-    );
+const colorInterpolation = {
```
HarelM, line 36: Better to have this as a static method with switch case instead of an object that is called with a key, more readable, easier find and replace, and more performant.

src/util/interpolate.ts
```diff
@@ -1,6 +1,25 @@
+import {hcl, lab, RGBColor} from './color_spaces';
 import Color from './color';
 import Padding from './padding';
 
+export type InterpolationColorSpace = keyof typeof colorInterpolation;
```
HarelM, line 5: isn't this simply `'rbg' | 'hcl' | 'lab'`? if so, it's much easier to understand using these simple 3 options instead of `keyof typeof ...`.

src/util/color_spaces.ts
```diff
@@ -27,111 +41,101 @@ const Xn = 0.950470, // D65 standard referent
...
+    }
+
+    return [f(0), f(8), f(4), alpha];
 }
 
 export const lab = {
-    forward: rgbToLab,
-    reverse: labToRgb,
-    interpolate: interpolateLab
+    fromRgb: rgbToLab,
```
HarelM, line 130: why not simply export the methods in this file instead of these objects?

src/util/color_spaces.ts
```diff
@@ -27,111 +41,101 @@ const Xn = 0.950470, // D65 standard referent
...
-    const h = hclColor.h * deg2rad,
-        c = hclColor.c,
-        l = hclColor.l;
-    return labToRgb({
-        l,
-        a: Math.cos(h) * c,
-        b: Math.sin(h) * c,
-        alpha: hclColor.alpha
-    });
+// sRGB -> HCL
```
HarelM, line 100: Regular jsdocs would be better here...

src/util/color_spaces.ts
```diff
@@ -1,20 +1,34 @@
...
-    l: number;
-    alpha: number;
-};
+/**
+ * @param r Red component 0..1
+ * @param g Green component 0..1
+ * @param b Blue component 0..1
+ * @param alpha Alpha component 0..1
+ */
+export type RGBColor = [r: number, g: number, b: number, alpha: number];
```
HarelM, line 7: I tend to say the previous implementation was better in terms of debugging, finding references etc as the new definition is basically all a 4 numbers array which is extremely error prone.

jest.setup.ts
```diff
@@ -0,0 +1,29 @@
+import Color from './src/util/color';
+
+expect.extend({
```
HarelM, line 3: Can we do it without this complication? While I get the point behind it, I would like to avoid these special extenstions, or better yet have this as part of the test utils so if we decide to migrate to a different test framework it will be easier.

src/util/color.ts
```diff
@@ -30,59 +32,88 @@ class Color {
...
-     * var translucentGreen = new Color.parse('rgba(26, 207, 26, .73)');
-     * translucentGreen.toString(); // = "rgba(26,207,26,0.73)"
+     * @param getterKey Getter key
+     * @param lazyValue Lazy value
+     * @private
      */
-    toString(): string {
-        const [r, g, b, a] = this.toArray();
-        return `rgba(${Math.round(r)},${Math.round(g)},${Math.round(b)},${a})`;
+    private overwriteGetter<T>(getterKey: string, lazyValue: T): T {
```
HarelM, line 100: Can you elaborate on the purpose behind this method? seems complicated...

HarelM:
Thanks for taking the time to push this forward!
I think this is great.
I've added some comments to reduce complexity in some cases.

src/function/index.ts
kajkal: Oh I thought it was clever, interpolation => i11n just like internationalization is i18n :<
I will restore the previous name.

HarelM: Yea, I got it 😀 but still, I would even prefer the full name, not just interp.

src/util/color.ts
kajkal: The intention behind this function is to parse strings, there should be no situation where an already parsed color is passed to `Color.parse` method, checking if `instanceof Color` in my opinion is beyond the scope of this function. I believe that this is a dead code from ancient times and I took the liberty of removing it.

Support for `String` objects has been added because of the `validateColor` flow, which now uses this function to check if a string is a valid color. Actually this can be avoided by casting `String` into `string` primitive in `validateColor`. So never mind, I'll change it.

src/util/color_spaces.ts
kajkal: They neatly group functions related to a single color space.

src/util/color_spaces.ts
kajkal: These comments are just to separate the fragments related to a particular transformation. Like marking that the functions `rgb2xyz` and `xyz2lab` are subject to and used only by `rgbToLab`.
In addition, the exported `lab` `hcl` and `hsl` objects in the context of the file name and the names of the methods they contain are self-explanatory enough that additional comments are not necessary.

src/util/color_spaces.ts
kajkal: These data types do not go beyond color interpolation - in other words, they are only used when called by the `interpolate.color` function which instead initially takes colors as instances of the `Color` class and which also returns colors as instances of the `Color` class. So I don't think that declaring this data as objects gives much in the way of improving the debuggability of the code. 
On the other hand, as arrays, they are much more versatile and allow, for example, for reuse of code to interpolate arrays 
https://github.com/maplibre/maplibre-style/blob/e679ba7e2bf31c5d7ed99110bae654c5d257beb2/src/util/interpolate.ts#L37
https://github.com/maplibre/maplibre-style/blob/e679ba7e2bf31c5d7ed99110bae654c5d257beb2/src/util/interpolate.ts#L70
instead of the need to create such code specifically for each object of different shape:
https://github.com/maplibre/maplibre-style/blob/2408da0c626110c8525b98d6a05b9985f6be885c/src/util/interpolate.ts#L17-L24
As for error proneness: because of the aforementioned isolation of this code from the rest (only used internally by `interpolate.color`) and tests written for them I don't believe that using arrays here is a bad idea.

jest.setup.ts
kajkal: This is the kind of test utils that, in the event of migrating from jest, would still require modification regardless of whether these utils are defined as an extension to `expect` or as separate functions.
The great advantage of defining them as an extension to jest is diff printing when tests are failing.
But it's not my project, and if you don't want custom machers I'll convert it into a separate test utils function (where I'll be forced to use jest specific syntax anyway).

src/util/color.ts
kajkal: This acts as a local lazy cache for each color instance. 
Once calculated the value of a given color in HCL/LAB color space can be reused instead of recalculated every time it is needed.
So when interpolating `red` to `blue` in HCL space, the colors `red` and `blue` are converted to HCL each time an intermediate point is calculated (i.e. `interpolate.color` is called), which can be counted in several hundreds of times for these two colors alone. Memoization of `red` and `blue` values in HCL should positively affect performance of interpolations in HCL/LAB. 
Treating sRGB values in a similar way is mainly for consistency purposes, but it serves the important role of storing rgb channel values when alpha=0.

Here you could read more about lazy getters:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#smart_self-overwriting_lazy_getters

src/util/interpolate.ts
kajkal: In this case, it is important that typescript understands it and that it is in low maintenance mode :D

src/util/interpolate.ts
kajkal: Hmm… I believe that the current implementation also has its advantages. From the fact that such an object exists, we have for free the type of supported interpolation color spaces:
https://github.com/maplibre/maplibre-style/blob/e679ba7e2bf31c5d7ed99110bae654c5d257beb2/src/util/interpolate.ts#L5
and an easy-to-maintain function to check if a space is supported:
https://github.com/maplibre/maplibre-style/blob/e679ba7e2bf31c5d7ed99110bae654c5d257beb2/src/util/interpolate.ts#L14-L16
An alternative could be:
```ts
const supportedInterpolationColorSpaces = ['rgb', 'hcl', 'lab'] as const; // an extra array
export type InterpolationColorSpace = typeof supportedInterpolationColorSpaces[number];
export function isSupportedInterpolationColorSpace(colorSpace: string): colorSpace is InterpolationColorSpace {
    return supportedInterpolationColorSpaces.includes(colorSpace as InterpolationColorSpace);
}
```
or another, high-maintenance version but the most readable:
```ts
export type InterpolationColorSpace = 'rgb' | 'hcl' | 'lab';
export function isSupportedInterpolationColorSpace(colorSpace: string): colorSpace is InterpolationColorSpace {
    return colorSpace === 'rgb' || colorSpace === 'hcl' || colorSpace === 'lab';
}
```
I do not know, in my opinion it is fine as it is now, but if you think otherwise I can change it to swich case.

src/util/color.ts
kajkal: I changed this constructor to make the colors created by the `rgb` and `rgba` expressions behave the same as those created from CSS string parsing. So it's no longer relevant, I think.

kajkal:
The 4 snapshot/json tests from maplibre-gl-js need to be updated, such as [this](https://github.com/maplibre/maplibre-gl-js/blob/6d55d5ba5c0b60d074ba19e80714551498c2af68/test/integration/expression/tests/interpolate-hcl/linear/test.json) beautiful test that makes sure the interpolation in HCL goes beyond the sRGB gamut 😂 
Wouldn't these tests ^ fit more in this repository than in gl-js?
And one test related to increasing the size of the bundle by 3239 bytes D:
5 failed tests total.
I will make a PR to gl-js in the near future to address this.

As for `Post benchmark scores.` I have no idea what exactly it's about so I'm going to ignore it.
And regarding `Add an entry to CHANGELOG.md under the ## main section.` - this PR changes slightly the interpretation of interpolation from colors with alpha=0. I wouldn't call it breaking changes but someone might be surprised that some styles will look different e.g. as from example sets 2,3 and 6 from the [demo](https://kajkal.github.io/maplibre-gl-style-spec-color-demo/). I don't know if this is something worth mentioning in the changelog. And under what main version should I add support for CSS Color 4? 18.0.0? Or put it under `Features and improvements` section?

jest.setup.ts
HarelM: Yes, it would be better to have this written in the test utils so that there's no need for the extra configuration and preferably simpler methods.

src/util/color.ts
HarelM: I mean you can write `constructor(readonly a: number...` instead of defining the members and then assigning them in the constructor.

src/util/color.ts
HarelM: Is there a more readable way to write this? while I understand the code after you explain it, I don't expect someone to understand this and this can cause problems.
Also, is there a problem we know this solves? I don't like premature optimizations...

src/util/color_spaces.ts
HarelM: Please simply rename the methods to `hclToRgb` and the like and export those, the extra object creates confusion and looks as if there's an actual object that has a purpose in other files, which it doesn't.

src/util/color_spaces.ts
HarelM: I would consider adding a `toArray` method to allow interpolation and not have the entire API based on 4 numbers array. The reason I'm saying that, is that if you try to debug this code, add a console log etc, there's no indication if the array holds RGB HSL or LAB, it all looks the same, and it's confusing and can cause very hard to fix bugs, I've seen this happen over and over in my carrier.
If the main reason is for the interpolation code to be easier to write then have a utility method at that entry point.
I see that in most cases you spread the array into its components, so this won't change much I think, but it will make the API easier to understand.

src/util/color_spaces.ts
HarelM: If the comments are not needed, remove them, if they are needed turn them to jsdocs comments.
Separation comments are a code smell...

src/util/interpolate.ts
HarelM: Yes, please change to the last option, I find it more readable.

HarelM:
Thanks for taking the time to read and answer my comments!
We are still thinking about where the test should be and which tests should be where.
Once this is merged and a new version released we can update maplibre-gl-js and add the relevant render tests there, which are integration tests that intergrade both libraries and are probably a good fit there, ATM.
I'm not sure post benchmarks is related to this repo, probably copied from the other repo PR template.
Please do add a changelog item with the prefix of "[Brekaing Changes]" as this is a breaking change.

src/util/color.ts
kajkal: Personally, I find the TS syntax of defining properties of a class in constructor hard to read.
I don't see a problem with their repetition, I treat it as much as defining an interface as some kind of documentation.
It is easier to identify the properties of a class when they are defined in this way
https://github.com/maplibre/maplibre-style/blob/e679ba7e2bf31c5d7ed99110bae654c5d257beb2/src/util/color.ts#L9-L15
rather than looking for arguments preceded by `public` `readonly` etc. in its constructor.
I will leave it as it is now, if you feel strongly about it you can change it later.

src/util/interpolate.ts
kajkal: Done

src/util/color_spaces.ts
kajkal: Removed

src/util/color_spaces.ts
kajkal: Done

jest.setup.ts
kajkal: Moved to test utils file.

src/util/color.ts
kajkal: This is implementation details for the `Color` class. Exposed for `Color` class users are 3 getters: `rgb`, `hcl` and `lab` each with a specified type.
However, if at some point someone will want to understand how this internal function works, specially for him/her I expanded jsdoc with a comprehensive and illustrative example that should clear all uncertainties.

src/util/color_spaces.ts
kajkal: I understand your concerns, however, please respect my own judgment and my experience.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+### Breaking changes
```
HarelM, line 1: This shouldn't be a different section see https://github.com/maplibre/maplibre-gl-js/blob/77b2782ad11cf6aba90f366a11462f4ac62230b8/CHANGELOG.md?plain=1#L9

src/util/interpolate.ts
```diff
@@ -1,6 +1,25 @@
...
+ * interpolation color spaces, `false` otherwise
+ */
+export function isSupportedInterpolationColorSpace(colorSpace: string): colorSpace is InterpolationColorSpace {
+    return colorSpace === 'rgb' || colorSpace === 'hcl' || colorSpace === 'lab';
+}
+
+/**
+ * @param interpolationType Interpolation type
+ * @returns interpolation fn
+ * @deprecated use `interpolate[type]` instead
```
HarelM, line 21: I definitely like this factory better than the exported object and avoid exporting the object at the end?
Can we use it instead of `interpolate[type]`?
It's a lot easier to understand from the calling code what can be the options here.

src/expression/definitions/interpolate.ts
```diff
@@ -174,12 +173,13 @@ class Interpolate implements Expression {
...
 
-        if (this.operator === 'interpolate') {
-            return ((interpolate[this.type.kind.toLowerCase()] as any))(outputLower, outputUpper, t); // eslint-disable-line import/namespace
-        } else if (this.operator === 'interpolate-hcl') {
-            return hcl.reverse(hcl.interpolate(hcl.forward(outputLower), hcl.forward(outputUpper), t));
-        } else {
-            return lab.reverse(lab.interpolate(lab.forward(outputLower), lab.forward(outputUpper), t));
+        switch (this.operator) {
+            case 'interpolate':
+                return interpolate[this.type.kind.toLowerCase()](outputLower, outputUpper, t);
```
HarelM, line 178: I was planning on commenting here about readability but then saw your deprecation comment, I think using the factory here would make this method a lot more readable.

test/testUtils.ts
```diff
@@ -0,0 +1,28 @@
+import Color from '../src/util/color';
```
HarelM, line 1: Please move to `test/unit/lib/expect-extensions.ts` or some other file you think has a better name in that folder.

HarelM: `import type Color...`?

jest.setup.ts
HarelM: Thanks, see my new comment about suggested location.

src/util/interpolate.ts
```diff
@@ -10,41 +29,71 @@ export const interpolateFactory = (interpolationType: 'number'|'color'|'array'|'
...
-        number(from.b, to.b, t),
-        number(from.a, to.a, t)
-    );
+function color(from: Color, to: Color, t: number, spaceKey: InterpolationColorSpace = 'rgb'): Color {
+    switch (spaceKey) {
+        case 'rgb': {
+            const [r, g, b, alpha] = array(from.rgb, to.rgb, t);
+            return new Color(r, g, b, alpha, false);
+        }
+        case 'hcl': {
```
HarelM, line 42: [Optional] Small nit picking: I would consider moving this case to be the last one as `lab` is a shorter method which makes it easier to read.

src/util/color_spaces.ts
```diff
@@ -27,111 +41,83 @@ const Xn = 0.950470, // D65 standard referent
...
-    reverse: labToRgb,
-    interpolate: interpolateLab
-};
-
-export const hcl = {
-    forward: rgbToHcl,
-    reverse: hclToRgb,
-    interpolate: interpolateHcl
-};
+    function f(n) {
```
HarelM, line 116: do we have a better name for this inline function by any chance?

CHANGELOG.md
kajkal: Updated.
I based it on the format used earlier in the file I modified
https://github.com/maplibre/maplibre-style/blob/16fc6a7eb84ea5a3c963474da09b3477e666ceb4/CHANGELOG.md?plain=1#L229-L241
If it shouldn't be a separate section maybe it would be a good idea to unify the changelogs entries written for previous versions?

test/testUtils.ts
kajkal: Moved to `test/unit/testUtils.ts`.
Definitely not  `import type Color...`. A js reference to `Color` class is needed.

src/util/interpolate.ts
kajkal: No, in many places where available color spaces are listed they are listed in the order rgb, hcl, lab. I don't see any valid reason to ignore this convention for this function.

CHANGELOG.md
HarelM: Sure, feel free to do so.

test/testUtils.ts
HarelM: File name convention is with "_" I think...

src/util/interpolate.ts
kajkal: Ideally, this `interpolate` object should be exported as:
```ts
export default {
    number,
    color,
    array,
    padding,
};
```
but because this project uses `dts-bundle-generator` which does not handle multiple export default of unnamed object; that is why this rule exists:
https://github.com/maplibre/maplibre-style/blob/16fc6a7eb84ea5a3c963474da09b3477e666ceb4/.eslintrc#L83
all default exports must have a name. If it's because of this repetition `const obj = { ... }; export default obj;` that you don't like the exported object you might consider migrating from `dts-bundle-generator` to for example `rollup-plugin-dts` which, at least in my local attempts, did not have any problems with multiple modules/files with unnamed default object exports.


As for this "factory function"... No, the object must be exported. If you have any doubt then look for `interpolates.number` calls in maplibre-gl-js codebase and imagine how that code would look like if replaced with this `interpolateFactory`.

You may not be aware of it but from TS point of view this function looks like this:
```ts
declare function number(from: number, to: number, t: number): number;
declare function color(from: Color, to: Color, t: number, spaceKey?: InterpolationColorSpace): Color;
declare function array<T extends number[]>(from: T, to: T, t: number): T;
declare function padding(from: Padding, to: Padding, t: number): Padding;
declare const interpolateFactory: (interpolationType: "number" | "color" | "array" | "padding") =>
    typeof array | typeof number | typeof color | typeof padding;
```
I hope I do not have to explain to you why this is wrong and why it is a bad candidate for being part of a library public api.

And no, I don't agree that it's easier to understand what options you have by calling a function that takes the function name as a string, it's easier to understand what options you have by looking at the object and what fields that object has. From my point of view, it's like saying that `consoleFactory('log', 'hello')` is somehow better than `console.log('hello')`.

test/testUtils.ts
kajkal: My bad, corrected.

src/util/interpolate.ts
HarelM: The equivalent is calling `console[operationType].write("do something")` which for me is a shot in the dark when reading this code as there is no good way to know what `operationType` is and no good way to safe guard it, in the example I gave, yes, I prefer `consoleFactory('log').write('hello')` since I can easily understand what I get and what I can do when I read the code.

kajkal: There are plenty of good ways to find out what type `operationType` is, like TS types or type guards, and on top of that both approaches have the same level of protection against an invalid argument, both will simply return undefined :D

Consider the following example ([ts playground version](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABGEBbARgUwE4ApjZyoBcyaW2ANIlHKShjtVPeTgJSuPaIDeAUIiGJsmKCGxICRRAGoaiAFSJctRAFpE01OwDc-AL79+oSLASIAhtmyWAnvkIky3ANoBdZnRcUPzLhScPjgefILCouKSWk4AdKiWAA64uAAm1DDsiAC8AHzBeOk0cK4wnjTseobGEAgAzlCIMGBQOIlwADaWrTl8DBTU1rZ2Bvr8tWANTS1tnd2YAGKW0HDYdr24za3Y7V3mYAAqdomYpADk-ThniAA+iGdD9mdZeWHCiHUA7jBQEAAWKi2sz28EOx0wWQE73eEEsdUw90u2DOpEiEiQSP00OEsPh90edhRIjE6KsNnsWOERlGxgA9LTEOg4TAIIhMAAPSyoRIdTDjeqNIYARl6riF1AATO59BMpkMJaKJdQAGzS-mTRqNbKIAAMsQArGNZY1sCLtUCdnNWksVmtcA9yYT2LhhYMbEqKrpEPSaHU2TZVsR+D6AILYADmaEwLUQcGANHBiLY2A81xgfrAcEFdTqMHDYEs6F5xUQiWsXLEOFj8agiYuycQADICqnYsGGabqwmTgVbi33DUBSIFeaZpa9phYgSXTZxWTsB6oHpvQzaycM0M4J9MKlELUbJhoB07NRsABmLtrhFIjzqqbn3oW3bzVwO4Zndwz01uhfMZc+q8NwDbdd33UQjxPEQLzjbtr2TW922mAA3SwOhgXcC1QBEOS5Hk+T3IdmhQtDUgAOQrXoziI1D0LOGUh2wAAWR8x2fa1lloO1qJI8isKsP1LDAOwqmNER9RY7Y2MwUowGI9DeIROErCE6UVzZdkeRZH49zhRpVlgxBI2sXdRAARxAGBRFSO8TWVCTgRfbj5IopT624a47jfJ41X4KFhFMaBQWmSSrUWDjVgcJ8rVBI4TnOJFnlIK8YMxcIhAC-Zgoc9jbUi1jooQWLTnxR1EtgmCCUpdLwECiwoonG1OLykKQUK8F4uTDySvfSE0uhL4fn+QF8tasETl67FsVxBE3IoIk0WiVLJuhabuqeVESWiSq+veIxdr60TsAAdns8d5kaiL7QJZ4vzneU-y9H1M0+JpGnTWMAGtMDsaogA)):
```ts
const arr1 = [1, 2];
const arr2 = [2, 6];
const t = 0.5;
const r1 = interpolateFactory('array')(arr1, arr2, t); // ts error:
// Argument of type 'number[]' is not assignable to parameter of type 'Color & number[] & number & Padding'.
// r1 of type number | Color | number[] | Padding

const r2 = interpolate.array(arr1, arr2, t); // types narrowed correctly, r2 of type number[]
const r3 = interpolate['array'](arr1, arr2, t); // types narrowed correctly, r3 of type number[]
```
For this example to work properly the `interpolateFactory` function would need to be overloaded for each `interpolationType`:
```ts
export function interpolateFactory(interpolationType: 'number'): typeof number;
export function interpolateFactory(interpolationType: 'color'): typeof color;
export function interpolateFactory(interpolationType: 'array'): typeof array;
export function interpolateFactory(interpolationType: 'padding'): typeof padding;
export function interpolateFactory(interpolationType: 'number' | 'color' | 'array' | 'padding') {
    // switch case
}
```

Also, check out all the places where this factory function is used and see how it is used:
https://github.com/maplibre/maplibre-gl-js/blob/c62155b28818ac5777e4853cffe2892a0ae59487/src/style/properties.ts#L486-L491
https://github.com/maplibre/maplibre-gl-js/blob/c62155b28818ac5777e4853cffe2892a0ae59487/src/style/properties.ts#L545-L550

I don't mean to be intentionally rude, I'm just trying to share my own experience, and I assure you that it's much better to expose in [public API](https://github.com/maplibre/maplibre-style/blob/16fc6a7eb84ea5a3c963474da09b3477e666ceb4/src/style-spec.ts) interpolate object rather than `interpolateFactory` function. `switch case` is not always the answer to all problems! 😜

src/expression/definitions/interpolate.ts
HarelM: If you find a better way to safeguard this that is not a factory, please do so, this code isn't great at it's current state.

kajkal: Done. 
Please keep in mind that I am doing all this for free in my spare time and don't expect me to fix all the shortcomings in the long existing code. Many of the comments I got from you were about improving the code that was here before me. I believe that I did much more in this PR anyway than the scope of my initial proposal would suggest.

kajkal:
I updated the tests in `maplibre-gl-js`, PR [#2376](https://github.com/maplibre/maplibre-gl-js/pull/2376).
And as for the changelog, the changed code is in this repo despite being gl-js specific. Should I duplicate the changelog entries in this repo and in gl-js or leave them only here, what do you think?

HarelM:
If it's ok with you, I'll leave my comments there once this is merged and a new version is published.

HarelM (approved):
THANKS!!!

wipfli:
Thanks for improving the color handling, @kajkal!

Does this change here mean that `hsl(110, 0.7, 0.055)` has to be replaced with `hsl(110, 70%, 5.5%)` in existing style.json documents?

kajkal:
@wipfli yes, that is correct. Former CSS parsing library `csscolorparser` supported this format even though it is not compatible with the CSS specification:
https://github.com/deanm/css-color-parser-js/blob/0d3798690d7e52539f43155a07d3d0fd484e6b8b/csscolorparser.js#L183-L184

Related:
https://github.com/maplibre/maplibre-gl-js/pull/2376#discussion_r1169988109
#135

louwers:
Shouldn't this change be reflected in the style specification (documentation) itself as well?

I am a bit confused by the purpose of this repository now. I thought it was intended to be a platform-independent specification, but looking at this PR it's more evolving to be a library of maplibre-gl-js.

HarelM:
> We should probably put some more thinking about how Native and Web can effectively work together.

I can't argue with that, my initial thought was around having here the validation logic in all the relevant coding languages and not the implementation.
It probably needs a bit more love and care to get to where I imagine it to be...

birkskyum:
> Thanks for improving the color handling, @kajkal!
> 
> Does this change here mean that `hsl(110, 0.7, 0.055)` has to be replaced with `hsl(110, 70%, 5.5%)` in existing style.json documents?

@kajkal , with the new color parser, do people still have to do this replace everywhere, or can we now remove this as a breaking change?

kajkal:
@birkskyum this is still a breaking change. Format `hsl(110, 0.7, 0.055)` is not compliant with the css specification and I believe major release 3.0 is a good time to cut support for it.
I've updated the migration script so these replacements should not be particularly troublesome for those who choose to use version 3.0 and are using this format in their styles.

birkskyum:
Thanks for a swift answer @kajkal - I'll leave the breaking change entry.

PR #97, opened 2023-04-06
Migrate expression tests

birkskyum:
The 358 json expression tests are used in GL Native as well, so it's best they live in this shared repo.

In order to run the integration tests from GL JS, the following geo classes are moved and exported from here:

MercatorCoordinate,
CanonicalTileID,
LngLat,
LngLatLike,

The expression test 'feature_filter' has some geo queries like 'within' so that explains the need for those.

I'm quite sure the .json belongs here, but there are two ways to go about the integration tests.

1. We move the test for each implementations (JS, Native) here, so we can evolve them with the spec, like typescript impl in this case. 

2. We have just the JSON here, and let each implementation import those and run the test like how Native has done so far, but similar for JS. I we go with this, those geo classes might not belong here for good as it would be spec-only.


birkskyum:
@louwers, @ntadej do you have some thoughts on this topic ?

HarelM:
I honestly don't know how I feel about moving these classes here...
I think the json files should be here as they define the tests. The code that does the evaluation also is here, so that doesn't leave a lot of options, doesn't it?

birkskyum:
I guess i does limited the options quite a lot of we want to have the tests close to at least one implementation.

birkskyum:
Currently, as a consequence of removing the maplibre-gl-js submodule, both the [v8.json](https://github.com/maplibre/maplibre-gl-native/tree/main/scripts/style-spec-reference), and all the [300 expression test .json](https://github.com/maplibre/maplibre-gl-native/tree/main/metrics/integration/expression-tests) files are copied into the MapLibre Native repo. It would be great if GL JS and Native could share these.

ChrisLoer:
@birkskyum With https://github.com/maplibre/maplibre-style-spec/pull/146 just merged into `main`, I think the `interpolate-hcl` and `interpolate-lab` tests will need to be updated with new values, since the white point we use in those color spaces has changed to new values. They should be small shifts -- in the unit tests (e.g. `interpolate.test.ts`) we just accepted the new values, while visually inspecting the results with @kajkal 's comparison demo at https://kajkal.github.io/maplibre-gl-style-spec-color-demo/swipe.html

birkskyum:
Let's start by moving just the tests here, release them as artifacts, download and execute them near the respective renderers expression implementations. This is the approach Native takes and GL JS might as well do the same.

HarelM:
@birkskyum can I delete the linked branch or is it still needed?

birkskyum:
feel free to delete it.

Issue #112, opened 2023-04-11
Design Proposal: Update text-offset to better support variable anchor placement


drwestco:
# Design Proposal: Update text-offset to better support variable anchor placement

## Motivation

`text-offset` and `text-radial-offset` both seem tailored to placement of labels around a center-anchored icon. For pushpins, teardrops, or other asymmetrical icons, There's no style syntax to distribute labels evenly around the icon's perimeter.

Existing style properties are all based on the anchor position (yellow circle); since text and icon are independent, and may even be on completely separate layers, there's no correlation with the bounds of the icon.

e.g., if the desired label placements for all eight `variable-anchor` positions are as follows, what's the best way to indicate this through the symbol layer style? The icon is anchored at the bottom (yellow circle indicates anchor), so the desired `text-offset` for `'left'` vs `'bottom-left'` would have a different y value.

![image](https://user-images.githubusercontent.com/743030/231822869-bb86def4-4247-439b-8d9c-49e013cdc39f.png)

The existing style properties have some capability, but are not sufficient to fully cover scenarios like these. In the examples below, to simplify the diagram and explanation, we'll just consider text placement in the four cardinal directions.

Using `"text-radial-offset: 1`, the nearest edge of the text is placed 1em away from the anchor. i.e., when the left or right text anchor is chosen, the offset is outward by 1em in the X direction; when the top or bottom text anchor is chosen , the offset is outward by 1em in the Y direction. Thus `text-radial-offset` is best for circular, square, or other icon shapes where width == height.

![image](https://user-images.githubusercontent.com/743030/231814489-2246c02a-59a9-412b-a309-94f5ee622104.png)

When used with `text-variable-anchor`, `text-offset` gives flexibility to specify the outward shift independently for the X and Y axes. A value like `"text-offset": [1.5, 6]` lets us fix the icon/text overlap when the top slot ("Anchor: bottom") is chosen. Since the offset is always outward relative to the anchor (not the icon center), it adversely affects the bottom slot ("Anchor: top"). So `text-offset` is best for symmetric icons, where the icon is centered over the anchor.

![image](https://user-images.githubusercontent.com/743030/231817496-00c5bd8d-2c2b-4e76-b022-7dc40746268c.png)

[...]

PR #114, opened 2023-04-11
Add offsetCollection type to support text-offset

drwestco:
## Launch Checklist

Adding a new `OffsetCollection` type to support variable type definition for `text-offset` (Issue #112). This type simplifies backwards compatibility by giving a common signature for single point offset values (`[number, number]`), and the new multi-point offsets (`[ [number, number], [number, number] ... ]`).

First time changing the style spec since the project was forked. Let me know what is needed for the Documentation task, if other than updating the `text-offset` property in `v8.json`.

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Can you add the design proposal to the TSC meeting agenda? I think it will be a good place to discuss these proposal.
Once approved we can see how to proceed with code changes.
Thanks for taking the time to open this PR and discussion!

drwestco:
We discussed at last JS meeting, and I updated the issue with clearer examples as requested. What's the next step?

HarelM:
Let's wait for the design approval. 
Also, please see if there's a need to update to docs' site as they were moved here and previous changes to the spec such as new types needed attention of the docs as well (hopefully this will be automated in the future, but it isn't right now)...

src/function/index.ts
```diff
@@ -24,9 +25,17 @@ export function createFunction(parameters, propertySpec) {
     const zoomDependent = zoomAndFeatureDependent || !featureDependent;
     const type = parameters.type || (supportsInterpolation(propertySpec) ? 'exponential' : 'interval');
 
-    if (isColor || propertySpec.type === 'padding') {
-        const parseFn = isColor ? Color.parse : Padding.parse;
+    let parseFn: (input: any) => any;
 
+    if (isColor) {
```
HarelM, line 30: Switch case?

src/reference/v8.json
```diff
@@ -2326,11 +2326,10 @@
...
     },
     "text-offset": {
-      "type": "array",
-      "doc": "Offset distance of text from its anchor. Positive values indicate right and down, while negative values indicate left and up. If used with text-variable-anchor, input values will be taken as absolute values. Offsets along the x- and y-axis will be applied automatically based on the anchor position.",
+      "type": "offsetCollection",
+      "doc": "Offset distance of text from its anchor. Positive values indicate right and down, while negative values indicate left and up. \n\n If used with `text-variable-anchor`, a single point or array containing a single point used for `text-offset` will be taken as absolute values. Offsets along the x- and y-axis will be applied automatically based on the anchor position. \n\n If an array of points is provided for `text-offset`, the number of points should match the number of entries provided via `text-variable-anchor`. Each point in the array is used as the offset distance of text from its anchor when the corresponding anchor is used by the renderer. If more entries are provided for `text-variable-anchor` than `text-offset`, the last value in `text-offset` will be used.",
```
HarelM, line 2330: I didn't understand the part about 
`text-variable-anchor` isn't text-offset enough for placing the labels in different locations?

HarelM:
It feels like there are missing unit tests in places where a case was added for the offset collection...
Also, there are files that do not match snake case I think...

src/reference/v8.json
drwestco: No; they work together. `text-variable-anchor` gives the possible slots to try when placing the anchor. `text-offset` provides the amount to shift the text when one of those slots is considered.

drwestco:
> 

Ah, I guess it should be `validate_offset_collection.ts`, shouldn't it?

src/reference/v8.json
HarelM: Ahh, I think I misinterpreted the design you wrote :-(
I'll re-read it and add my comments there, I have a gut feeling this is better of with a new property - something like text-offset-for-variable-anchors or something, but I need to re-read... I'll keep you posted soon.

drwestco: We did have a discussion on "new property" vs "update existing" for this back in November. Opinions can change, of course..

HarelM: True, I want to re-read those again, sorry for the circular conversation...

drwestco: I thought about it a bit more last night. A new property just for the variable anchor offset could make the implementation and documentation simpler, though we'd then have three properties (text-offset, text-radial-offset, text-variable-offset) that are variations on the same thing.

xabbu42: For whats it worth, I would expect something else under text-variable-offset, that is a list of possible offsets the placer should try (analog but *orthogonal* to text-variable-anchor). This could be a useful separate feature in the future and I would recommend to therefore not use that property name.

drwestco: Harel suggested `text-offset-for-variable-anchors` as an off-the-cuff alternative. That's a little verbose for me. Open to suggestions for something in between.

xabbu42: It just occurred to me that this feature would also be useful if the text-anchor is set by an expression and the current design does not help there. given that, i would either use text-anchor-offsets but with a fixed order independent of text-variable-anchors so it also works with a text-anchor expression, or separate properties text-offset-left, text-offset-right, text-offset-center etc for each possible anchor setting. perhaps with this in mind its better again to just expand text-offset, as now the documentation (and presumably also the implementation) would actually be independent of text-variable-anchor and just use the actual anchor no matter how this anchor was set.

Issue #24, opened 2023-02-26
Export validate from bundle


birkskyum:
It's necessary to export validate (validate-style.ts) from style-spec.ts, as validate is imported and used in i.e. tileserver gl

HarelM:
I might be stating the obvious, but we need to understand first how to export things without increasing the maplibre-gl-js package size.

acalcutt:
I just want to mention I had brought this up to @birkskyum because tileserver-gl has used validate from the npm package. It uses it to [check the styles](https://github.com/maptiler/tileserver-gl/blob/master/src/serve_style.js#L99) that users are hosting.

tileserver-gl Dependabot picked up this update, but out workflow didn't try to merge it because failed our CT tests due validate being missing
https://github.com/maptiler/tileserver-gl/actions/runs/4266913087/jobs/7427971765

HarelM:
I definitely agree that this should be exported, no doubt, but we need to better understand the mechanics of the maplibre-gl-js package size, something is weird there...

HarelM:
It might be, IDK... But we need to find out...

acalcutt:
Just an FYI. like suggested [I tried](https://github.com/maptiler/tileserver-gl/pull/861/commits/0b39db6a2596c023b8e3121164e1a8f1328433a3) to link directly to validate_style like the bin example, but this didn't seem to work
`import validate from '../node_modules/@maplibre/maplibre-gl-style-spec/src/validate_style';`

Is this because that file is typescript extension while tileserver is not? It gets an error it can't find the module

HarelM:
Yes, probably...

HarelM:
See my comments in the linked PR.
Once all the PRs that I have opened are merged I'll be able to publish a new version.
Happy new year!

PR #206, opened 2023-06-06
Reorganize style-spec repository

birkskyum:
The situation around the Style Spec is not as organized as it should be. The spec itself is not versioned, the utilities are including files to maplibre gl js which shouldn't be there, and there are things they can't export because it would blow up gl js bundle size. This PR is the first step towards resolving the situation by isolating the specification .json files, and having everything else, which is implementation code (.ts / .c++ etc.), separate.

The three root folders of the repo are style-spec, docs, utils.

`style-spec` only has .json files, and a CHANGELOG with changes related to the .json files - which parameters are allowed etc. The idea is that these files with be a github release bundle (a .tar / .zip something like that), and that each implementation will pull that in, instead of installing the big style utility npm package.

`docs` contain the documentation for the style-spec, and it does require a bit of implementation code, which from now comes from the utils package, but likely could come directly from maplibre-gl-js instead.

`utils` this is the npm package with style spec utilities, such as validators. Currently this package is included in GL JS, but the intention is to go the other way around, so this package will pull the structures it needs from GL JS instead, so we can expose as many functions as we like from this without increasing the render bundle size. This can solve a longstanding issue #24


utils/package.json
```diff
@@ -44,6 +45,7 @@
     "@mapbox/point-geometry": "^0.1.0",
     "@mapbox/unitbezier": "^0.0.1",
     "@types/mapbox__point-geometry": "^0.1.2",
+    "copyfiles": "^2.4.1",
```
HarelM, line 48: Dev dependency?

utils/src/style-spec.ts
```diff
@@ -173,7 +171,7 @@ export {
 
     ColorType,
     interpolates,
-    v8,
+    latest as v8,
```
HarelM, line 174: I think dts-bundle-generator has problems with exporting `something as somthingelse`...
I would reccomend simply importing it using `import v8 from './style-spec/latest'`.

utils/src/style-spec/latest.ts
```diff
@@ -0,0 +1,3 @@
+
+import spec from './specification.json';
```
HarelM, line 2: Should we use asset type json here or something?

birkskyum:
In regards to this 

> There should be a folder related to running the tests in the maplibre-style-spec-gl/util folder with the relevant logic (may it be jest, puppeteer or what ever we choose in the future).

I think these package should only have the tests that aren't already covered in maplibre-gl, so that would be validation/style-spec conversion tests etc., but not i.e. expression tests

HarelM:
@birkskyum is this PR still relevant? Do we want to push it through or close it?

HarelM:
Closing for now as it looks like it will need a rework anyway.
Feel free to open a new one if this is still relevant.

PR #207, opened 2023-06-06
remove dependency on @mapbox/point-geometry

rotu:
Mapbox has a particular implementation of 2D geometry. Make the style spec only demand an x, y point object.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/expression/index.ts
+++ b/src/expression/index.ts
@@ -28,8 +28,7 @@ import type {Result} from '../util/result';
 import type {InterpolationType} from './definitions/interpolate';
 import type {PropertyValueSpecification} from '../types.g';
 import type {FormattedSection} from './types/formatted';
-import type {Point2D} from "../point2d"
-
+import type {Point2D} from '../point2d';
 
 export type Feature = {
     readonly type: 1 | 2 | 3 | 'Unknown' | 'Point' | 'MultiPoint' | 'LineString' | 'MultiLineString' | 'Polygon' | 'MultiPolygon';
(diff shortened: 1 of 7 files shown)


rotu:
I'm not sure on the `type` vs `interface` choice here. I think interface might provide less friction due to [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) but I'm far from confident.

HarelM:
I think interface is correct since there is a class that implemented it, and I like to have types for data objects that don't have methods, but it might be a personal preference.

rotu:
@HarelM sounds good. I'll switch to `interface`. The main draw of `type` was to be uniform with the rest of the declarations in the file (though likely those declarations should be transitioned over to `interface` too.

src/expression/index.ts
```diff
@@ -41,7 +41,7 @@ export type Feature = {
             'max': string;
         };
     };
-    readonly geometry?: Array<Array<Point>>;
+    readonly geometry?: Array<Array<Point2D>>;
```
HarelM, line 44: I'm not sure what problems this week cause when imported in the main maplibre-gl repo.
Is there a way to use geojson feature here and extend it? The definition of geometry here is incorrect in general I think as fir example for point geometry you only have one a single point, don't you?

HarelM: I hoped we could use a regular geojson instead of array of array of x,y, but that probably won't work...

HarelM (approved):
We will need to make sure this doesn't affect the exported types of maplibre-gl-js

src/expression/index.ts
rotu: @HarelM Why not?

PR #238, opened 2023-06-29
Add coverage to codecov and automerge to dependabot

HarelM:
Add coverage using codecov.

Let's see if this works...

### DIFF
--- a/.github/workflows/automerge-dependabot.yml
+++ b/.github/workflows/automerge-dependabot.yml
@@ -1,0 +1,28 @@
+name: Automerge Dependabot
+
+on: pull_request
+
+permissions: write-all
+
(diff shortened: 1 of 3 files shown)


HarelM:
The report is working:
image

HarelM:
I've also added the integration tests to the coverage report and it seems that we are at about 77%, which is OK.

I've added the auto merge of the dependabot while I'm at it...

birkskyum (approved):

louwers:
> about 77%

The battle is on I guess.

HarelM:
Seems like you're leading 😀
@nyurik you need to catch up...
Screenshot_20230629_181341_Chrome

PR #255, opened 2023-07-07
Trim css color before parsing

kajkal:
Fix for [#2812](https://github.com/maplibre/maplibre-gl-js/issues/2812)

Allow for leading or trailing spaces in css color values.
Treat values like `" rgba(220, 230, 200, 1)"` as valid.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/util/parse_css_color.test.ts
+++ b/src/util/parse_css_color.test.ts
@@ -93,8 +93,8 @@ describe('parseCssColor', () => {
             expect(parse('rgb(0 51 0)')).toEqual(parse('rgba(0, 51, 0, 100%)'));
             expect(parse('rgb(0 51 0)')).toEqual(parse('rgba( 0, 51, 0, 100% )'));
             expect(parse('rgb(0 51 0)')).toEqual(parse('rgba( 00 ,51 ,0 ,100% )'));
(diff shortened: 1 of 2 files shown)


HarelM:
What about other places where there is a space? Like " red "?

kajkal:
If we want to support values like `" rgba(220, 230, 200, 1)"` then I think we should similarly support values like `" red "`. It could easily be made so that only the `rgb()/rgba()/hsl()/hsla()` formats support leading/trailing spaces, but in such a situation I think this would not be consistent behavior - in my opinion it is better to either support this behavior in every format or not support it at all.

Personally, I am not a fan of supporting color values with leading/trailing spaces but if there are established styles that use such values I think we should adapt.

HarelM (approved):

HarelM:
Can you open another PR with the addition of this fix to the changelog?
Forgot to ask it before I merged.
Also can you bump the patch version so I'll be able to easily release a new version with this fix?

kajkal:
I didn't add the changelog entry because I wasn't sure in which repo it should be added (`maplibre-style-spec` or `maplibre-gl-js`).
I opened a simple PR for the 19.2.2 release - #265.

PR #353, opened 2023-09-27
Improve hillshade layers not using raster-dem sources error message

ibesora:
This PR improves the errors around adding `hillshade` layers that don't use a `raster-dem` source.

Before this PR, adding a `hillshade` layer with a `raster` source said `layer \"test\" requires a vector source"` which is not right.
With this PR, the same use case results in the following error: `layer \"test\" requires a raster-dem source`.
It also adds a couple of tests to check that and an extra one that tests a condition that was not met before (using a `raster-dem` source type on layers that are not `hillshade`)

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Let's add this to 19.3.2 as well...

HarelM (approved):

Issue #311, opened 2023-08-22
Change default value for text-font to Noto Sans Regular


bdon:
# Design Proposal: Change the default value of text-font.

## Motivation

The current default `text-font` if no value is specified is `Open Sans Regular, Arial Unicode MS Regular`

https://github.com/maplibre/maplibre-style-spec/blob/d6654968fbbf59dd0ab1dc7c5e4e8d2ae7298081/src/reference/v8.json#L1783

The issues with this current situation:

* The Arial Unicode MS font is proprietary, and users of MapLibre generally do not have a license to this font, making this default useless.
* It requires two fonts for sufficient coverage of the Unicode BMP, since Open Sans has [only Latin, Greek, Cyrillic and Hebrew](https://en.wikipedia.org/wiki/Open_Sans). 

## Proposed Change

Change the default value to `Noto Sans Regular`. This is an [openly licensed font designed for maximum coverage](https://en.wikipedia.org/wiki/Noto_fonts), and can be made available for most MapLibre users. 

## API Modifications

This is a breaking change if users depend on the old default behavior. There shouldn't be many, unless their client is fetching from Mapbox. 

## Migration Plan and Compatibility

This should accompany a major version release of maplibre-gl-js and maplibre-native, because of the change in default value behavior. 

## Rejected Alternatives

* There are other font families like Source Sans and DejaVu with the same goals. Those are viable alternatives.

## Related Issues

https://github.com/maplibre/maplibre-gl-js/issues/2990
https://github.com/maplibre/maplibre-gl-js/issues/1051
https://github.com/maplibre/font-maker/issues/16

HarelM:
I don't have any objections, but there isn't a near future plan to release a major breaking change version.
Version 3 was released a few month ago, so I would aim this for early next year, what do you think?

bdon:
That's perfectly fine, we can revisit this at at a later date - just seems odd to have an unusable default

HarelM:
I can't argue with that...

wipfli:
Good idea in general, but this would be a breaking change in the style specification which I think we should avoid. Or am I missing something?

HarelM:
Technically yes, but if you relay on the default font you are probably doing something wrong.
It is best practice to define the font and make sure you serve it from your server.
Much like specifying the icon and making sure it is in the sprite.
Generally speaking, in most cases, as described here nothing will show up if you don't specify the font, so I think it's better than current situation...

pnorman:
[...]

PR #436, opened 2023-12-04
Change default font-face to Noto Sans Regular [#311]

bdon:
* Change it from Open Sans Regular, Arial Unicode MS Regular which does not have a freely licensed reproducible build.

See https://github.com/maplibre/maplibre-style-spec/issues/311 for discussion.

I don't think that this change is high priority, I'd be fine with keeping the current defaults indefinitely. The consensus seems to be that this default should nearly never be used anyways. 

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -1823,8 +1823,7 @@
       "type": "array",
       "value": "string",
       "default": [
-        "Open Sans Regular",
-        "Arial Unicode MS Regular"
+        "Noto Sans Regular"
       ],
       "doc": "Font stack to use for displaying text.",
       "requires": [


bdon:
While I think that is a better solution, it introduce a new feature to the style, which I find undesirable, and can accomplish it using a preprocessor anyways.

wipfli:
Thanks @bdon for opening this pull request.

This looks like a breaking change in the style specification and I think breaking changes here should be avoided.

Maybe one day in the future we can make a "New MapLibre Style Spec"  and have a bunch of breaking changes there, but until then I would say let's avoid breaking changes.

So I would say let's not merge this one...

bdon:
OK, sounds good to me. Thanks for reviewing.

PR #440, opened 2023-12-05
Upgrade to node 20

HarelM:
Use node 20.10 - current LTS.


louwers (approved):
Why `--no-warnings`?

HarelM:
It says there are experimental loaders flags warning for like 4 versions now, these warnings aren't helpful...

PR #461, opened 2023-12-28
Reduce npm package files

HarelM:
This removes some of the files that currently exist in the npm package for no good reason, I think.
Also fixed some comments from a previous PR.


neodescis:
Seems fine, but is the `files` change actually needed? I only see those things in the package already anyway. Why are we including `src` though?

https://unpkg.com/browse/maplibre-gl@3.6.2/

HarelM:
The link is not for this package 😀
I think adding only the dist is enough, but if anyone would like to see the source that generated it, it might help, IDK...

HarelM:
@neodescis can you approve so I could merge?

neodescis:
> The link is not for this package 😀 I think adding only the dist is enough, but if anyone would like to see the source that generated it, it might help, IDK...

Oh, hahaha, I forgot what repo I was in!

neodescis (approved):

Issue #474, opened 2024-01-05
Freshly built style-spec is incompatible with maplibre-gl-js


sbachinin:
Maplibre-gl-js still imports **operations** from maplibre-style-spec. (https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style.ts#L20C84-L20C95).

But recently operations were removed from style-spec code. (https://github.com/maplibre/maplibre-style-spec/commit/9eaffbd37e1acd53c79a96818dcf6f4c8f2aba62)
This change is not yet reflected in the published style-spec bundle so everything works fine.
But if you re-build style-spec, then maplibre-gl-js will fail to compile because of a broken import of operations.
This makes it hard to make any other changes to style-spec. (There are at least a couple of recent issues in gl-js that require changes in style-spec. I tried to work on them but it turned out impossible).

HarelM:
I have a branch which solves this, you can work on that branch, or wait a bit.
I want to add the sky spec before releasing a new version...

PR #475, opened 2024-01-05
restore export of operations object

sbachinin:
Fixes https://github.com/maplibre/maplibre-style-spec/issues/474

Recently **operations** object was removed from style-spec code but it's still imported by maplibre-gl-js.
Here I restore this object to its shape before commit https://github.com/maplibre/maplibre-style-spec/commit/9eaffbd37e1acd53c79a96818dcf6f4c8f2aba62.

Now all the operations' keys are hardcoded twice. I don't know how to do it DRY. But at least now TS will force you to keep the operations object's keys in sync with the interface.

### DIFF
--- a/src/diff.ts
+++ b/src/diff.ts
@@ -31,6 +31,29 @@ export type DiffOperationsMap = {
 
 export type DiffOperations = keyof DiffOperationsMap;
 
+const operations: Record<DiffOperations, string> = {
+    'setStyle': 'setStyle',
+    'addLayer': 'addLayer',
+    'removeLayer': 'removeLayer',
+    'setPaintProperty': 'setPaintProperty',
+    'setLayoutProperty': 'setLayoutProperty',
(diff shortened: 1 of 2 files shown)


HarelM:
Feels like going backwards, doesn't it? I prefer to remove this unneeded class instead of keeping it...

sbachinin:
Maybe in the grand scheme it's better to remove it, I don't know. It will surely require changes in both style-spec and gl-js.
Now I just want to fix a bug, otherwise it's impossible to make any changes to style-spec

HarelM:
You can create a branch from previous release tag and see that everything is working.
I'll having a hard time relating to the problem, what am I missing?

sbachinin:
I described the problem here
https://github.com/maplibre/maplibre-style-spec/issues/474

HarelM:
Which I have two different valid solutions... Didn't I?

PR #476, opened 2024-01-06
Add expression tests

HarelM:
## Launch Checklist

This adds the expression tests to this repo, as the code related to their functionality is also here.
I've also made some minor modification to the test execution code to be more readable.
After this is merged, I'll remove the tests and their files from the maplibre-gl-js repo.

The issue that prompted it is the following:
- https://github.com/maplibre/maplibre-gl-js/issues/3540

As when I tried to fix it I found out the code and the tests are residing in different repositories.

 - [x] Link to related issues.


HarelM:
When reviewing, the only interesting files are the non json files, mainly:
test/integration/style-spec/validate_spec.test.ts
test/integration/expression/expression.test.ts
package.json and the files added to the lib folder in expression.

louwers:
I didn't know the MapLibre GL JS expression parser lives in this repo.

> As when I tried to fix it I found out the code and the tests are residing in different repositories.

This sounds bad, and the solution to move the tests here sounds like a good fix, but maybe we should also think about the architecture supporting parity between GL JS and Native a bit more. For example, should the expression related code from MapLibre Native also move to this repo? I don't think that is desirable.

A monorepo would be a pretty clean solution, but GitHub does not support it well (the issues and PRs are not separated enough and a single repo has only a single landing page). We are running into that with Native, where developers with vastly different backgrounds are put off by the C++ monster lurking beneath the Android / Qt / iOS platform interfaces that they are familiar with.

Just thinking out loud, but maybe manually copying these tests over to the Native repo (as we do right now) is an OK solution for now... It is better than moving C++ code out of the Native repo in here or having to open two PRs for a single fix.

test/integration/expression/expression.test.ts
```diff
@@ -0,0 +1,183 @@
...
+describe('expression', () => {
+
+    for (const expressionTestFileName of expressionTestFileNames) {
+        test(expressionTestFileName, () => {
+
+            const fixture = JSON.parse(fs.readFileSync(path.join(__dirname, expressionTestFileName), 'utf8'));
+
+            const result = evaluateFixture(fixture);
+
+            if (process.env.UPDATE) {
```
louwers, line 46: Needs some documentation maybe.

HarelM: Good point, I'll look to see if this is mentioned somewhere.

louwers (approved):

HarelM:
This is a big question which we have yet to find a good solution for.
The main reason that there is code here is that there's a npm package that this repo serves that is related to the style spec, that allows validating it. It makes sense to have it in a different repo as it guards from using stuff from the other repo.
The expression tests definition is cross platform I believe, not sure how to solve the sharing part of it though, and also there might be issues with parity when considering the expected results.
In any case, I don't have a good solution, we can talk about it in the next TSC maybe...

birkskyum:
regarding parity between gl js / native, it's still not super clear for me what to do with the .ts code, but moving the json expression tests as done in this PR seems like a natural step in the right direction - and a bit closer to having native also consuming them from here instead of holding a copy.

test/integration/expression/expression.test.ts
HarelM: Done in the contributing guide.

HarelM:
Also @louwers the coverage here is now 92%, I'm totally kicking your butt :-P

PR #477, opened 2024-01-07
fix querying within feature collection of polygons

sbachinin:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/3540

A bit verbose perhaps. But if we need to query stuff within multiple polygons, then Within.geometries has to be an array and therefore some changes are necessary here and there.

Of course this will be thoroughly tested after tests' migration from gl-js.

BTW this bug occurs not only with points but **with lines too**. (This is fixed here too).

### DIFF
--- a/src/expression/definitions/within.ts
+++ b/src/expression/definitions/within.ts
@@ -295,12 +295,24 @@ class Within implements Expression {
         if (isValue(args[1])) {
             const geojson = (args[1] as any);
             if (geojson.type === 'FeatureCollection') {
+                const polygonsCoords = [];
                 for (let i = 0; i < geojson.features.length; ++i) {
-                    const type = geojson.features[i].geometry.type;
-                    if (type === 'Polygon' || type === 'MultiPolygon') {
-                        return new Within(geojson, geojson.features[i].geometry);
+                    const {type, coordinates} = geojson.features[i].geometry;
+                    if (type === 'Polygon') {
+                        polygonsCoords.push(coordinates);
+                    }
+                    if (type === 'MultiPolygon') {
+                        polygonsCoords.push(...coordinates);
                     }
                 }
+                if (polygonsCoords.length) {
+                    const multipolygonWrapper: GeoJSON.MultiPolygon = {
+                        type: 'MultiPolygon',
+                        coordinates: polygonsCoords
+                    };
+                    return new Within(geojson, multipolygonWrapper);
+                }
+
             } else if (geojson.type === 'Feature') {
                 const type = geojson.geometry.type;
                 if (type === 'Polygon' || type === 'MultiPolygon') {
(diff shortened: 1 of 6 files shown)


HarelM:
I've merged the addition of the expression tests.
Please make sure to add the relevant test to the newly added tests.
Also, I think `geometries` in `WithIn` can be a multiploygon, and you can simply add all the polygons to it, I don't think there's a need to change it to be an array of multipolygon.
I think the solution can be simplified by collection all the polygons from the multipolygons and "throwing" it into a single multpoligon and init `WithIn` with it, right?

sbachinin:
Done.
Simplified solution using Multipolygon, as you prescribed.
+ Added tests

src/expression/definitions/within.ts
```diff
@@ -295,12 +295,24 @@ class Within implements Expression {
         if (isValue(args[1])) {
             const geojson = (args[1] as any);
             if (geojson.type === 'FeatureCollection') {
+                const polygonsCoords = [];
                 for (let i = 0; i < geojson.features.length; ++i) {
-                    const type = geojson.features[i].geometry.type;
-                    if (type === 'Polygon' || type === 'MultiPolygon') {
-                        return new Within(geojson, geojson.features[i].geometry);
+                    const {type, coordinates} = geojson.features[i].geometry;
```
HarelM, line 300: I think you can change the for i... to be for let polygon of ...
It would help typescript inference.

src/expression/definitions/within.ts
```diff
@@ -295,12 +295,24 @@ class Within implements Expression {
         if (isValue(args[1])) {
             const geojson = (args[1] as any);
             if (geojson.type === 'FeatureCollection') {
+                const polygonsCoords = [];
```
HarelM, line 298: Can you add a type here? Something like `Position[][][]`

src/expression/definitions/within.ts
```diff
@@ -295,12 +295,24 @@ class Within implements Expression {
...
                 for (let i = 0; i < geojson.features.length; ++i) {
-                    const type = geojson.features[i].geometry.type;
-                    if (type === 'Polygon' || type === 'MultiPolygon') {
-                        return new Within(geojson, geojson.features[i].geometry);
+                    const {type, coordinates} = geojson.features[i].geometry;
+                    if (type === 'Polygon') {
+                        polygonsCoords.push(coordinates);
                     }
+                    if (type === 'MultiPolygon') {
+                        polygonsCoords.push(...coordinates);
```
HarelM, line 305: Seems like this adds an array of ploygons, I would expect to see `flat` somewhere around here...

HarelM:
Did you merge from main? The coverage doesn't report the right number...

HarelM:
Thanks!! Added a few nit picking...

src/expression/definitions/within.ts
sbachinin: No, here polygons are pushed one by one. So it's already flat

sbachinin:
Done
Merged from main

HarelM:
Can you add an item to the changelog?
After that I'll merge and create a new version.

HarelM (approved):
THANKS!!

PR #501, opened 2024-01-18
Fix incorrect `<` char

fynngodau:
I noticed an `&lt;` literal in the docs that should be a `<` literal. This PR fixes that occurrence.

Before | After
---|---
image | image

(This doesn't seem worth mentioning in the changelog.)


HarelM (approved):
Thanks!!

PR #519, opened 2024-02-02
"geometry-type" to identify Multi- features

zstadler:
## What does this PR do

The style spec for [`["geometry-type"]`](https://maplibre.org/maplibre-style-spec/expressions/#geometry-type) says it _"Gets the feature's geometry type: `Point`, `MultiPoint`, `LineString`, `MultiLineString`, `Polygon`, `MultiPolygon`."_ However the implementation for tile-based geometries was identical to that of `$type` and the possible values were `Point`, `LineString`, `Polygon`. This also applied to GeoJSON sources in Maplibre-GL-JS because they are converted and stored internally as tile-based geometries.

The PR adds aligns the implementation of `["geometry-type"]` with its spec and adds the distinction between `Point` and `MultiPoint`, between `LineString` and `MultiLineString`, and between `Polygon` and `MultiPolygon` for tile-based geometries.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


zstadler:
Resolves https://github.com/maplibre/maplibre-gl-js/issues/3516

wipfli:
The result from the poll in https://github.com/maplibre/maplibre-style-spec/discussions/536 is that 8 out of 10 people voted for:

> "geometry-type" will be implemented accord to the spec, this will happen only in a breaking change version (both native and web)"

while only 2 out of 10 voted for:

> Keep the current behavior (duplicate operators), update the spec accordingly, and introduce a new operator to get "multi"s

I am in favor of the second option with the simple reason that we do not have a good way of informing our users that there has been a breaking style spec change. Up until today, when I as a map designer make a style sheet there have been two options: 

* either the style will render identically in MapLibre GL JS and MapLibre Native and it does not matter which version I use
* or the version of GL JS or Native will not support a feature I am using and therefore the who rendering will crash.

If we start introducing breaking changes in the style specification we will have to add a third point to this list of options which is:

* how your map will look like depends on the version of MapLibre you are using

There are plenty of bad things in the MapLibre Style Specification and I would love to through away entirely things like `"{name}"` and replace it everywhere with `"["get", "name"]"`. But as ugly as the style spec is in many places, it is also an incredible achievement of this project and all the people who worked on it to get a spec that has been without breaking changes over the last 8 years or so. There is a lot of value in stability and I think we should keep following this path.

HarelM:
I generally agree, except that this case is different - $type is used in most cases and there's no good way to get a multi geometry type. I believe that this case makes sense. The community had voted so.
The way to tell users, much like other breaking changes, is in the release notes of the relevant breaking changes version (5.x of malibre-gl in this case probably).
We had other changes that changed how the map looked due to bug fixes, I think this case is very similar in that aspect that it feels more like a bug fix and not an actual spec change...

wipfli:
Thanks for the feedback, @HarelM. I doubt that we actually have a good way of knowing if people use `$type` more often than `geometry-type`. But even if it is less often used, it does not mean that changing its behavior is not important. 

I remember we did have some other breaking change in the style spec where we fixed something with colors, do you remember what the other breaking changes were?

Fixing bugs in the style spec sounds attractive, but keep in mind how old this toolkit now is and how many people rely on it. It is like with any other old library. People know the bugs and the expect them to be there. If you fix the bugs, they will say why did you change the behavior - the old way was fine for us and now we have extra work because you made the code look better.

Maybe I am too conservative and we can start to be more dynamic with how we handle the spec, I don't know, but I feel like it would be much better for our users to freeze the behavior and leave bugs and bad aesthetics where they are...

zstadler:
None of the [styles available in Maputnik](https://github.com/maplibre/maputnik/blob/main/src/config/styles.json) uses `["geometry-type"]`. I was looking for styles on [maplibre.org home page](https://maplibre.org/), and found none.

HarelM:
@zstadler any chance you can update this PR with the latest changes?
I think some stuff were moved but I'm not sure.
I'm looking into merging this as the release of version 5.0 of maplibre is no in motion.

src/expression/evaluation_context.ts
```diff
@@ -29,8 +39,59 @@ class EvaluationContext {
         return this.feature && 'id' in this.feature ? this.feature.id : null;
     }
 
+    // Bluntly copied from
+    // https://github.com/mapbox/vector-tile-js/blob/77851380b63b07fd0af3d5a3f144cc86fb39fdd1/lib/vectortilefeature.js#L225-L233
+    _signedArea(ring) {
```
HarelM, line 44: Is it possible that some of the methods added here already exists in the following file: 
https://github.com/maplibre/maplibre-style-spec/blob/63b660a03a444c358d8d363d578e8bf29e2a8c3d/src/util/classify_rings.ts#L56

HarelM:
Is $type a deprecated expression type that can't be used in "newer" expressions?

src/expression/evaluation_context.ts
zstadler: Done in [cbdd941](https://github.com/maplibre/maplibre-style-spec/pull/519/commits/cbdd9418c3c56469e56d297e15f6be946bbe5511)

src/feature_filter/feature_filter.test.ts
```diff
@@ -329,9 +329,81 @@ function legacyFilterTests(createFilterExpr) {
     });
 
     test('==, $type', () => {
-        const f = createFilterExpr(['==', '$type', 'LineString']).filter;
-        expect(f({zoom: 0}, {type: 1})).toBe(false);
-        expect(f({zoom: 0}, {type: 2})).toBe(true);
+        const fPoint = createFilterExpr(['==', '$type', 'Point']).filter;
```
HarelM, line 332: Can this test be split into a few different tests to indicate what is being tested?

src/feature_filter/convert.ts
```diff
@@ -128,7 +128,7 @@ function runtimeTypeChecks(expectedTypes: ExpectedTypes): ExpressionFilterSpecif
 function convertComparisonOp(property: string, value: any, op: string, expectedTypes?: ExpectedTypes | null): ExpressionFilterSpecification {
     let get;
     if (property === '$type') {
-        return [op, ['geometry-type'], value] as ExpressionFilterSpecification;
+        return convertInOp('$type', [value], op === '!=');
```
HarelM, line 131: Can you elaborate a bit on this change and why it is needed?

src/expression/evaluation_context.ts
```diff
@@ -29,8 +40,47 @@ class EvaluationContext {
...
+                    switch (geometryType) {
+                        case 'Point':
+                            geometryType = 'MultiPoint';
+                            break;
+                        case 'LineString':
+                            geometryType = 'MultiLineString';
+                            break;
+                        case 'Polygon':
+                            // Following https://github.com/mapbox/vector-tile-js/blob/77851380b63b07fd0af3d5a3f144cc86fb39fdd1/lib/vectortilefeature.js#L197
+                            for (let i = 0, ccw; i < len; i++) {
```
HarelM, line 66: Can this part move to classify rings file? I think it has similar code and I think it would be better to have them side-by-side.

src/expression/evaluation_context.ts
```diff
@@ -29,8 +40,47 @@ class EvaluationContext {
...
+                                }
+                            }
+                            break;
+                        default:
+                            break;
+                    }
+                }
+            }
+        }
+        return geometryType;
```
HarelM, line 83: Should this response be cached in order to avoid recalculating this? Or is it OK to calculate this each time as it's not really possible to cache anything?

src/expression/evaluation_context.ts
```diff
@@ -29,8 +40,47 @@ class EvaluationContext {
...
+        let geometryType = this.feature.type;
+        if (typeof geometryType === 'number') {
+            geometryType = geometryTypes[this.feature.type];
+            if (geometryType !== 'Unknown') {
+                const geom = this.geometry();
+                const len = geom.length;
+                if (len > 1) {
+                    switch (geometryType) {
+                        case 'Point':
+                            geometryType = 'MultiPoint';
```
HarelM, line 59: Is it possible to simply return the relevant value? I think it would simply the method a bit...

src/feature_filter/convert.ts
zstadler: That's the essence of the PR.
The `Multi` and non-`Multi` variants of `geometry-type` must be considered when comparing with the fewer possible values of `$type`. Later on, `==` and `!=` expressions are transformed into `in` and `!in` expssions.
For example, `["``",  "$type", "LineString` ]` becomes `["in", ["geomerty-type"], ["LineString", "MultiLineString"]]`

HarelM: Doesn't this mean that it will convert `geomerty-type` to "old style" expression and there won't be a way to use it with new expressions?

src/feature_filter/feature_filter.test.ts
```diff
@@ -328,10 +328,88 @@ function legacyFilterTests(createFilterExpr) {
...
+
+    test('==, $type, LineString', () => {
+        const fLineString = createFilterExpr(['==', '$type', 'LineString']).filter;
+        expect(fLineString({zoom: 0}, {type: 0})).toBe(false);  // Unknown geometry-type
+        expect(fLineString({zoom: 0}, {type: 1, geometry: [[{x: 0, y: 0}]]})).toBe(false);      // Point geometry-type
+        expect(fLineString({zoom: 0}, {type: 1, geometry: [[{x: 0, y: 0}], [{x: 0, y: 0}]]})).toBe(false);      // MultiPoint geometry-type
+        expect(fLineString({zoom: 0}, {type: 2, geometry: [[{x: 0, y: 0}, {x: 1, y: 1}]]})).toBe(true); // Linestring geometry-type
+        expect(fLineString({zoom: 0}, {type: 2, geometry: [[{x: 0, y: 0}, {x: 1, y: 1}], [{x: 2, y: 0}, {x: 1, y: 0}]]})).toBe(true); // MultiLineString geometry-type
+        expect(fLineString({zoom: 0}, {type: 3, geometry: [[{x: 0, y: 0}, {x: 3, y: 0}, {x: 3, y: 3}, {x: 0, y: 3}, {x: 0, y: 0}],
+            [{x: 0, y: 0}, {x: 0, y: 2}, {x: 2, y: 2}, {x: 2, y: 0}, {x: 0, y: 0}]]})).toBe(false);     // Polygon geometry-type
```
HarelM, line 360: The comments here are also probably an indication of a further need to split tests I guess. 
Better to have a string for test name instead of a comment if you'd ask me.

src/expression/evaluation_context.ts
```diff
@@ -29,8 +40,41 @@ class EvaluationContext {
...
+        const geom = this.geometry();
+        const len = geom.length;
+        if (len === 1) {
+            return geometryType;
+        }
+        if (geometryType !== 'Polygon') {
+            return `Multi${geometryType}`;
+        }
+        // Following https://github.com/mapbox/vector-tile-js/blob/77851380b63b07fd0af3d5a3f144cc86fb39fdd1/lib/vectortilefeature.js#L197
+        for (let i = 0, ccw; i < len; i++) {
```
HarelM, line 67: Can this "for" move to classify rings file?

src/util/classify_rings.ts
```diff
@@ -69,3 +69,27 @@ export function calculateSignedArea(ring: Point2D[]): number {
     }
     return sum;
 }
+
+/**
+ * Returns if there are multiple outer rings.
+ * The first ring is a outer ring. Its direction, cw or ccw, defines the direction of outer rings.
```
HarelM, line 75:
```suggestion
 * The first ring is an outer ring. Its direction, cw or ccw, defines the direction of outer rings.
```

HarelM: > defines the direction of outer rings

defined the direction of the other outer rings?

src/feature_filter/convert.ts
HarelM: Apparently it upgrades the expression to new style. So no problem here.

HarelM:
I've added some minor comments around the wording in the comments, looks good otherwise, thanks!

src/expression/evaluation_context.ts
HarelM: I'm not aware of such mechanism, besides maybe storing it in the geometry itself...? IDK...

src/expression/evaluation_context.ts
```diff
@@ -29,8 +40,33 @@ class EvaluationContext {
         return this.feature && 'id' in this.feature ? this.feature.id : null;
     }
 
+    geometryDollarType() {
+        return this.feature ?
+            typeof this.feature.type === 'number' ? geometryTypes[this.feature.type] : simpleGeometryType[this.feature.type] :
+            null;
```
HarelM, line 46: According to coverage report, `null` is never returned. Is it possible that it is dead code? I can't think of a way to the code to reach it...

src/expression/evaluation_context.ts
zstadler: I'd rather not introduce a new optimization mechanism for a problem we don't know that has a significant impact.

src/expression/evaluation_context.ts
zstadler: The `null` was in the original expression.  I guess it is a safety net in case the method is called when the context object has no feature.

The coverage report is a result if a new line break. I'm OK with a longer line, if you prefer.

HarelM: I think that if something is wrong it will return a null before this null return value, so I think it's safe to remove it.

test/integration/expression/tests/geometry-type/MultiLineString/test.json
HarelM, line 1: Sorry for nit-picking, can the folder name be kebab case like other tests? or at least lower characters?

src/expression/evaluation_context.ts
HarelM: Never mind, My mistake, null is for when the feature is null...

HarelM (approved):

PR #563, opened 2024-03-11
Docs: Remove "support table" at the end of deprecations

tordans:
At the end of the Deprecations page (https://maplibre.org/maplibre-style-spec/deprecations/) there is this support table:
> image

I don't understand why that would be there talking about the support of `has / !has` which is not what the section "[Combining filters](https://maplibre.org/maplibre-style-spec/deprecations/#combining-filters)" is about.

I am assuming someone forgot to remove it from the page at some point…

This PR removes the table.

## Launch Checklist

### DIFF
--- a/docs/src/routes/deprecations/index.tsx
+++ b/docs/src/routes/deprecations/index.tsx
@@ -304,24 +304,6 @@ The combining filter "all" takes the three other filters that follow it and requ
 ]
 \`\`\`
 `}/>
-
-            <SDKSupportTable
-                supportItems={{
-                    'basic functionality': {
-                        js: '0.10.0',
-                        android: '2.0.1',
-                        ios: '2.0.0',
-                        macos: '0.1.0'
-                    },
-                    '\`has\` / \`!has\`': {
-                        js: '0.19.0',
-                        android: '4.1.0',
-                        ios: '3.3.0',
-                        macos: '0.1.0'
-                    }}}
-
-            />
-
         </div>
     );
 }


HarelM:
The original docs has this as well, while I'm not 100% sure why, it is there:
https://docs.mapbox.com/style-spec/reference/other/

So it might indicate which version of mapbox/libre supports these old type filters.

HarelM:
@tordans any response to my above comment? Where are we with this PR?

tordans:
> The original docs has this as well, while I'm not 100% sure why, it is there: https://docs.mapbox.com/style-spec/reference/other/
> 
> So it might indicate which version of mapbox/libre supports these old type filters.

I assumed that it is either an error in Mapbox's docs or an artifact of some layout system that requires one table ans someone just put `has` in there so the page works. Did not check their docs system, though.

Anyways, I don't understand what it is supposed to tell me in this place. Why is there a table about `has / !has` there? What does `basic functionality` mean in this context of a headline that is "Combining filters" which is about `all`, `any`, `none`.

It's just confusing, so I think it should be removed.

HarelM:
If I understand correctly, "Other filters" is what this table is talking about - i.e. `in`, `!in` and the other that are specified in this part of the docs.
"basic functionality" I assume is all the other except `has, !has` which was added in a later version.
This is my understanding of this table in the docs, and it is aligned with other part of the docs explaining about a feature and then adding this table to indicate which versions supports what.
So I'm not eager to remove this table, but if you think there's a way to make this less confusing, feel free to suggest.

HarelM:
I'm rewriting the docs, so this will not be relevant.
I can understand the reasoning behind this table, so I'll keep it for now.
If this is important to you feel free to update this PR or create a new PR against the new docs.

Thanks for taking the time to open this PR!

PR #565, opened 2024-03-12
add .nojekyll

birkskyum:
I believe this is what we need to disable jekyll:

https://docs.github.com/en/enterprise/2.14/user/articles/files-that-start-with-an-underscore-are-missing
https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/


HarelM:
@birkskyum this should be added to the build as well somehow, otherwise, in the next merge to main it will break again.
Can you take care of it?

HarelM:
I think this is described here:
https://dev.to/lexlohr/deploy-a-solid-start-app-on-github-pages-2l2l
the `touch .output/public/.nojekyll` part.
image

PR #623, opened 2024-04-23
Add contour source

### LINKED ISSUE #583: Design Proposal: Contour Line Source from Raster DEM Tiles
# Design Proposal: Contour Line Source from Raster DEM Tiles

## Motivation

Give users a built-in way to render contour lines in maplibre-gl from the same DEM tiles that are already used for terrain and hillshading, like this:

image

## Proposed Change

Create a new `contour` source type in maplibre style spec that takes a raster-dem source as input and generates contour isolines as output that can be styled using line layers, for example:

```js
sources: {
  dem: {
    type: "raster-dem",
    encoding: "terrarium",
    tiles: ["https://elevation-tiles-prod.s3.amazonaws.com/terrarium/{z}/{x}/{y}.png"],
    maxzoom: 13,
    tileSize: 256,
  },
  contours: {
    type: "contour",
    source: "dem",
    multiplier: 3.28084, // feet
    thresholds: {
      // zoom: [minor, major]
      11: [200, 1000],
      12: [100, 500],
      13: [100, 500],
      14: [50, 200],
      15: [20, 100],
    },
    // overzoom z13 DEM tiles up to z16 to generate smooth high-zoom contour lines:
    maxzoom: 16,
    // overzoom z10 tiles to generate z11 contour lines, z11 to make z12, etc...
    overzoom: 1,
    // customize the tags on output vector line features
    elevationKey: "ele",
    levelKey: "level",
    contourLayer: "contours",
  },
},
```

This offloads details about how to retrieve and parse DEM tiles to the DEM source definition, and gives style layers the flexibility to render any number of visible lines derived from that contour source.

I've already prototyped this in the [maplibre-contour plugin](https://github.com/onthegomap/maplibre-contour), here are some of the issues I had to work through to get these contours to look nice:

#### DEM "overzooming" (smoothing)

The contour lines look blocky when you zoom in much further past the maximum zoom for a DEM source, but they can look nice and smooth if you "overzoom" the DEM tiles by applying iterative bilinear interpolation before generating isolines. For example for onthegomap I use 512px z11 tiles, but overzoom the z11 tiles up to z15 so that the [contour lines look smooth even at high zooms](https://onthegomap.com/?m=t&u=mi&lat=44.27038&lng=-71.30339&zoom=17.1). This is why the proposal lets you specify a `maxzoom` that is higher than the `maxzoom` of the raster-dem source.

Also to generate smooth contour lines at the border between tiles, the algorithm needs to look at adjacent tiles. This means you need 9 DEM tiles to render a single contour tile. To mitigate this, the `overzoom` parameter lets you use overzoomed DEM tiles from a lower zoom level to generate contours at the current zoom level, for example `overzoom=1` means use the top-left, top-right, botom-left, or bottom-right z10 tile to render a z11 contour line tile. This means you only need 4 DEM tiles to render a single contour tile:

image

#### Contour levels and units

The user needs to be able to choose what elevations to draw contour lines at, which changes by zoom level. They may also designate "major" and "minor" levels, for example to generate thin contour lines every 200m but bold every 1000m at z11 and z12, and then thin contours every 100m but bold every 500m >= z13:

```js
thresholds: {
  11: [200, 1000],
  13: [100, 500]
},
```

This puts a `level=0` property on each regular line feauture, but `level=1` every multiple of 1000m (or 500m >= z13) that line layers can use to style differently. You could also add more levels if you wanted to style differently based on `level=2`, `level=3`, etc.

The `multiplier` attribute multiplies raw meter values by a certain amount to change the unit, for example `multiplier=3.28084` changes from meters to feet. When you click the distance indicator on [onthegomap](https://onthegomap.com/?m=t&u=mi&lat=44.27038&lng=-71.30339&zoom=17.1), it toggles between `multiplier=1` and `multiplier=3.28084`.

#### Performance and Bundle Size

I've already implemented the [smoothing logic](https://github.com/onthegomap/maplibre-contour/blob/main/src/height-tile.ts) and [isoline generation](https://github.com/onthegomap/maplibre-contour/blob/main/src/isolines.ts) in the maplibre-contour plugin so we would just need to bring that into maplibre-gl-js and port into the native projects. The overall plugin is 33kb (11kb gzipped) but most of that is replicating the web worker communication, cancelable message passing, and vector tile encoding that maplibre-gl-js already has. The actual business logic is only 3.7kb (1.6kb gzipped).

The isoline generation algorithm was derived from [d3-contour](https://github.com/d3/d3-contour) but is much more efficient because it generates isolines in a single pass through the DEM tile instead of using a pass per contour. For onthegomap users on a range of devices (mostly mobile phones) generating a 512px contour tile takes:

- &lt;10ms 40% of the time
- &gt;50ms 10% of the time
- &gt;100ms 2% of the time
- &gt;200ms 0.5% of the time
- &gt;500ms 0.05% of the time
- &gt;1s 0.006% of the time

## API Modifications

This should only change the style spec, but shouldn't require any js or native API changes.

This is new functionality, so no migration is necessary.

### Build a plugin for this

I maintain the [maplibre-contour plugin](https://github.com/onthegomap/maplibre-contour) which already lets you do this by using the `addProtocol` integration, but it has a few downsides:

1. It's an extra step to install: rendering contour lines is a common use-case that users should be able to do by default
2. 90% of the plugin is duplicating things that maplibre already does like spawning a web worker, communicating with it using cancelable messages, and decoding DEM tiles. The actual code for computing the contours is a small fraction of the overall plugin.
3. It has to do a wasteful extra step of encoding the result as vector tile bytes only for maplibre to decode immediately after in its own web workers (see https://github.com/onthegomap/maplibre-contour/blob/main/architecture.png)
4. It can't make use of other registered maplibre request interceptors or protocols like DEM tiles served out of a pmtiles archive
5. It doesn't work in maplibre-native

### Pre-render contour lines

You can render contour line vector tiles ahead of time and serve those for the planet, this will save some browser CPU cycles but rendering them on the fly from DEM tiles has a few advantages:

1. There are a lot of parameters you can tweak when generating contour lines from elevation data like units, thresholds, and smoothing parameters. Pre-generated contour vector tiles require 100+gb of storage for each variation you want to generate and host. Generating them on-the-fly in the browser gives infinite control over the variations you can use on a map from the same source of raw elevation data that maplibre uses to render terrain and hillshade.
2. You're likely already downloading DEM tiles for hillshading and terrain, so this eliminates the extra bandwidth used to download those vector tiles.

### Implement as a new layer type

We could implement this as a new layer type instead of a source type, but that would tightly couple the display parameters to the logic for how contour lines are generated, and potentially require us to generate the contours in multiple passes over the source DEM data. It seems cleaner to separate this into 2 steps: generate contour lines, then style visible layers from them.

### Take a DEM tile source URL as input

A contour layer could take as input `tiles: ["server.com/{z}/{x}/{y}.png"],` but there are a lot of different knobs to tune for how these are interpreted, so by depending on a DEM source we re-use the DEM source control all of those parameters.

msbarry:
Add a new source type that generates contour isolines from raster-dem sources.

Fixes #583 - see that issue for the full design and discussion.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.
 - [ ] Major vs. index terminology?
 - [x] Use expression for interval and major/index that only support `["zoom"]` as an argument
 - [ ] should updating a raster-dem source used by a contour source cause remove/add on both sources?


HarelM:
This looks great overall, the only question left is the expression fir the intervals and multiplier.
Thanks!!!

build/expressions_patrial.md
```diff
@@ -0,0 +1,136 @@
+# Expressions
+
+The value for any `layout property`, `paint property`, or `filter may be specified as an _expression_. An expression defines a formula for computing the value of the property using the _operators_ described below. The set of expression operators provided by MapLibre includes:
```
zstadler, line 3: please use \`filter\` rather than \`filter

HarelM:
@msbarry any updates on this? Is there something I can help with?

msbarry:
I was planning to keep this open as I implement the maplibre-gl-js side of it in case I have to make any changes here.  I got a bit sidetracked trying to get a planetiler release out with parquet/overture support but should have that wrapped up soon so I can get back to this.

My in-progress changes to maplibre-gl-js are here: https://github.com/msbarry/maplibre-gl-js/tree/contour-source - I have it working but it's pretty inefficient (overfetching a lot of tiles) and doesn't handle adding border pixels yet. Need to do some more thinking about how to handle those... It might make sense to refactor source_cache first to decouple staging the tile pyramid for a source consumer from fetching/caching tiles if you had any ideas there? I think if terrain and hillshading could read from the same source, then it should be ready for contour tiles to read from it as well.

HarelM:
No rush, just wanted to make sure you haven't forgot about it 🙂 
I don't have a good perspective about source cache ATM in order to retractor it, so you'll probably beat me to it...

HarelM:
@msbarry where are we with this?
We are in the middle of 5.x version of maplibre, so this might be a good opportunity to break things if they are needed as part of this feature.

msbarry:
I'm still hoping to do this eventually, but it's gotten bumped further down my todo list in the short term 😞   I had been thinking of this as a purely additive thing. The only change to existing functionality would be if I can fix source cache and remove the need to define the same DEM source twice to use across terrain and hillshading, but that's just  a convention change, not breaking.  Was there anything you were thinking.

HarelM:
I'm not sure what we need to break if at all in order to implement this feature. I recall stuff that might be related to source cache, although source cache is not a public API, but still.
If the feature doesn't require any changes, we can introduce it later on, in a minor release.
But since you looked at the code to implement it, you might have more insights as to what needs to change in order to implement this smoothly.
In any case, I hope the following PR would allow for most of what you need:
https://github.com/maplibre/maplibre-gl-js/pull/4779

PR #626, opened 2024-04-24
Add mkdocs

### LINKED ISSUE #562: Add Native specific note to GeoJSON source
> The GeoJSON document must be on the same domain or accessible using [CORS](http://enable-cors.org/).

CORS does not apply to MapLibre Native. I think it can also read GeoJSON from the file system

### LINKED ISSUE #604: The site needs search
# Design Proposal: Add search to docs site.

## Motivation

All good documentation sites include search.

## Proposed Change

Add a search toolbar.

## API Modifications

Possibly front end only but a great example are the godot docs.

No current search.

Make a statement about the compatibility of your new functionality with existing functionality.

### LINKED ISSUE #618: Add docs how to serve docs locally
I ran

```
cd docs
npm install
npm run dev
```

But when I navigate to `http://localhost:3000/` I see 

```
{
  "statusCode": 404,
  "statusMessage": "Cannot find any path matching /.",
  "stack": []
}
```

We should add a readme at `docs/README.md`.

@birkskyum Could you help me out?

http://localhost:3000/maplibre-style-spec also does not work

HarelM:
## Launch Checklist

Fixes #618
Fixes #604
Fixes #562

The main changes in the PR are adding a generate-docs.ts that generates the markdown files (md).
And adding the ability to spin off the mkdocs container (npm script mainly).

For expressions, I have removed the "Sytnax" part for now, we should consider adding an example to each expression instead of the syntax.

This is how the initial docs now look:
There are still places where there's a need to add more content.
image


birkskyum:
Remember the "Expressions" page. It's arguably the most complex

HarelM:
Sure, thanks for the tip!

HarelM:
This is now ready for review.
Don't be alarm on the number of files, most of them are deleted files.
It doesn't look exactly the same but I believe the content is very close to be identical.
There are less images in the generated files, and I tried to improve the v8 examples and doc comments to allow a more "single point of truth".
I have removed the syntax part of the expression page as I wasn't sure how it was created, and I tend to say the I think using examples in the v8.json would be a better approach as most expressions (if not all) do not have examples.

Any feedback is welcome, I think the simplicity here outweighs the stuff that are less polished.

birkskyum:
I see a lot of this, when I try to run the container
Screenshot 2024-04-25 at 10 42 41

birkskyum:
if possible, i'd like to be able to click a heading, to get the link.
Screenshot 2024-04-25 at 10 45 41

HarelM:
Make sure to delete the node_modules from the docs folder - this is what causing the errors.
Not sure how I can get a link from clicking the heading, one can do that by clicking the right side nav as a workaround.

birkskyum:
> I have removed the syntax part of the expression page as I wasn't sure how it was created, and I tend to say the I think using examples in the v8.json would be a better approach as most expressions (if not all) do not have examples.

I actually think these syntax are the examples from the v8.json extracted

HarelM:
I see that there is some missing content from the sources as it is not present in the v8.json file, I'll try and find an elegant way to add this.

HarelM:
> I actually think these syntax are the examples from the v8.json extracted

Can you give me a pointer? I was not able to find it...

HarelM:
Here for example is the `let` defintion in v8:
```
"expression_name": {
    "doc": "",
    "type": "enum",
    "values": {
      "let": {
        "doc": "Binds expressions to named variables, which can then be referenced in the result expression using [\"var\", \"variable_name\"].",
        "group": "Variable binding",
        "sdk-support": {
          "basic functionality": {
            "js": "0.41.0",
            "android": "6.0.0",
            "ios": "4.0.0",
            "macos": "0.7.0"
          }
        }
      },
```

birkskyum:
right, it comes from here
https://github.com/maplibre/maplibre-style-spec/blob/85a63c49c471c534cbd1af3c02e075feb1e94f69/docs/src/expression/expression-metadata.tsx#L226-L234

birkskyum:
> Not sure how I can get a link from clicking the heading, one can do that by clicking the right side nav as a workaround.

Maybe this can add the links https://mkdocstrings.github.io/autorefs/

HarelM:
I've added permalinks to toc plugin, which is what I found in the docs, seems to work as expected.

HarelM:
I've also added that the toc will follow the scrolling of the page now.

birkskyum:
I guess the `syntax` always was JS-centric anyway, considering the types used in the examples - I assume the syntax examples doesn't map directly to Native

birkskyum (approved):

build/generate-docs.ts
```diff
@@ -382,6 +382,9 @@ function createExpressionsContent() {
             }
             content += `\n### ${key}\n`;
             content += `${value.doc}\n`;
+            if ('example' in value) {
+                content += '\n`"some-propetry": ' + JSON.stringify(value.example) + '`\n';
```
birkskyum, line 386: propetry, typo

HarelM: Thanks! I probably have a lot more typos here :-)
We'll fix them eventually...

HarelM:
Right! Thanks for spotting this!

birkskyum:
There seem to be other missing information and links too

before:
Screenshot 2024-04-25 at 13 20 52

after:
Screenshot 2024-04-25 at 13 20 45

HarelM:
Added requires, feature-state, type (paint, layout) and fixed some broken links:
image

Let me know if there are other properties I might have missed.

birkskyum:
Screenshot 2024-04-25 at 14 00 36

can the link to `feature-state` and to `interpolate` be added again?

birkskyum:
I noticed the "Related" section was gone, just if it was unintentional.

Screenshot 2024-04-25 at 14 29 21
Screenshot 2024-04-25 at 14 29 29

HarelM:
I see that this information is stored in https://github.com/maplibre/maplibre-style-spec/blob/main/docs/src/expression/expressions-related.json
:-(

Not sure I want to copy it...
Maybe I can get examples from the linked pages and kill two birds in one stone...

HarelM:
Ok, so I think all the issues are fixed, the only things remaining are the expression links and expression syntax that I plan to solve by simply adding examples on how to use all kind of expression based on the linked examples.
I can add the examples to the v8 docs section for every expression, but I'm not sure it won't be too verbose...
Let me know what you think.

birkskyum:
I think these examples hold some value, so it would be good to have them in the spec.

louwers:
For me it's parsing through all the Markdown in `node_modules`

louwers:
Added social media previews:

image

An SVG logo (previous one was low res) and MapLibre colors:

image

Also added additional social media options:

louwers (approved):
I love it! 

I think we should split things up a bit more in an API reference and guides. But that can be addressed in the future.

Please look over my changes, but looks good to go. 👍

HarelM:
Thanks @louwers!
Do you see value in actively removing the TOC?
In mobile it's not present anyway and in desktop there's plenty of room. 
Just nit picking.
I'll go over the links to the examples, hopefully add some links and examples to the expressions and merge this. 
@louwers please let me know if any part of what I wrote is not clear as I would like you to be able to fix stuff in this if needed, although it turned out to be very very simple and straight forward, so I hope anyone would be able to fix stuff now.

birkskyum:
It's really nice to have toc on desktop i think

birkskyum (approved):

birkskyum:
> I think we should split things up a bit more in an API reference and guides. But that can be addressed in the future.

This sounds like a very good idea. The images that were removed from the layers section really belongs more in a guides-section.

louwers:
The TOC is a very nice to have, but on the Introduction page and the changelog it was not helpful in my opinion.

build/generate-docs.ts
```diff
@@ -0,0 +1,498 @@
+import v8 from '../src/reference/v8.json' assert { type: 'json' };
+import fs from 'fs';
```
louwers, line 2: Having a look over this now.

package.json
```diff
@@ -20,6 +20,9 @@
     "build": "rollup --configPlugin @rollup/plugin-typescript -c rollup.config.ts && cp ./src/reference/v8.json ./dist/latest.json",
     "generate-style-spec": "node --no-warnings --loader ts-node/esm build/generate-style-spec.ts",
     "generate-typings": "node --no-warnings --loader ts-node/esm build/generate-typings.ts",
+    "docs": "node --no-warnings --loader ts-node/esm build/generate-docs.ts",
+    "start-docs": "docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material",
```
louwers, line 24: Maybe we can have it rebuild the docs automatically when `v8.json` changes.

birkskyum:
aaah, i see. yeah, if it was only the version number it could have been sane, but I agree that like this it's too verbose 
Screenshot 2024-04-25 at 22 59 57

package.json
HarelM: Maybe, I'm not sure how, I simply run `npm start-docs` in one terminal and `npm run docs` after I change the v8.json file.
Feel free to investigate...

louwers:
I am going over the build script now to make sure I understand everything.

HarelM:
My last commit added example to expressions which has links and added the links to the v8.json.
From my point of view this can be merged now.

package.json
louwers: I found out how. :) Will push something.

HarelM:
> I am going over the build script now to make sure I understand everything.

Feel free to add/update comment above the methods, I tried to keep it as readable as possible.

birkskyum:
> I can remove the changelog, I'm not sure it's helpful...

I think that's a good idea, or alternatively have it as a "CLI Changelog" to clarify

HarelM:
I've removed the changelog.

HarelM:
Let me know if there's something else before I merge.

birkskyum:
Nothing from my side. Sounded like @louwers was working on some CI

louwers:
Why not keep the changelog?

louwers (approved):
Ship it! ❤️

HarelM:
Seems like there's a need to remove the touch .nojekyll step as mkdocs does it out of the box I think...

birkskyum:
@HarelM , thank you for pulling through with this. I really couldn't wrap my head around what it would take to pull this off, because of all this inherited expression complexity, but it seems like the style-spec docs are now finally in a good place.

HarelM:
Thanks for your kind words! I was too under the impression that is a big lift, but it ended up a lot smaller than I anticipated. 
I hope that going forward it will make changing the docs easy and we'll do it often to improve the docs. 
I think the main remaining gap is expression examples, I added like 30 or so, but there's still more, it's not a huge effort though.

Issue #89, opened 2023-04-06
Generate expression metadata


birkskyum:
The expression-metadata should be generated rather than hardcoded in the docs.

https://github.com/maplibre/maplibre-style/pull/88/files/45a6e0c9dec8e4f049d922f81ac232afe8837377#diff-c77c8d013c561aabeb096e2cb19647ad9e32566daeab5a6af2cb1137bf4e49ba

HarelM:
@birkskyum can you elaborate here a bit?
The link provided above doesn't lead anywhere.
Is this still an issue? If so how can we solve/close it?

birkskyum:
This is still a bit of an issue - it's just the way the docs are generated from this code, which is quite difficult to understand, and it's almost like this is part of the spec (the expression-metadata mainly).

https://github.com/maplibre/maplibre-style-spec/blob/main/docs/src/expression/expression-types.ts
https://github.com/maplibre/maplibre-style-spec/blob/main/docs/src/expression/expressions.tsx
https://github.com/maplibre/maplibre-style-spec/blob/main/docs/src/expression/expression-metadata.tsx

HarelM:
This is basically the same issue as described in my PR:
- #626 

My suggestion is simply to add `example` for the v8 expression, this way there's no extra work on parsing etc and it is aligned with the rest of the spec.

birkskyum:
I agree that having it in the v8.json would be a great simple approach,

HarelM:
It might be that examples won't cover everything and there might be a need to move the expression syntax definition into the v8.json file.
I'm not entirely sure how to do it yet, but the typescript expression type definition also have this issue where the definition is hard coded into the generation script, see here:
https://github.com/maplibre/maplibre-style-spec/blob/c74b5030cf1749e00958f6ba1617e5019837e976/build/generate-style-spec.ts#L121-L311

My initial thought was to extend the example with something like
```
example: {
  syntax: { "special object for syntax"}, // this will define the parameters and return value
  value: '["any", ... ]' // this is an actual example of the above syntax
}
```

So that the docs generation can use this special example to generate the syntax and example from it and also the typescript generation can use that too.
I think that types page need to be updated with all the return types, otherwise things are not fully defined, IMHO.

@louwers do you have any process in the native repos that takes the v8.json file and generate something from it?

louwers:
Yes we do. We use a very old one though.

HarelM:
Can you elaborate please?

[...]

PR #639, opened 2024-05-01
Added example and syntax to all the expressions.

HarelM:
## Launch Checklist

This adds syntax and example to all the expressions.
I've used co-pilot a lot here, so some things may not be 100% accurate, but it's better than nothing I think.
Some expressions are not well defined in terms of the syntax like match, case, step, and I wasn't sure what to place there, but this can be improved over time.
I used the following file a lot to extract some of the information:
https://github.com/maplibre/maplibre-style-spec/blob/85a63c49c471c534cbd1af3c02e075feb1e94f69/docs/src/expression/expression-metadata.tsx

image

- #89

Resolves: #89 

 - [x] Link to related issues.

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -448,9 +448,9 @@ function createExpressionsContent() {
             }
             content += `\n### ${key}\n`;
             content += `${value.doc}\n`;
-            if ('example' in value) {
-                content += `\n\`\`\`json\n"some-property": ${JSON.stringify(value.example)}\n\`\`\`\n`;
-            }
+            value.example.syntax.method.unshift(`"${key}"`);
+            content += `\nSyntax:\n\`\`\`js\n[${value.example.syntax.method.join(", ")}]: ${value.example.syntax.result}\n\`\`\`\n`;
+            content += `\nExample:\n\`\`\`json\n"some-property": ${JSON.stringify(value.example.value)}\n\`\`\`\n`;
             content += sdkSupportToMarkdown(value['sdk-support'] as any);
             content += '\n';
         }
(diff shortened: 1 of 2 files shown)


louwers (changes requested):
image

Longer ones get cut off, I feel like this is something that should be addressed.

build/generate-docs.ts
```diff
@@ -448,9 +448,9 @@ function createExpressionsContent() {
             }
             content += `\n### ${key}\n`;
             content += `${value.doc}\n`;
-            if ('example' in value) {
-                content += `\n\`\`\`json\n"some-property": ${JSON.stringify(value.example)}\n\`\`\`\n`;
-            }
+            value.example.syntax.method.unshift(`"${key}"`);
+            content += `\nSyntax:\n\`\`\`js\n[${value.example.syntax.method.join(', ')}]: ${value.example.syntax.result}\n\`\`\`\n`;
```
louwers, line 452: Consider making a little helper function that wraps a string in backticks, I feel this would make this a bit more readable.

build/generate-docs.ts
```diff
@@ -448,9 +448,9 @@ function createExpressionsContent() {
...
+            content += `\nExample:\n\`\`\`json\n"some-property": ${JSON.stringify(value.example.value)}\n\`\`\`\n`;
```
louwers, line 453: Factor the example generation out in a separate function.

HarelM: I tried that, it created a json that was less helpful, IMHO. Right now you can scroll horizontally, which I found to be better.
Another option is to change the example to be a string and then decide when to insert a new line, but I find it a bit too much...

louwers: OK let me know when you are done with this, I have an idea I want to try. :)

HarelM: I've updated the code, feel free to hack.

build/generate-docs.ts
HarelM: Done.

louwers (approved):
I think this is a good compromise.

PR #641, opened 2024-05-02
Document number ranges

### LINKED ISSUE #640: Add accepted number ranges (minimum and maximum) to docs when they exist
**Describe the bug**
Some expressions have a doc specification that indicates they take a "number". Some of these accept numbers only in certain ranges. If would be great if these ranges were posted in the docs. Currently the only want to get these numbers is to experiment or dive into the repo.

One example is posted below and accepts numbers in [-1,1] https://github.com/maplibre/maplibre-style-spec/blob/main/src/reference/v8.json#L6417-L6422

**To Reproduce**
1. Go to https://maplibre.org/maplibre-style-spec/layers/#raster-saturation

**Expected behavior**
I would expect this to indicate that the range of accepted values is -1 to 1 inclusive.

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All

louwers:
Closes #640

This is what it looks like:

image

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -29,6 +29,8 @@ type JsonObject = {
     expression?: { interpolated?: boolean; parameters?: string[]};
     transition?: boolean;
     values?: {[key: string]: { doc: string; 'sdk-support'?: JsonSdkSupport }} | number[];
+    minimum?: number;
+    maximum?: number;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


build/generate-docs.ts
```diff
@@ -169,10 +177,17 @@ function convertPropertyToMarkdown(key: string, value: JsonObject, keyPrefix = '
...
     const valueType = typeToMakrdownLink(value.type);
     if (value.required) {
-        markdown += `Required${valueType}. `;
+        markdown += `Required${valueType}`;
     } else {
-        markdown += `Optional${valueType}. `;
+        markdown += `Optional${valueType}`;
+    }
+
+    if (value.minimum !== undefined || value.maximum !== undefined) {
```
HarelM, line 185: Consider moving this if into the format range method so that it will write the entire sentence.

HarelM (approved):

build/generate-docs.ts
louwers: I like pure functions better.

PR #644, opened 2024-05-05
Track parity with issues, show link to issues in docs

louwers:
This is what is looks like:

image

When a link is included instead of a version number, it indicates there is no support yet but there is a tracking issue where people can track progress, or that people can pick up.

I will remove the macOS column in a future PR, as we don't have releases for it.

~I also changed the `type` from `variableAnchorOffsetCollection` to `array` because it is a dead link right now.~

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -87,6 +87,22 @@ function codeBlockMarkdown(code: string, language = 'json'): string {
     return `\`\`\`${language}\n${code}\n\`\`\`\n`;
 }
 
+/** Renders the contents of a cell in the support matrix
+ * @param support - the support string in the style spec
+ * @returns Markdown for support cell
+ */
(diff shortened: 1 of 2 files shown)


HarelM (approved):
Great!! Don't forget the sprite sdk support. I can't think of other static pages, but if those exist then fix those as well. 
We should have an issue for every item that is not supported, right?

louwers:
@HarelM I think it makes sense to have a tracking issue for everything that is not supported (yet), but I am not 100% sure it makes sense if there is no active development at all for those missing features.

HarelM:
I think a link to a stale issue is valuable as a link to an active issue.
The user who looks at it can get the current state of thing if he is interested.
It's like a "read more" kind of link, which I believe is a better UX than the current one that only states "not supported yet".

Bottom line, it's a small increment.

PR #655, opened 2024-05-06
Document array of numbers in padding not supported with MapLibre Native

westnordost:
I noticed that the Android MapLibre API does not offer a function like `PropertyFactory::iconPadding(Float[] values)`, and using `LayoutPropertyValue("icon-padding", arrayOf(0f, 0f, -14f, -5f))` manually leads to an error being logged:

`Error setting property: icon-padding value must be a number`

I conclude that MapLibre native does not support specifying different paddings for different sides, so let's document that.

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -116,7 +116,7 @@ Enums are a closed set of possible string values. Failing to provide a value wit
 
     MapLibre Native only supports a single float as padding (see [#2363](https://github.com/maplibre/maplibre-native/issues/2368)).
 
-Values are declared similar to [CSS padding syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/padding) syntax:
+An array of floats with syntax similar to [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/padding):
 
 - A single value applies to all four sides, e.g. `[2]`;
 - two values apply to [top/bottom, left/right], e.g. `[2, 3]`;
--- a/docs/types.md
+++ b/docs/types.md
@@ -112,9 +112,10 @@ Enums are a closed set of possible string values. Failing to provide a value wit
 
 ## Padding
 
-!!! note
-
-    MapLibre Native only supports a single float as padding (see [#2363](https://github.com/maplibre/maplibre-native/issues/2368)).
+|SDK Support|MapLibre GL JS|MapLibre Native<br>Android|MapLibre Native<br>iOS|
+|-----------|--------------|-----------|-------|
+|Single number| 0.10.0 | 2.0.1 | 2.0.0 |
+|Array of numbers| 2.2.0 | ❌ ([#2363](https://github.com/maplibre/maplibre-native/issues/2368)) | ❌ ([#2363](https://github.com/maplibre/maplibre-native/issues/2368)) |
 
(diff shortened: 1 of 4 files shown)


src/reference/v8.json
```diff
@@ -1513,6 +1513,9 @@
           "android": "2.0.1",
           "ios": "2.0.0"
         },
+        "different paddings per side": {
+          "js": "2.2.0",
```
HarelM, line 1517:
```suggestion
          "js": "2.2.0"
```

HarelM:
@louwers can you confirm? 
Please approve and merge if this is the case.

louwers (changes requested):
Let me confirm, but it would make more sense to document this at the type since it would apply to all properties that take padding.

HarelM:
Sure, makes sense.

louwers (commented):
Confirmed. In MapLibre Native it can only be a single float. I'm guessing a single float (no array) also works for MapLibre GL JS? Otherwise this would be inconsistent.

It doesn't seem that any other property takes a padding (in the sense that an array can also be provided).

westnordost:
> Confirmed. In MapLibre Native it can only be a single float. I'm guessing a single float (no array) also works for MapLibre GL JS? Otherwise this would be inconsistent.

At least according to the documentation: 

> For backwards compatibility, a single bare number is accepted, and treated the same as a one-element array - padding applied to all sides.

louwers (approved):

louwers:
I created a tracking issue in the Native repo and moved the description of the padding type to `types.md`. I also added a note with a link to the tracking issue.

docs/types.md
```diff
@@ -112,10 +112,23 @@ Enums are a closed set of possible string values. Failing to provide a value wit
 
 ## Padding
 
-Paddings are similar to [CCS' padding object](https://developer.mozilla.org/en-US/docs/Web/CSS/padding) where you can specify padding in all directions using a single number, top-bottom left-right using two numbers and top, right, bottom, left using four numbers.
+!!! note
```
HarelM, line 115: Does this look better in mkdocs? Currently it doesn't look good when viewed in github.

I suggest to create the regular sdk support table beneath this type and specify the support matrix as usual, like we did in the sprite.md file.

HarelM: Support for asymmetric padding was added in maplibre 2.2, in the following PR:
https://github.com/maplibre/maplibre-gl-js/pull/1289
So a table would be the ideal solution here...

louwers: OK.

louwers: Do you want to use a macro for the support table?

Copying and duplicating it all over the place might not be a very sustainable solution.

HarelM: I don't want to complicate things, while I don't like copy pasting stuff, I find it practical for this purpose.

louwers:
I added a cross instead of just an issue number, so it is clearer it is not supported.

image

HarelM:
Shouldn't this cross be added to other support tables too? How is this support table different from the others?

louwers:
@HarelM I did add it to the other support table that also just contains a link.

I can replace "Not yet supported" with a cross in a follow up PR (but it is also pretty clear).

HarelM (approved):

HarelM:
Sorry, missed the relevant code change, my bad.

Issue #646, opened 2024-05-05
variableAnchorOffsetCollection link is dead


HarelM:
[variableAnchorOffsetCollection](https://maplibre.org/maplibre-style-spec/layers/variableAnchorOffsetCollection.md) is a dead link in the documentation.

https://maplibre.org/maplibre-style-spec/layers/?h=variableanchoroffsetcollection#text-variable-anchor-offset

This is due to recent changes in the docs, this item probably needs to be added to the types.md file or update the link in the docs generation.

PR #656, opened 2024-05-06
Ensure all types have corresponding documentation

louwers:
Adds an intergration test that checks if types have documentation.

E.g. if you remove the "Padding" section it will error with

>     Type 'padding' has anchor 'padding' which does not exist in 'types.md'

This is to avoid that we have dead links in the documentation like https://github.com/maplibre/maplibre-style-spec/issues/646

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -139,17 +139,19 @@ function requiresToMarkdown(requires: any[]): string {
     return markdown;
 }
 
+type MdLink = {text: string, link: string, anchor?: string};
+
 /**
  * Converts the type to markdown link format - the link should be to the relevant section in the right file.
  * @param type - the type of the property
  * @returns the markdown link string
  */
-function typeToMakrdownLink(type: string): string {
+export function typeToMarkdownLink(type: string): MdLink | null {
     switch (type.toLocaleLowerCase()) {
         case '*':
-            return '';
+            return null;
         case 'promoteid':
-            return ` [${type}](types.md)`;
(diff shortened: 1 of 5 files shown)


build/generate-docs.ts
```diff
@@ -190,7 +197,9 @@ function convertPropertyToMarkdown(key: string, value: JsonObject, keyPrefix = '
     if (paintLayoutText) {
         markdown += `[${paintLayoutText}](#${paintLayoutText.toLowerCase()}) property. `;
     }
-    const valueType = typeToMakrdownLink(value.type);
+
+    const mdLink = typeToMarkdownLink(value.type);
+    const valueType = mdLink ? ` ${formatMdLink(mdLink)}` : '';
```
HarelM, line 202: this doesn't take into consideration the `null` return value where it should not create an extra space.

test/integration/docs/types.test.ts
```diff
@@ -0,0 +1,65 @@
+import * as fs from 'node:fs/promises';
```
HarelM, line 1: Doesn't the mkdocs build check that and inform about dead links?
Most of the time that I fixed links was because the build informed about them...

HarelM:
Shouldn't this build fail if there's a test about dead links and we know there are dead links in the docs right now?

HarelM:
Here's the build output:
image
I don't think we need an extra test for it, maybe the following strict is the solution for this issue:
https://www.mkdocs.org/user-guide/configuration/#strict

PR #657, opened 2024-05-07
Check for dead links with mkdocs-htmlproofer-plugin

louwers:
Uses the [mkdocs-htmlproofer-plugin](https://github.com/manuzhang/mkdocs-htmlproofer-plugin) to make sure we don't have any dead links. It even checks anchor tags!

Since it's quite easy to introduce dead links by just changing a heading I think this is very useful to have. Also CI will fail on dead links now. I was able to find a dead link with it already.

I have disabled the plugin for `npm run mkdocs` (used for local development), because it takes some time to check all external links and you don't want this to happen when working on the documentation.


mkdocs.yml
```diff
@@ -60,4 +60,9 @@ plugins:
   - search
   - social:
       cards_layout_options:
-        background_color: '#295DAA'
\ No newline at end of file
+        background_color: '#295DAA'
+  - htmlproofer:
+      enabled: !ENV [ENABLED_HTMLPROOFER, True]
+      raise_error_after_finish: True
+      raise_error_excludes:
```
HarelM, line 67: Why not simply fix this?

louwers: I don't think it makes sense to create a heading in `types.md` for one-off types.

Let me see what I can do.

louwers: image

This is what it looks like now.

package.json
```diff
@@ -21,8 +21,8 @@
     "generate-style-spec": "node --no-warnings --loader ts-node/esm build/generate-style-spec.ts",
     "generate-typings": "node --no-warnings --loader ts-node/esm build/generate-typings.ts",
     "generate-docs": "node ${WATCH+--watch} --no-warnings --loader ts-node/esm build/generate-docs.ts",
-    "mkdocs": "docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material",
-    "mkdocs-build": "npm run generate-docs && docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build",
+    "mkdocs": "docker build -t squidfunk/mkdocs-material . && docker run --rm -it -p 8000:8000 -v ${PWD}:/docs -e ENABLED_HTMLPROOFER=false squidfunk/mkdocs-material",
```
HarelM, line 24: Doesn't look like the plugin is disabled in local run.

Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM squidfunk/mkdocs-material
```
HarelM, line 1: I would advise to change the file name or the file location.
Currently it looks as if the package in this repo is defined by this docker file.
If we can avoid adding this file and run a more complicated command as part of the docs build process it will be even better, but I'm not familiar with a way to do it.

louwers: @HarelM This is the way that Material for MkDocs recommends. https://squidfunk.github.io/mkdocs-material/getting-started/?h=docker#with-docker

image

package.json
louwers: `-e ENABLED_HTMLPROOFER=false`

docs/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM squidfunk/mkdocs-material
```
HarelM, line 1: This won't appear in the generated docs somehow by mistake?

package.json
HarelM: Why not simply remove the docker image build in local run?

build/generate-docs.ts
```diff
@@ -144,9 +144,10 @@ function requiresToMarkdown(requires: any[]): string {
  * @param type - the type of the property
  * @returns the markdown link string
  */
-function typeToMakrdownLink(type: string): string {
+function typeToMarkdownLink(type: string): string {
     switch (type.toLocaleLowerCase()) {
         case '*':
+        case 'variableanchoroffsetcollection':
```
HarelM, line 150: Can you check that the build fails if you remove this line? i.e. remove, commit, bring back, commit and see that the first run fails and the second passes.

docs/Dockerfile
louwers: It doesn't show up but it can be downloaded.

I can exclude it.

build/generate-docs.ts
louwers: image

HarelM: Great! thanks!

HarelM:
Last comment: https://github.com/maplibre/maplibre-style-spec/pull/657#discussion_r1592281491

package.json
louwers: Because even though the plugin is disabled, it still needs to be present because it is in the config.

HarelM (approved):

HarelM:
@louwers please note that this plugin causes the build of the docs to fail from time to time as can be seen here:
https://github.com/maplibre/maplibre-style-spec/actions/runs/9199800686/job/25305227262?pr=667

I think I prefer broken links over an unstable build, what do you think?

louwers:
@HarelM Let me see if I can configure it to only check internal links.

Issue #555, opened 2024-03-09
`(filter) expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has, within], "match" found"`


tordans:
**Describe the bug**

I am getting console error messages about filters which do work but still produce an error. I don't understand why they would be wrong or how to change them to prevent the error. I am merging filters in React (using react-map-gl) and only really have control over some of the filters but the external filters.

I can reproduce the error when adding a filter to the integration tests.

**To Reproduce online**

1. Go to https://www.osm-verkehrswende.org/cqi/map/?map=12%2F13.39%2F52.518&anzeige=incompleteness
2. Select any of the legend items to filter, eg. "Sehr unvollständig"
3. See console: `Error: layers.lane markings dashed.filter[2][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has, within], "match" found`

**To Reproduce in tests**

1. Open `test/integration/style-spec/tests/filters.input.json`
2. Append 
    ```json
        {
          "id": "mixing `match` with `==`",
          "type": "line",
          "source": "source",
          "source-layer": "source-layer",
          "filter": ["all", ["==", "Constructi", 1930], ["match", ["get", "Constructi"], [1930], true, false]]
        }
    ```
3. Run `npm run test-integration`
    ```diff
    +   },
    +   ValidationError {
    +     "line": 166,
    +     "message": "layers[16].filter[2][0]: expected one of [==, !=, >, >=, <, <=, in, !in, all, any, none, has, !has, within], \"match\" found",
        },
    ```

**Expected behavior**
I expect `match` filters to be valid filters the same as `==`

**Additional context**

I was wondering it the issue is, that `match` has a return type of `OutputType ` according to https://maplibre.org/maplibre-style-spec/expressions/#match whereas other listed "allowed" filter return `boolean` (eg `in` https://maplibre.org/maplibre-style-spec/expressions/#in) and `all` expect `boolean` https://maplibre.org/maplibre-style-spec/expressions/#all

tordans:
**Update:**

When I change my test case like this

```
    {
      "id": "mixing `match` with `==`",
      "type": "line",
      "source": "source",
      "source-layer": "source-layer",
      "filter": ["all", ["==", "Constructi", 1930], ["match", "Constructi", [1930], true, false]]
    }
```

`npm run test-integration` works fine (AKA fails with unrelated errrors)

---

[...]

PR #664, opened 2024-05-15
Docs: Improve `in` expression example

tordans:
The goal is to improve https://maplibre.org/maplibre-style-spec/expressions/#in to better explain the two way `in` can be used.

- `["in", ["get", "firstName"], "Hans Müller"]` which would be true for `firstName: "Hans"` (at least that is how I understand the spec)
- `["in", ["get", "firstName"], ["literal", ["Hans", "Dieter", "Rudolf"]]]` which would be true for `firstName: "Hans"`

See also https://github.com/maplibre/maplibre-style-spec/issues/555#issuecomment-2105132417

This change does not reflect this goal, but I think it's the best we can do which how the v8.json file is structured ATM. It does not look like it allows for multiple examples or for the examples to have variations.

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -2791,7 +2791,7 @@
         "doc": "Determines whether an item exists in an array or a substring exists in a string.\n\n - [Measure distances](https://maplibre.org/maplibre-gl-js/docs/examples/measure/)",
         "example": {
           "syntax": {
-            "method": ["value", "value"],
+            "method": [["get", "value"], ["literal", "value"]],
             "result": "boolean"
           },
           "value": ["in", "$type", "Point"]


src/reference/v8.json
```diff
@@ -2791,7 +2791,7 @@
         "doc": "Determines whether an item exists in an array or a substring exists in a string.\n\n - [Measure distances](https://maplibre.org/maplibre-gl-js/docs/examples/measure/)",
         "example": {
           "syntax": {
-            "method": ["value", "value"],
+            "method": [["get", "value"], ["literal", "value"]],
```
HarelM, line 2794: This is the syntax part of the example, so this won't be the right place to put it. 
You can use links to places in the description, for example add a page in the gl js examples with all forms of "in".
You can link to tests in this repo too.
Or your could extend the code that parses the v8 into docs to have multiple examples for expressions.

tordans:
Closed for now, follow up todos in https://github.com/maplibre/maplibre-style-spec/pull/664#discussion_r1601875482

Issue #668, opened 2024-05-18
Factor out custom stringify function


smellman:
**Describe the bug**
Same as https://github.com/maplibre/maplibre-gl-js/issues/967
In `test/lib/utils.ts` we have a stringify and the following comment:

```typescript
import compactStringify from 'json-stringify-pretty-compact';

// we have to handle this edge case here because we have test fixtures for this
// edge case, and we don't want UPDATE=1 to mess with them
export function stringify(v) {
    let s = compactStringify(v);

    if (s.indexOf('\u2028') >= 0) {
        s = s.replace(/\u2028/g, '\\u2028');
    }
    if (s.indexOf('\u2029') >= 0) {
        s = s.replace(/\u2029/g, '\\u2029');
    }
    return s;
}
```

maplibre-gl-js has removed all this function but maplibre-style-spec still available.
If this implementation is needed, close this issue.
Note: Also maplibre-style-spec uses stringifyPretty from json-stringify-pretty-compact.

PR #670, opened 2024-05-18
#668 remove custom stringify function in test

smellman:
#668 Remove custom `stringify` function in `test/lib/utils.ts`

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Thanks for taking the time to open this PR!
There is a comment above the removed function about stringify of test cases results. 
Did you try and run the tests with UPDATE=1 and see which files change? 
Maybe we can save them after the regular stringify?

smellman:
@HarelM Running UPDATE=1 npm run test generates many differences in JSON files.
I have committed these changes to the dev-668-update-1 branch.
https://github.com/smellman/maplibre-style-spec/commit/cd1574a7509e7eccc8e8b3a75113c9aed65cd8f8
Normally, UPDATE=1 is recognized as making destructive changes to make it harder to alter these files.

test/integration/expression/expression.test.ts
```diff
@@ -52,7 +51,7 @@ describe('expression', () => {
                 delete fixture.metadata;
 
                 const fname = path.join(__dirname, expressionTestFileName);
-                fs.writeFileSync(fname, `${stringify(fixture)}\n`);
+                fs.writeFileSync(fname, JSON.stringify(fixture));
```
HarelM, line 54: This is writing everything is a single line, as can be seen in the linked commit.
I suggest to change it to:
```suggestion
                fs.writeFileSync(fname, JSON.stringify(fixture, null, 4));
```

HarelM:
It's true that update=1 does destructive updates, but the difference between json compact and regular json stigngify should not be dramatic. 

It's also worth mentioning that we are using this library for the docs, so I'm not sure we will be able to remove it completely even if we remove this method. Just FYI.

test/integration/expression/expression.test.ts
smellman: I use 2 space for format.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- remove custom `stringify` function in test [#668](https://github.com/maplibre/maplibre-style-spec/issues/668))
```
HarelM, line 4: I didn't think it's interesting enough to be in the changelog...

smellman:
I just upload https://github.com/smellman/maplibre-style-spec/commit/947ebcf07ee91ca009282933b6fe0e026b1538df with 2 space.
It may be a bit dramatic.

CHANGELOG.md
smellman: OK, I will remove it.

HarelM:
I'm trying to understand if we can use a `replacer` in the json stringify to support both these tests, but more importantly, the docs, which basically suffers from the same issue: 
Expression, which are basically an array, are expanded to new lines, which makes them harder to read.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer

In any case, I went over the expression test files and they look ok to me. 

So feel free to research the above or commit the file changes as is.

smellman:
My understanding is that the `replacer` is not same as the `formatter`.
Therefore, I have uploaded all converted files.

HarelM (approved):
Thanks!

PR #676, opened 2024-05-23
Replace htmlproofer with built-in (anchor) link checking

louwers:
~Disables checking for external URLs with htmlproofer.~

~This was causing the build to fail every now and then.~

~We can still do this check manually every now and then by uncommenting this line locally.~

---

Replaces htmlproofer with built-in anchor link checking (available since MkDocs 1.6).

Uses the recommended settings from https://www.mkdocs.org/user-guide/configuration/#validation


HarelM:
Internal links are already validated as part of the build, can't we leverage that and simply the process?
Is there a timeout we can define instead of totally disabling the plugin?

louwers:
> Internal links are already validated as part of the build

Not anchor links, and the build does not fail when links are missing.

We can disable problematic external URLs. A timeout will not help (the default timeout is 10 seconds already). We are getting rate limited (429 https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429).

Let's merge this and I'll open an issue to handle 429 status codes.

HarelM:
> Not anchor links

It's all markdown, what do you mean by anchor links? 
I suggest to start with strict:
https://www.mkdocs.org/user-guide/configuration/#strict

If we see that's not enough let's complicated the build and the dependencies.
The more dependencies we have the higher the probability of stuff getting unmaintained.

Let me know if you want me to take this.

louwers:
@HarelM We use anchor links e.g. https://maplibre.org/maplibre-style-spec/layers/#layout

The built in link checker doesn't check those.

HarelM:
Are you use anchor links are not checked? I'm pretty sure I saw warnings on them when they were incorrect. But it might have been the runtime check and not the build check IDK, but it should be easy to check. 

If possible, I prefer to remove this plugin and keep it as simple as possible.

louwers:
@HarelM I can check again, but I am pretty sure that warning was for a missing page, not for a missing target of an anchor link.

louwers:
OK there is actually an `INFO` log entry without htmlproofer

```
INFO    -  Doc file 'layers.md' contains a link 'types.md#padding', but the doc 'types.md' does not contain an anchor '#padding'.
```

But strict does not fail the build. I can look into how I can make it fail the build.

louwers:
@HarelM LOL. It's a new feature since a week. https://github.com/squidfunk/mkdocs-material/discussions/2699#discussioncomment-9449965

louwers:
@HarelM Ready. :)

image

HarelM (approved):

Issue #568, opened 2024-03-14
Design Proposal: Projection


HarelM:
# Design Proposal: Projection

## Motivation

Allow users to set a projection for the map

## Proposed Change

Work in progress...

## API Modifications

Besides the part of the spec that needs to be added APIs for `setProjection` and `getProjection` should be added to the map object.

## Migration Plan and Compatibility

This is a new feature. The default would be Mercator.

## Rejected Alternatives

N/A

HarelM:
This is related to the following projection bounty:
https://github.com/maplibre/maplibre/issues/272
Which is related to globe bounty:
https://github.com/maplibre/maplibre/issues/190
As part of the globe implementation, we have found out that we need to properly define this.
See this globe initial PR:
https://github.com/maplibre/maplibre-gl-js/pull/3783
Discussion about globe:
https://github.com/maplibre/maplibre/discussions/161

cc: @louwers @sjg-wdw @wipfli @pheonor @kubapelc

wipfli:
Could there be a plugin system for projections? Because there are many projections and we probably don't want to include all of them in the library. Might be wrong though...

wipfli:
I would of course love to get support for https://en.m.wikipedia.org/wiki/Swiss_coordinate_system.

Talking of which, what is the difference between a projection and a coordinate system? And would the tiles always be assumed to use web mercator?

wipfli:
Cc @pka since you are looking into Equal Earth...

kubapelc:
In globe, anything that satisfies the Projection interface can supply its own shader code as a string. But not shader uniforms, those are currently hardcoded. I'm a bit more worried about how custom projections would interact with transform, panning the map, zooming, etc. There is a lot of complex logic in the transform class. I haven't even looked at that for globe yet.

sjg-wdw:
Having implemented all of this in another toolkit here's some random thoughts.

Web Mercator is a deceptively simple projection.  You can project back and forth with fairly simple functions and no iteration, no weird gridded offsets, that sort of thing.  As such it can be unhelpful to design around.  For example, doing the work in a shader is pretty easy, but much less so for something like UTM with a specialized ellipsoid and some strange offsets.

I'm not saying you need to implement all that madness, just design around it and drop in the pieces you already have.

A plugin system for projections is a good idea.  It should be able to:
- project between coordinate systems
- project to model globe coordinates (kind of geocentric, but maybe not exactly)
[...]

PR #687, opened 2024-05-31
Add Projection specifications

HarelM:
## Launch Checklist

This PR adds the projection specification according to the discussion here:
- #568

This is still experimental.

The following is how the docs would look like:
image

 - [x] Link to related issues.


birkskyum:
Is this to be understood as that the provided projection will be the one for low zoom levels, and then web-mercator on high zoom, and we'll add a transition at appropriate zoom levels?

HarelM:
The current implementation of globe transitions from globe to mercator without any zoom configurations or user input, it just does it. I don't know if we can actually keep the globe projection while zooming in further.
Having said that, this is still experimental and I want to be able to use it in the globe branch and see how it behaves and then decide if and how to change it.
CC: @kubapelc

birkskyum (approved):

birkskyum:
That's a good point, as long as it's experimental this will do fine.

kubapelc:
> The current implementation of globe transitions from globe to mercator without any zoom configurations or user input, it just does it. I don't know if we can actually keep the globe projection while zooming in further.

Hi, it needs to do that at a certain zoom level, because otherwise 32 bit floats (or GPU inverse trigonometry) will run out of precision and artifacts will appear. 

The transition zoom level (and the animation duration) is currently not configurable, but it is just a few numbers and I don't see any reason why users shouldn't be able to adjust it. The current values (possibly good candidates for default values) make the transition virtually unnoticeable, while the artifacts are also well hidden.

PR #779, opened 2024-08-17
Count surrogate pair as single character

### LINKED ISSUE #778: String expressions operate on UTF-16 code units instead of characters
The TypeScript reference implementations of various string expression operators operate on [UTF-16 code units](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters) instead of full Unicode characters. This is most apparent when a string contains a character outside the Basic Multilingual Plane (BMP): that character is represented by a surrogate pair of two UTF-16 code units, but the expression operators split the surrogate pair in half.

## Examples

[This example](https://jsbin.com/juvabap/5/edit?html,output) contains two symbol layers that display the length and name, respectively, of a point feature in a GeoJSON source. The latter symbol layer is filtered to only features that contain “市镇” (meaning “town”) at the zero-based index of 2.

Two labels should appear, reading “4” and “丐𦨭市镇”. However, only one label appears, reading “5”.

The number 5 all alone.

This is because JavaScript stores 𦨭 (U+26A2D) as two UTF-16 code units: `D85A DE2D`.

For a more complex example, OpenStreetMap Americana labels place names in both the user’s preferred language and the prevailing local language. To avoid clutter, it applies some heuristics to deduplicate matching names between the two languages (ignoring diacritics when comparing against English). This requires a find-and-replace operation, but since maplibre/maplibre-gl-js#2064 and maplibre/maplibre-gl-js#2059 were both declined, the application includes [some complex subexpressions](https://github.com/osm-americana/openstreetmap-americana/blob/f9a8682ea7f24e4ec604b8a2199c17f0fec0832d/src/constants/label.js#L329-L401) that depend on `length`. This works fine now but will begin to return bizarre results once maplibre/maplibre-gl-js#4550 lands, even among strings that don’t contain any surrogate pairs.

Before | After
:--:|:--:
[则拉市镇 (Chợ Lách)](https://americanamap.org/#map=12/10.25458/106.13179&language=zh)<br>[则拉市镇 (Chợ Lách)](https://www.openstreetmap.org/node/369490118) | 则拉市镇 (则hợ Lách<br>[则拉市镇 (则hợ Lách](https://www.openstreetmap.org/node/369490118)

## Impact

Until now, the impact would’ve been minimal, because GL JS has avoided rendering any codepoint beyond U+FFFF that would require surrogate pairs. The filters and properties would’ve evaluated incorrectly for any affected feature, but that would’ve been less noticeable than the abridged label for the same feature. However, maplibre/maplibre-gl-js#4550 would implement support for rendering non-BMP characters, making this issue more noticeable.

## Platform information

This issue reproduces in GL JS v4.5.2 but has probably been present ever since expressions were first implemented. The native implementation is even less intuitive, counting individual bytes: maplibre/maplibre-native#2730.

## Diagnosis

The implementations of these expression operators use traditional `String` methods that operate on UTF-16 code units. As in maplibre/maplibre-gl-js#4550, they need to be replaced with a [string iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Symbol.iterator).

https://github.com/maplibre/maplibre-style-spec/blob/76aabe5d6e586872db1a208ab32b6a9197868bf2/src/expression/definitions/length.ts#L35-L36 https://github.com/maplibre/maplibre-style-spec/blob/76aabe5d6e586872db1a208ab32b6a9197868bf2/src/expression/definitions/slice.ts#L65 https://github.com/maplibre/maplibre-style-spec/blob/76aabe5d6e586872db1a208ab32b6a9197868bf2/src/expression/definitions/index_of.ts#L69

1ec5:
The string expression operators `index-of`, `length`, and `slice` now count UTF-16 surrogate pairs as single characters instead of splitting them up into individual surrogates. Also added unit tests of these expression operators.

Fixes #778.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/expression/definitions/index_of.ts
+++ b/src/expression/definitions/index_of.ts
@@ -66,10 +66,6 @@ class IndexOf implements Expression {
         }
 
         if (isValidNativeType(haystack, ['string'])) {
-            if (needle === '') {
-                // String.prototype.indexOf considers any string to contain an empty string, unlike Array.prototype.indexOf.
-                return 0;
-            }
             const rawIndex = haystack.indexOf(needle, fromIndex);
             if (rawIndex === -1) {
                 return -1;
--- a/src/expression/definitions/index_of.ts
+++ b/src/expression/definitions/index_of.ts
@@ -60,16 +60,28 @@ class IndexOf implements Expression {
             throw new RuntimeError(`Expected first argument to be of type boolean, string, number or null, but found ${toString(typeOf(needle))} instead.`);
         }
 
-        if (!isValidNativeType(haystack, ['string', 'array'])) {
+        let fromIndex;
+        if (this.fromIndex) {
+            fromIndex = (this.fromIndex.evaluate(ctx) as number);
+        }
+
+        if (isValidNativeType(haystack, ['string'])) {
+            if (needle === '') {
+                // String.prototype.indexOf considers any string to contain an empty string, unlike Array.prototype.indexOf.
+                return 0;
+            }
+            const rawIndex = haystack.indexOf(needle, fromIndex);
+            if (rawIndex === -1) {
+                return -1;
+            } else {
+                // The index may be affected by surrogate pairs, so get the length of the preceding substring.
+                return [...haystack.slice(0, rawIndex)].length;
+            }
+        } else if (isValidNativeType(haystack, ['array'])) {
+            return haystack.indexOf(needle, fromIndex);
+        } else {
             throw new RuntimeError(`Expected second argument to be of type array or string, but found ${toString(typeOf(haystack))} instead.`);
         }
-
-        if (this.fromIndex) {
-            const fromIndex = (this.fromIndex.evaluate(ctx) as number);
-            return haystack.indexOf(needle, fromIndex);
-        }
-
-        return haystack.indexOf(needle);
     }
 
     eachChild(fn: (_: Expression) => void) {
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -2826,7 +2826,7 @@
         }
       },
       "index-of": {
-        "doc": "Returns the first position at which an item can be found in an array or a substring can be found in a string, or `-1` if the input cannot be found. Accepts an optional index from where to begin the search.",
+        "doc": "Returns the first position at which an item can be found in an array or a substring can be found in a string, or `-1` if the input cannot be found. Accepts an optional index from where to begin the search. In a string, a UTF-16 surrogate pair counts as a single position.",
         "example": {
           "syntax": {
             "method": ["value", "value", "number?"],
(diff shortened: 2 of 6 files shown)


HarelM (approved):
Kudos on all the tests you wrote!

src/reference/v8.json
```diff
@@ -2826,7 +2826,7 @@
         }
       },
       "index-of": {
-        "doc": "Returns the first position at which an item can be found in an array or a substring can be found in a string, or `-1` if the input cannot be found. Accepts an optional index from where to begin the search.",
+        "doc": "Returns the first position at which an item can be found in an array or a substring can be found in a string, or `-1` if the input cannot be found. Accepts an optional index from where to begin the search. In a string, a UTF-16 surrogate pair counts as a single position.",
```
1ec5, line 2829: I worded this a bit vaguely to leave open the possibility of adding support for [grapheme clusters](http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries) in the future. There are several potential real-world use cases for supporting grapheme clusters on maps, for example:

* The Esperanto letter `ĝ` has no precomposed character, so it must be represented as a base letter and a combining diacritic (U+0067 U+0302).
* The Hangul syllable `각` may be decomposed as U+1100 U+1161 U+11A8. The best practice is to normalize it into a single precomposed character, U+AC01, but neither vector-tile-js nor maplibre-gl-js normalizes strings, and I’m unsure if any vector tile generator does either.
* Some emoji sequences like `🇺🇳` appear as a single character but are composed of multiple underlying characters (U+1F1FA U+1F1F3).

In TypeScript, we could support grapheme clusters using the [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter) API, but it’s extremely new and I don’t know if it performs well enough for more common cases. On the native platforms, ICU has a similar API that might end up being the easiest solution for maplibre/maplibre-native#2730.

src/expression/definitions/index_of.ts
```diff
@@ -60,16 +60,28 @@ class IndexOf implements Expression {
...
-            throw new RuntimeError(`Expected second argument to be of type array or string, but found ${toString(typeOf(haystack))} instead.`);
+        let fromIndex;
+        if (this.fromIndex) {
+            fromIndex = (this.fromIndex.evaluate(ctx) as number);
         }
 
-        if (this.fromIndex) {
-            const fromIndex = (this.fromIndex.evaluate(ctx) as number);
+        if (isValidNativeType(haystack, ['string'])) {
+            if (needle === '') {
```
HarelM, line 69: Can you clarify this? won't we get the same results if we remove this if (and the return 0)?

1ec5: Oh you’re right, I changed it to stop using `Array.prototype.indexOf` at the last minute, so this is no longer necessary. 👍

HarelM:
CC: @louwers - I don't think this is a dramatic change, more in the realm of a bug fix, but we need to make sure this is OK with native.
I'll make sure this ends up as a render test(s) so that we will be able to make sure parity is achieved.

PR #791, opened 2024-08-23
Don't type style spec as `any`

louwers:
This makes it a bit easier to patch it for MapLibre Native.

@HarelM Tdoes not seem to be the same file as `v8.json` for some reason...

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -70,7 +70,7 @@ export type StylePropertySpecification = {
 };
 
 import v8Spec from './reference/v8.json' with {type: 'json'};
-const v8 = v8Spec as any;
+const v8 = v8Spec;
 import latest from './reference/latest';
 import derefLayers from './deref';
 import diff from './diff';


HarelM:
Not sure I understand the last comment...?

HarelM (approved):
I think there was an issue with dts bundle generator that this was mandatory. But if CI doesn't fall I'm good.

louwers:
@HarelM When I import `v8` or `latest` I get a different object than `v8.json`...

HarelM:
When importing v8 you get an optimized for size object for runtime check, this is why `docs` and `examples` are not part of it.
Maputnik is using the following import:
https://github.com/maplibre/maputnik/blob/8184ac8393db9ed3867ab7888206a2451822b168/src/components/FieldMaxZoom.tsx#L3

`import latest from '@maplibre/maplibre-gl-style-spec/dist/latest.json'`

Hope that helps.

louwers:
@HarelM Thanks!!

PR #797, opened 2024-08-29
Validate `sdk-support` in spec, add missing tracking issues

louwers:
- This PR adds a test that ensures there is a tracking issue for each feature that is at least supported on one platform. This test will fail with a helpful message if this is not the case.
- Using this test, I was able to find all missing tracking issues (and created a few more).
- A special value `"wontfix"` can be used if a feature will never be supported. I used this for MapLibre GL JS and [`volatile`](https://maplibre.org/maplibre-style-spec/sources/#volatile). I think we should avoid these going forward, the spec should be cross-platform.

### DIFF
--- a/test/integration/style-spec/validate_spec.test.ts
+++ b/test/integration/style-spec/validate_spec.test.ts
@@ -19,80 +19,78 @@ describe('validate_spec', () => {
     });
 
     test('errors from validate do not contain line numbers', () => {
-        const style = JSON.parse(
-            fs.readFileSync(
-                'test/integration/style-spec/tests/bad-color.input.json',
-                'utf8'
-            )
-        );
+        const style = JSON.parse(fs.readFileSync('test/integration/style-spec/tests/bad-color.input.json', 'utf8'));
 
         const result = validate(style, reference);
         expect(result[0].line).toBeUndefined();
     });
+});
 
-    test('Validate sdk-support in spec', () => {
-        const issueTrackers = {
-            js: 'https://github.com/maplibre/maplibre-gl-js/issues',
-            android: 'https://github.com/maplibre/maplibre-native/issues',
-            ios: 'https://github.com/maplibre/maplibre-native/issues',
-        };
-        const platforms = Object.keys(issueTrackers);
+describe('Validate sdk-support in spec', () => {
+    const issueTrackers = {
+        js: 'https://github.com/maplibre/maplibre-gl-js/issues',
+        android: 'https://github.com/maplibre/maplibre-native/issues',
+        ios: 'https://github.com/maplibre/maplibre-native/issues',
+    };
+    const platforms = Object.keys(issueTrackers);
 
-        function validatePlatforms(platformSupport, path) {
-            const notSupportedOnAnyPlatform = platforms.every(
-                (platform) => !platformSupport[platform]
-            );
+    function validatePlatforms(platformSupport, path) {
+        const notSupportedOnAnyPlatform = platforms.every(
+            (platform) => !platformSupport[platform]
+        );
 
-            for (const platform of platforms) {
+        for (const platform of platforms) {
+            test(`Validate sdk-support ${path.join('.')}`, () => {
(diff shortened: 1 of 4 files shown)


test/integration/style-spec/validate_spec.test.ts
```diff
@@ -19,10 +19,80 @@ describe('validate_spec', () => {
...
+                        `Unsuppored sdk-support value '${support}'. Use one of the following\n` +
+              '- If supported: version number (e.g. 1.0.0) to indicate support since this version (or "supported" to indicate it has always been supported)\n' +
+              '- If it will never be supported: "wontfix" to indicate it will never be supported\n' +
+              '- A link to a tracking issue'
+                    );
+                }
+            }
+        }
+
+        function validateSdkSupport(sdkSupportObj, path) {
```
HarelM, line 78: I would consider removing `validateSdkSupport` and `checkRoot` methods as I don't think they are adding to the readability of the test.

test/integration/style-spec/validate_spec.test.ts
```diff
@@ -19,10 +19,80 @@ describe('validate_spec', () => {
...
+                'test/integration/style-spec/tests/bad-color.input.json',
+                'utf8'
+            )
+        );
 
         const result = validate(style, reference);
         expect(result[0].line).toBeUndefined();
     });
 
+    test('Validate sdk-support in spec', () => {
```
HarelM, line 33: You can create the tests dynamically like done above so you can have the test name indicate which definition in the spec failed. This is just a though.

test/integration/style-spec/validate_spec.test.ts
```diff
@@ -19,10 +19,80 @@ describe('validate_spec', () => {
...
+                const version = /^\d+\.\d+\.\d+$/;
+                const values = new Set(['supported', 'wontfix']);
+
+                const support = platformSupport[platform];
+                if (
+                    !support.match(maplibreIssue) &&
+          !support.match(version) &&
+          !values.has(support)
+                ) {
+                    throw new Error(
```
HarelM, line 68: This is not how tests should work - you should have `expect(somethign).toBe(something elese).
In this case I would recommend something like:
```ts
expect(support.match(maplibreIssue).toBeFalsy();
expect(support.match(version)).toBeFalsy();
expect(values.has(support)).toBeFalsy();
```

HarelM:
Besides some nit-picking this looks great! Thanks!

test/integration/style-spec/validate_spec.test.ts
```diff
@@ -19,10 +19,80 @@ describe('validate_spec', () => {
     });
 
     test('errors from validate do not contain line numbers', () => {
-        const style = JSON.parse(fs.readFileSync('test/integration/style-spec/tests/bad-color.input.json', 'utf8'));
+        const style = JSON.parse(
+            fs.readFileSync(
```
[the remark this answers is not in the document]
louwers: Oh yeah a formatter did this.

test/integration/style-spec/validate_spec.test.ts
louwers: That is not equivalant though, and there should be a helpful message when it fails.

HarelM: You can have the test name to be meaningful, so that when it fails it helps understanding what went wrong. 
What I'm concerned about is that lint and other tools may report that the test doesn't have an expect, which isn't great in terms of future proofing this.

test/integration/style-spec/validate_spec.test.ts
louwers: Done

test/integration/style-spec/validate_spec.test.ts
louwers: Since this is a recursive function, it is not easy to rewrite.

I have added some comments for clarification.

HarelM (approved):

HarelM:
Thanks!

Issue #851, opened 2024-10-10
Add Elevation


NathanMOlson:
# Design Proposal: Add Center Elevation

## Motivation

I would like to be able to pitch the camera above 90 degrees.
https://github.com/maplibre/maplibre-gl-js/issues/4717
https://github.com/NathanMOlson/maplibre-gl-js/pull/1

To keep the camera above the terrain, the center point elevation must be above terrain when pitch is > 90. 2D center point is no longer sufficient to place the camera.

## Proposed Change

I would like to add a field `elevation` to the root style spec object. `elevation` sets the center point elevation in meters above sea level, and joins its friend `center`.

## API Modifications

Root style spec gets a new `elevation` field. `jumpTo()`, `easeTo()`, and `flyTo()` get `elevation` added to their options.

## Migration Plan and Compatibility

This is a pure extension. To achieve existing functionality, just leave `elevation` undefined (default 0).

## Rejected Alternatives

I considered adding `elevation` to the API but not to the style spec. Rejected because:

1. `elevation` is needed in style spec for render tests.
2. Similar field `center` already exists in style spec

NathanMOlson:
Understood about the render tests.

The elevation proposed here corresponds to `map.transform.elevation`, which is the elevation of the center point (`map.transform.center`).

For pitch angles >= 90 degrees, if the center point is at the same elevation as the map, the camera will be below the map. To allow the camera to be above the map, it must look at a center point which hovers above the map (`map.transform.elevation` must be greater than the elevation of the map, which is zero in the absence of terrain).

HarelM:
Is it possible to add a third number to center?

wipfli:
Is `center` the location of the camera or is it the location of where the camera looks at?

wipfli:
I know `center` and `zoom` are already part of the style spec so we will not remove them, but in my opinion they should never have been added. I think the style sheet should encode how the map looks like, and the MapLibe GL JS / Native API should take care of what you do with the map, i.e., where to look at, zoom level, field of view etc...

HarelM:
Center is where the camera is looking at.
I respectfully disagree about the fact that center shouldn't be part of the spec.
Consider the case where you want to present a map of a location (for example a web page with a business showing where it resides).
Setting zoom, center, pitch, bearing allows a designer to indicate exactly how the map should look without needing to change a single line of code.

wipfli:
OK, so
- `center` is the location of the point where the camera is looking at
[...]

PR #875, opened 2024-10-29
Add centerAltitude to style spec

NathanMOlson:
## Launch Checklist

Add `centerAltitude` to style spec, as discussed here: https://github.com/maplibre/maplibre-style-spec/issues/851

 - [X] Link to related issues.

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -36,7 +36,7 @@
     },
     "centerAltitude": {
       "type": "number",
-      "doc": "Default map center altitude in meters above sea level.  The style center altitude will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
+      "doc": "Default map center altitude in meters above sea level. The style center altitude defines the altitude where the camera is looking at and will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
       "example": 123.4,
       "sdk-support": {
         "basic functionality": {
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -34,6 +34,11 @@
         40.7736
       ]
     },
+    "centerAltitude": {
+      "type": "number",
+      "doc": "Default map center altitude in meters above sea level.  The style center altitude will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
+      "example": 123.4
+    },
     "zoom": {
       "type": "number",
       "doc": "Default zoom level.  The style zoom will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -37,7 +37,14 @@
     "centerAltitude": {
       "type": "number",
       "doc": "Default map center altitude in meters above sea level.  The style center altitude will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
-      "example": 123.4
+      "example": 123.4,
+      "sdk-support": {
+        "basic functionality": {
+          "js": "5.0.0",
+          "android": "https://github.com/maplibre/maplibre-native/issues/2980",
+          "ios": "https://github.com/maplibre/maplibre-native/issues/2980"
+        }
+      }
     },
     "zoom": {
       "type": "number",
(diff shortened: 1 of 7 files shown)


src/reference/v8.json
```diff
@@ -34,6 +34,11 @@
         40.7736
       ]
     },
+    "centerAltitude": {
```
HarelM, line 37: Are we expecting other altitudes? Why not simply "altitude"?

NathanMOlson: I am not expecting other altitudes in the stylesheet, but centerAltitude makes it clear that 1) it is related to the existing `center` field, and 2) it is not the camera altitude.

HarelM:
LGTM.
@ibesora @louwers please approve as well.

src/reference/v8.json
louwers: Sorry if this is not a camera altitude, then what is the altitude of?

HarelM: The center of the map, what/where the camera is looking at.

louwers: OK I get it now... Sorry I am a bit slow.

I honestly don't understand the use case of this property (or `bearing`, `pitch`, `roll` or even `center` for that matter). I'm with @wipfli that this should be done with an API.

> I know `center` and `zoom` are already part of the style spec so we will not remove them, but in my opinion they should never have been added. I think the style sheet should encode how the map looks like, and the MapLibe GL JS / Native API should take care of what you do with the map, i.e., where to look at, zoom level, field of view etc...

I'm can't think of a reason why you'd want to put these in a style.

Fun fact: `center` is broken on MapLibre Native because as part of map initialization the map is shifted a bit so it detects 'map interaction' and doesn't set the center... Nobody has complained about it in 5 years.

HarelM: While this might be hard to do, try to put yourself in the shoes of a map designer, when you don't know how to write code at all, and even a json file is not a straight forward concept.
You work on maputnik for example, style the map set where the center, zoom, pitch etc is and export the style to be consumed by a different team, or even export the html page that maputnik exports. 
When these are defined in the style all you need to do is edit a json, no code, less complexity, and as a designer, you have fun control of how it would look. 
As opposed to what you might be familiar with, in big corporations, these split between those who wrote code and those that doesn't is very common, and the time of those that write code is usually very expensive, so as an organization, you try to facilitate as much as possible without needing someone to code...

louwers: Using that same argument we can add animations and user interactions, UI controls and a whole suite of other features that have little to do with how a map looks to the spec.

> You work on maputnik for example

Right now you have to input numbers in the Center/Zoom/Bearing/Pitch in Maputnik. There's no way to preview the values you entered or copy the current settings from the map. So I think (at least now) the workflow you describe is hypothetical.

louwers: Doesn't this property conflict with `zoom` and probably `pitch` as well? Won't this over-specifying the initial camera position?

Anyway, I'm pretty sure this property is not really relevant for MapLibre Native. I'll just trust @HarelM's and @ibesora's judgement on this.

ibesora: > Doesn't this property conflict with zoom and probably pitch as well? Won't this over-specifying the initial camera position?

It’s only used if no other means of setting the camera have been used and it enables the “orbiting around a point not in the terrain” use case.

I’m with @HarelM about this making sense in the style. To me, where the camera is looking is an integral part of how a map looks like, similar to center.

ibesora (approved):

src/reference/v8.json
```diff
@@ -34,6 +34,18 @@
...
+      "type": "number",
+      "doc": "Default map center altitude in meters above sea level.  The style center altitude will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
```
HarelM, line 39:
```suggestion
      "doc": "Default map center altitude in meters above sea level. The style center altitude defines the altitude where the camera is looking at and will be used only if the map has not been positioned by other means (e.g. map options or user interaction).",
```

HarelM (approved):

PR #878, opened 2024-11-02
Rename adaptive `globe` to `globe-to-mercator`

birkskyum:
Rename of the adaptive globe from `globe` to `globe-to-mercator` as was suggested here:

> [Birk](https://github.com/maplibre/maplibre-gl-js/pull/3963#issuecomment-2366733199): it would be great to have the projection API extend well to: https://github.com/maplibre/maplibre-gl-js/issues/168

> _[Harel](https://github.com/maplibre/maplibre-gl-js/pull/3963#issuecomment-2367467741): Maybe we can name it globe-mercator to indicate both are used, IDK._

> _[Jakub](https://github.com/maplibre/maplibre-gl-js/pull/3963#issuecomment-2368243309): I think globe-mercator is a good name._

There are many good [reasons](https://github.com/maplibre/maplibre-style-spec/pull/878#issuecomment-2454718370) for leaving `globe` open to a globe-only projection (no mercator transition), and to be more explicit about what `globe-to-mercator` does. Doing an early rename of the adaptive projection will minimize the impact on plugin authors that are now starting to use our adaptive globe projection:
- https://github.com/maptiler/maptiler-sdk-js/pull/132)

## Launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
-- Rename adaptive globe from `globe` to `globe-mercator` ([#878](https://github.com/maplibre/maplibre-style-spec/pull/878))
+- Rename adaptive globe from `globe` to `globe-to-mercator` ([#878](https://github.com/maplibre/maplibre-style-spec/pull/878))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- Rename adaptive globe from `globe` to `globe-mercator` ([#878](https://github.com/maplibre/maplibre-style-spec/pull/878))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/src/diff.test.ts
+++ b/src/diff.test.ts
@@ -607,10 +607,10 @@ describe('diff', () => {
         } as StyleSpecification,
         {
             projection: {
-                type: 'globe'
+                type: 'globe-mercator'
             }
         } as StyleSpecification)).toEqual([
-            {command: 'setProjection', args: [{type: 'globe'}]},
+            {command: 'setProjection', args: [{type: 'globe-mercator'}]},
         ]);
     });
 });
--- a/src/diff.test.ts
+++ b/src/diff.test.ts
@@ -607,10 +607,10 @@ describe('diff', () => {
         } as StyleSpecification,
         {
             projection: {
-                type: 'globe-mercator'
+                type: 'globe-to-mercator'
             }
         } as StyleSpecification)).toEqual([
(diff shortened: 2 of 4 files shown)


HarelM:
When looking at this PR I can't help but think about making the value an interpulatable with expression. This way you have full control over the transition and can develop other layers (custom mainly) knowing exactly when things change instead of making assumptions. 
This can lead in the future to allow transition from any protection to any protection based on zoom level (or maybe other stuff, IDK).
Let me know what you think.
Cc: @pessimistress, @ibesora @louwers

birkskyum:
It's an interesting idea. I think we're better off hardcoding the useful options for the time being, and then later on re-asses if we can generalize something.

HarelM:
Let's continue the discussion in the relevant issue.

birkskyum:
That's a good idea. I consider this pr, merely adding a setting to use the already-implemented globe projection without an adaptive transition, as a quite separate concern than the discussion of expanding the adaptive transitions with advanced expressions.

birkskyum:
@HarelM , I've reduced this to just the rename of the adaptive projection to `globe-mercator` like you suggested.

louwers:
I think `globe-mercator` if pretty confusing. 🫠 

Also, now that I think of it, is a globe even a projection? I mean, it is in the mathematical sense, but a 'map projection' is commonly a projection of a globe on a plane, not the other way around. https://en.wikipedia.org/wiki/Map_projection

We should maybe bring some GIS experts into the discussion?

birkskyum:
@louwers the globe is a Spherical projection

louwers:
Maybe `globe-to-mercator` would be a bit more descriptive?

Wr.t. to @HarelM's comment about expressions. At a later stage we can take an object for this property that you can use to configure where transitions occur. We would keep the string version for backward compatibility and so that people can easily pick a sensible default without having to supply a complicated object-based configuration if they don't need it.

pka:
`globe-to-mercator` sounds good to me. I've made a prototype switching from Equal Earth projection to Web Mercator, so it could even make sense to declare more than one projection. But transitioning from a globe to a flat map is special enough to justify a dedicatet combined projection name.

louwers (approved):

birkskyum:
@pka Equal Earth support sounds great! Looking forward to see it.

birkskyum:
Breaking [experimental / under-development properties](https://maplibre.org/maplibre-style-spec/projection/), that's a SemVer minor, right?

HarelM:
If someone is counting on this version to change when there are breaking changes then yes.
Since I don't think a lot of people care about the version of the style-spec I would advise to change the major version of it to keep semver standard.
I must say I still don't think this is a good approach to the style spec - mainly because if we think of the projection as an equivalent to a color name, then you don't have `blue-to-black` colors when you want to change the color based on zoom level, you have blue, black and an expression if you want to do "transitioning" colors...
Anyway... I guess I'm in the minority here...

louwers:
@HarelM On the other hand, if 99% of the people where happy with `blue-to-black` and didn't want to customize it further, it would make sense to have a shorthand for it.

PR #888, opened 2024-11-04
Add projection expression syntax

birkskyum:
Upcoming syntax:

See Video examples [here](https://github.com/maplibre/maplibre-style-spec/pull/888#issuecomment-2472017486)

projection: { type: VALUE }

VALUE can be either of below
- string (projection) - "mercator" / "vertical-perspective"
- [string, string, number] - [vertical-perspective, mercator, 0.2]
- `step` Expression

```
["step", ["zoom"],
"vertical-perspective", 
11, "mercator"] -> string
```

- `interpolate-projection` Expression

```
["interpolate-projection", ["linear"], ["zoom"], 
10, "vertical-perspective", 
12, "mercator"] -> string | [string, string, number]
```

- string (preset expression) - "globe" (we'll replace it in the client with either the shader-globe, or with an `interpolate-projection` expression like above.

--- 

Just one of many motivations for this feature is the need for a non-adaptive `globe` for the reasons [here](https://github.com/maplibre/maplibre-style-spec/pull/878#issuecomment-2454718370)

Generally, we need a way to specify adaptive and non-adaptive transitions in the style spec, with stops, and we need much more flexibility regarding projections, where this syntax is future proof for +proj and WKT strings support.


birkskyum:
@HarelM , I understand what you mean.

Taking a more conservative approach, I've changed this PR to remove the globe-to-mercator, meaning that there's just the `globe` and the `mercator`. That's the future proof way to do the spec, but it does away with the adaptive part of `globe`, meaning that users of GL JS should expect a non-adaptive spherical projection when using `globe`, until we figure out how to make an expression syntax. This would go a long way towards making that possible, because users then won't need mercator for the terrain, when using a non-adaptive `globe`:
- https://github.com/maplibre/maplibre-gl-js/pull/4977

birkskyum:
I've designed an expression based on the step. We'd have to make the interpolation in the client. @HarelM , is it something like this you have in mind?

jest.config.ts
```diff
@@ -18,7 +18,7 @@ const config: Config = {
             displayName: 'unit',
             testEnvironment: 'jsdom',
             testMatch: [
-                '<rootDir>/src/**/*.test.{ts,js}'
+                '<rootDir>/src/**/validate_projection.test.{ts,js}'
```
HarelM, line 21: Is this change intended?

build/generate-style-spec.ts
```diff
@@ -17,31 +17,51 @@ function propertyType(property) {
...
-                return property.type;
-            case 'enum':
-                return unionType(property.values);
-            case 'array': {
-                const elementType = propertyType(typeof property.value === 'string' ? {type: property.value, values: property.values} : property.value);
-                if (property.length) {
-                    return `[${Array(property.length).fill(elementType).join(', ')}]`;
-                } else {
-                    return `Array<${elementType}>`;
+        if (typeof property.type === 'string') {
```
HarelM, line 20: Isn't this always string?
Can you elaborate on what the generation output changes are here?

HarelM:
I think this is in the right direction, thanks!
How would the interpolation go then?
I think for example, color interpolation code is written in this repo, should there be an interpolation code for projection type here as well?
If this is not needed here, that's great too.

build/generate-style-spec.ts
birkskyum: i was just confused by the `type`, because of projection.`type` = `mercator`, and `type`: `projection` and `type`: `emum`

birkskyum:
So since it's a step, there's no interpolation in the expression.

We need adapters for the combination, so say it's:
10, 'globe',
12. 'mercator,

Then we have to figure, okay it's the `globe-to-mercator` transition, and then interpolate the globeness there.

If we don't have an adapter, it would make a hard cut from one to the other.

build/generate-style-spec.ts
```diff
@@ -38,6 +38,8 @@ function propertyType(property) {
                 return 'SkySpecification';
             case 'sources':
                 return '{[_: string]: SourceSpecification}';
+            case 'projection:':
+                return 'ProjectionSpecification';
```
HarelM, line 42: I think you need to define this somewhere so that it can be added to the typescript definitions, we did a similar thing for padding, I think you can try and copy from it.

HarelM:
Besides the comment above this looks good.
I've added a comment in the other PR about using `npm link`. Let me know in which direction you would like to proceed!
Thanks for all the work around this! Truly appreciated!

build/generate-style-spec.ts
birkskyum: I ran generate-typings, and I think it might be working already.

export type ProjectionSpecification = {
	"type"?: PropertyValueSpecification<"mercator" | "globe">;
};

Screenshot 2024-11-05 at 08 57 18

birkskyum:
@HarelM , I have some troubles running the mkdocs. I don't know why there's a recursion error in some social plugin.

    return self._resolve_font(family, fallback)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/material/plugins/social/plugin.py", line 504, in _resolve_font
    return self._resolve_font(family, fallback)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/material/plugins/social/plugin.py", line 504, in _resolve_font
    return self._resolve_font(family, fallback)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 977 more times]
  File "/usr/local/lib/python3.11/site-packages/material/plugins/social/plugin.py", line 472, in _resolve_font
    path = os.path.join(self.config.cache_dir, "fonts", family)
                        ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/mkdocs/config/base.py", line 97, in __get__
    if not isinstance(obj, Config):
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen abc>", line 119, in __instancecheck__
RecursionError: maximum recursion depth exceeded in comparison

birkskyum:
Ah, I had an old `.cache` folder in root

HarelM:
I think I had it too, don't remember how I solved it, you can comment out the social part in the mkdocs config to overcome this locally, just make sure you don't commit this change.

HarelM:
You should be able to call `evaluate` and give it the relevant zoom, feature state and tile id, I think. 
given that we have interpolate-hcl and lab, interoperate-proj (or maybe use the full word, IDK) sounds like a good idea. 

I'm not sure about the return value though. 
I would expect the return value to be a Projection type maybe, like a new kind of object that we add to the style spec, like color or padding, or resolved Image.
https://maplibre.org/maplibre-style-spec/types/

In theory (not relevant to current implementation) we might consider an object with all the field that are described in proj4js:
https://github.com/proj4js/proj4js/blob/master/lib/Proj.js

Right now it can be a "named" protection to simplify things. 
We should consider aligning the names to proj specifications in this early stage. Just a thought.

birkskyum:
@HarelM , the return type is a ProjectionTransition (`interpolate-projection`, renamed it) or Projection. (`step`,  `string`)

src/expression/types.ts
```diff
@@ -13,6 +13,9 @@ export type BooleanTypeT = {
 export type ColorTypeT = {
     kind: 'color';
 };
+export type ProjectionTypeT = {
+    kind: 'string';
```
HarelM, line 17: I think this should be something else, shouldn't it?

src/expression/types.ts
```diff
@@ -56,6 +59,7 @@ export const NumberType = {kind: 'number'} as NumberTypeT;
 export const StringType = {kind: 'string'} as StringTypeT;
 export const BooleanType = {kind: 'boolean'} as BooleanTypeT;
 export const ColorType = {kind: 'color'} as ColorTypeT;
+export const ProjectionType = {kind: 'string'} as ProjectionTypeT;
```
HarelM, line 62: Same here

src/util/interpolate.test.ts
```diff
@@ -117,6 +118,16 @@ describe('interpolate', () => {
...
     });
 
+    test('interpolate projection', () => {
+        
+        const Globe = new Projection('globe');
+        const Mercator = new Projection('mercator');
+
+        const i11nFn = (t: number) => interpolate.projection(Globe, Mercator, t);
+        expect(i11nFn(0.5)).toBeInstanceOf(ProjectionTransition);
+        expect(i11nFn(0.5).toString()).toBe('["globe", "mercator", 0.5]');
```
HarelM, line 128: This what got me confused, I would advise to compare using "toEqual" against an object without the toString conversation.

src/util/interpolate.ts
```diff
@@ -37,6 +38,10 @@ function number(from: number, to: number, t: number): number {
     return from + t * (to - from);
 }
 
+function projection(from: Projection, to: Projection, interpolation: number): ProjectionTransition {
```
HarelM, line 41: If we compare this method to other interpolation methods we see that the input from and to are the same type as the return value. 
I think we should try and keep that concept in mind and see what can be done to define the projection object better to support this notion.

birkskyum: The projection is unique in this regards, because an `orange` return color can hold the information of being between `yellow` and `red`, but I don't have a way to encapsulate an intermediate between `mercator` and `globe` in the same fashion, that's why i return the minimal information `[globe, mercator, 0.8]`, which will allow gl js to calculate what this intermediate projection look like.

HarelM: Maybe we can define "pure" protection as one that have both from and to the same with 1 for "weight".
and have the interpolate value calc a new Projection with new from to and weight.

src/util/interpolate.test.ts
HarelM: You can check the babies of from, to etc separately.

src/util/interpolate.ts
birkskyum: I was a bit concerned it would be verbose, but let's try to see if it's at least possible to implement with this approach. I've updated the code to follow this pattern.

src/util/interpolate.ts
```diff
@@ -37,6 +38,10 @@ function number(from: number, to: number, t: number): number {
     return from + t * (to - from);
 }
 
+function projection(from: Projection, to: Projection, interpolation: number): Projection {
+    return new Projection(from, to, interpolation);
```
HarelM, line 42:
```suggestion
    return new Projection(from.from, to.to, interpolation);
```
That way Projection can always hold a string in from and to?

birkskyum: Doesn't this throw away the information of
- from.to
- from.transition
- to.from
- to.transiton

HarelM: It does, do we need it?

birkskyum: only for very recursion - going between two transitions.

src/util/interpolate.ts
birkskyum:
```ts
let errors = validateProjection({validateSpec, value: {type: ['step', ['zoom'], ['literal',{'from': 'globe', 'to': 'globe', transition: 1}], 10, ['literal',{'from': 'mercator', 'to': 'mercator', transition: 1}]]}, styleSpec: v8, style: {} as any});

errors = validateProjection({validateSpec, value: {type: ['literal', {'from': 'mercator', 'to': 'mercator', transition: 1}]}, styleSpec: v8, style: {} as any});

errors = validateProjection({validateSpec, value: {type: ['interpolate-projection', ['linear'],['zoom'], 0, {'from': 'globe', 'to': 'globe', transition: 1}, 10, {'from': 'mercator', 'to': 'mercator', transition: 1}]}, styleSpec: v8, style: {} as any});

```

it's verbose. i'd really like if we can also allow `globe` instead of `['literal',{'from': 'globe', 'to': 'globe', transition: 1}]`

birkskyum:
I think i got most of it worked out now, with the types. there's something with the  `interpolate-projection`. typescript doeasn't complaint, but the validate test does.

src/util/interpolate.ts
birkskyum: We can probably disregard that usecase

HarelM: Much like color, we should be able to parse "globe" to be `["globe", "globe", 1]`, like red is converted to rgba under the hood.

HarelM:
Can you check if the coverage can be improved in terms of tests?

build/generate-style-spec.ts
```diff
@@ -122,6 +124,37 @@ fs.writeFileSync('src/types.g.ts',
...
 export type ColorSpecification = string;
 
+export type PrimitiveProjection = 'mercator' | 'globe';
+export type ProjectionTransition = [PrimitiveProjection, PrimitiveProjection, number];
+export type ProjectionType = PrimitiveProjection | ProjectionTransition | PropertyValueSpecification<ProjectionTransition>
+export type ProjectionSpecification = {"mode":ProjectionType}
+
+
+
+export function isProjectionType(value: unknown): value is ProjectionType {
```
HarelM, line 134: This file generates code, I don't think it should have methods here, these methods can be written in regular typescript files that can hold this logic.

docs/types.md
```diff
@@ -133,3 +133,43 @@ The following example applies 2em padding on top and bottom and 3em padding left
...
+There are currently two projection modes implemented.
+
+- `mercator` - A web mercator projection
+- `globe` - A spherical projection
+
+
+**Use a single projection**
+
+```ts
+mode: "mercator"
```
HarelM, line 148: "mode" is not part of the style spec, right? I think this is a bit confusing, isn't it?

src/expression/definitions/interpolate.ts
```diff
@@ -19,18 +19,18 @@ export type InterpolationType = {
...
     controlPoints: [number, number, number, number];
 };
-type InterpolatedValueType = NumberTypeT | ColorTypeT | PaddingTypeT | VariableAnchorOffsetCollectionTypeT | ArrayType<NumberTypeT>;
+type InterpolatedValueType = NumberTypeT | ColorTypeT | ProjectionTypeT | PaddingTypeT | VariableAnchorOffsetCollectionTypeT | ArrayType<NumberTypeT>;
 
 class Interpolate implements Expression {
     type: InterpolatedValueType;
 
-    operator: 'interpolate' | 'interpolate-hcl' | 'interpolate-lab';
+    operator: 'interpolate' | 'interpolate-hcl' | 'interpolate-lab' | 'interpolate-projection';
```
HarelM, line 27: Looks like `'interpolate' | 'interpolate-hcl' | 'interpolate-lab' | 'interpolate-projection'` deserves a type of its own.

src/expression/values.ts
```diff
@@ -37,6 +38,7 @@ export function isValue(mixed: unknown): boolean {
         typeof mixed === 'string' ||
         typeof mixed === 'boolean' ||
         typeof mixed === 'number' ||
+        isProjectionTransition(mixed) ||
```
HarelM, line 41: Can't we use `mixed instanceof Projection`?

src/util/interpolate.ts
HarelM: Check out the parse method of color here:
https://github.com/maplibre/maplibre-style-spec/blob/5788145317e1b7cf444a324207822706eec9071a/src/util/color.ts#L61

Can we do something similar for projection?
I'm expecting in the future to get wkt projection definition here in theory, so it would be good to have a parser that can support the simple types for now, not need to support wkt right now, but I think we should take that into consideration  in terms of design.

build/generate-docs.ts
```diff
@@ -166,6 +166,7 @@ function typeToMarkdownLink(type: string): string {
         case 'promoteid':
             return ` [${type}](types.md)`;
         case 'color':
+        case 'projection-mode':
```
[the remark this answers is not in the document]
birkskyum: "projection" is already used for the top-level category, and we need a value type for the projection expression, and it has to have a different name than the top level to avoid recursion issues.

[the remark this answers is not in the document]
birkskyum: Top level
https://github.com/maplibre/maplibre-style-spec/pull/888/files#diff-d9c8777c5eac8bd8dadac08a9fdc12df4109935355f2c98613b191436125a09bR129

docs/types.md
birkskyum: I renamed `type` to `mode`, because `type` is notoriously overloaded in the style spec.

HarelM: My only concern about using `style.projection` is what will happen in the future if we decide to add things to projection settings, `mode` and `type` are synonymous, so I don't have strong feelings either way, but if you decide to change the style spec definition I think it should be noted somewhere besides this PR, right?

src/expression/definitions/interpolate.ts
```diff
@@ -171,9 +174,15 @@ class Interpolate implements Expression {
...
         const outputUpper = outputs[index + 1].evaluate(ctx);
 
+        if (this.type.kind === 'projection') {
+            return interpolate.projection(outputLower, outputUpper, t);
+        }
+
         switch (this.operator) {
             case 'interpolate':
-                return interpolate[this.type.kind](outputLower, outputUpper, t);
+                return interpolate[ this.type.kind ](outputLower, outputUpper, t);
```
HarelM, line 183:
```suggestion
                return interpolate[this.type.kind](outputLower, outputUpper, t);
```

src/util/projection.ts
```diff
@@ -0,0 +1,51 @@
...
+    readonly to: PrimitiveProjection;
+    readonly transition: number;
+    constructor(from: PrimitiveProjection, to: PrimitiveProjection, transition: number){
+        this.from = from;
+        this.to = to;
+        this.transition = transition;
+    }
+
+    static parse(input: Projection | string | undefined | null): Projection | undefined {
+        // in zoom-and-property function input could be an instance of Color class
```
HarelM, line 14: Is this comment relevant here?

src/util/projection.ts
```diff
@@ -0,0 +1,51 @@
...
+    readonly from: PrimitiveProjection;
+    readonly to: PrimitiveProjection;
+    readonly transition: number;
+    constructor(from: PrimitiveProjection, to: PrimitiveProjection, transition: number){
+        this.from = from;
+        this.to = to;
+        this.transition = transition;
+    }
+
+    static parse(input: Projection | string | undefined | null): Projection | undefined {
```
HarelM, line 13: Is this method being used somewhere in the code? I haven't seen it, I'm guessing it is important, right?

src/validate/validate_projection.ts
```diff
@@ -26,11 +26,12 @@ export default function validateProjection(options: ValidateProjectionOptions) {
 
     let errors = [];
     for (const key in projection) {
-        if (projectionSpec[key]) {
+        console.log('KEY', key, projection[key])
```
HarelM, line 29: Remove console.log

src/validate/validate_projection.ts
```diff
@@ -14,7 +14,7 @@ interface ValidateProjectionOptions {
 export default function validateProjection(options: ValidateProjectionOptions) {
     const projection = options.value;
     const styleSpec = options.styleSpec;
-    const projectionSpec = styleSpec.projection;
+    const projSpec = styleSpec.projection;
```
HarelM, line 17: Not sure I understand this change...

src/validate/validate.ts
```diff
@@ -47,6 +48,7 @@ const VALIDATORS = {
     'sky': validateSky,
     'terrain': validateTerrain,
     'projection': validateProjection,
+    'projection-mode': validateProjectionMode,
```
HarelM, line 51: Not sure I fully understand this definition. How would that look in a style json file?

birkskyum:
```
{
  sky: {}
  terrain: {}
  projection: { 
    mode: ['step', ['zoom'], 'globe', 10, 'mercator']
  }
}
```

docs/types.md
birkskyum: okay, let's keep the nesting so we can expand projection settings down the line. I think it's best we do the rename to `mode`, because `type` is a reserved word, just like in typescript, due to it's role in type validation:

https://github.com/maplibre/maplibre-style-spec/blob/5788145317e1b7cf444a324207822706eec9071a/src/validate/validate.ts#L87-L102

src/util/interpolate.ts
```diff
@@ -23,10 +24,11 @@ export function isSupportedInterpolationColorSpace(colorSpace: string): colorSpa
  * @returns interpolation fn
  * @deprecated use `interpolate[type]` instead
  */
-export const interpolateFactory = (interpolationType: 'number'|'color'|'array'|'padding'|'variableAnchorOffsetCollection') => {
+export const interpolateFactory = (interpolationType: 'number'|'color'|'projection-mode'|'array'|'padding'|'variableAnchorOffsetCollection') => {
     switch (interpolationType) {
         case 'number': return number;
         case 'color': return color;
+        case 'projection-mode': return projection;
```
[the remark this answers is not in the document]
birkskyum: it's projectionMode now. Apparently the spec has some autogeneration that can't handle the dash, but a camelcase like `projectionType` will be auto converted to ProjectionTypeSpecification, and lowercased for url's.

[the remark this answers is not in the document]
birkskyum: The `projection` type is used for the root "projection" key. If you prefer, then I can rename all the root property "projectionConfig" (it's never shown on the website), and reserve the "projection" for the projection.type key you point to.

birkskyum:
@HarelM , I think it might have a design now. I'd recommend building the docs and go here
http://0.0.0.0:8000/maplibre-style-spec/types/#projectiontransition

docs/types.md
```diff
@@ -133,3 +133,90 @@ The following example applies 2em padding on top and bottom and 3em padding left
...
+The `projectionTransition` is used to configure which projection to use for the map.
+
+There are currently two projections implemented.
+
+- `mercator` - [Web Mercator projection](https://en.wikipedia.org/wiki/Web_Mercator_projection)
+- `stereographic` - [Stereographic projection](https://en.wikipedia.org/wiki/Stereographic_projection)
+
+And the following [presets](#use-a-projection-preset)
+
+- `globe` - Globe projection
```
HarelM, line 150: Is this in the right place?

birkskyum: It's in the `v8.projection.type.values` - I pull it from there and display it on the projection site as well

https://maplibre.org/maplibre-style-spec/projection/

But the types.md doesn't allow for that, so I just wrote it by hand there too.

HarelM: I get it now as there is only one preset at the moment. 
But since it is explained below, I would only link to it as you did and remove this part.

birkskyum: ah, yes. remove it

HarelM:
Thanks! I think it is in a good direction, but still requires a small change. 
I'll try to explain: 
As far as I understand, the transition is caused by the expression syntax, if we are using "mercator", or [mercator, mercator, 1], it is still a protection, so is [mercator, stereographic, 0.5]. It defines how to project a point from one coordinate system to the other. 
Much like color, that you can define as [red, green, 0.5] would be a color that has 128 red and 128 green, which is still a color and not a color transition. 
So the expression can create a "method" that receives zoom and returns a protection, but but a protection transition.
I understand the v8 definition issue, but if the object is internal and not exposed I worked call it projectionSpecification in the v8 file, which would be converted to ProjectionSpecification in typescript. 
I hope this helps, let me know if you want to have a quick call on this, or even in slack, whichever works for you. 

Thanks for all the effort put into it!

birkskyum:
@HarelM I've updated it so the type of the root projection property is projectionConfig, allowing for the projection type to be used for the projection value. 

see https://github.com/maplibre/maplibre-style-spec/pull/888#issuecomment-2461318216

build/generate-style-spec.ts
```diff
@@ -317,9 +329,9 @@ ${objectDeclaration('LightSpecification', spec.light)}
 
 ${objectDeclaration('SkySpecification', spec.sky)}
 
-${objectDeclaration('TerrainSpecification', spec.terrain)}
+${objectDeclaration('ProjectionConfigSpecification', spec.projectionConfig)}
```
HarelM, line 332: Not sure I like it, but I don't think I have a better idea how to solve it.
I see that color is using ColorSpecification, so ProjectionSpecification is save to the spec of the projection object used inside the root projection definition.
the only idea that comes to my mind is `RootProjection` but that's not good...
Feel free to resolve this comment if you can't this of a better name for this type... I can't think of one right now...

src/expression/definitions/interpolate.ts
```diff
@@ -171,9 +174,15 @@ class Interpolate implements Expression {
         const outputLower = outputs[index].evaluate(ctx);
         const outputUpper = outputs[index + 1].evaluate(ctx);
 
+        if (this.type.kind === 'projection') {
```
HarelM, line 177: Is this if needed given the below switch case? I find this "if" odd here...

src/expression/values.ts
```diff
@@ -112,11 +116,11 @@ export function toString(value: Value) {
...
-    } else if (value instanceof Color || value instanceof Formatted || value instanceof Padding || value instanceof VariableAnchorOffsetCollection || value instanceof ResolvedImage) {
+    } else if (value instanceof Color || value instanceof Projection || value instanceof Formatted || value instanceof Padding || value instanceof VariableAnchorOffsetCollection || value instanceof ResolvedImage) {
         return value.toString();
     } else {
         return JSON.stringify(value);
     }
 }
 
-export {Color, Collator, Padding, VariableAnchorOffsetCollection};
+export {Color, Collator, Projection as Projection, Padding, VariableAnchorOffsetCollection};
```
HarelM, line 126: No need for "as", this probably due to a rename...

src/index.ts
```diff
@@ -81,6 +81,7 @@ import featureFilter, {isExpressionFilter} from './feature_filter';
 
 import convertFilter from './feature_filter/convert';
 import Color from './util/color';
+import {Projection} from './util/projection';
```
HarelM, line 84: Others are "default export"ed, to align with other files I would consider doing the same here, and maybe afterwards do a PR to change things to named exports.

src/reference/v8.json
```diff
@@ -4568,19 +4591,33 @@
...
+      "expression": {
+        "interpolated": true,
+        "parameters": [
+          "zoom"
+        ]
+      },
       "values": {
-        "mercator": {
-          "doc": "The Mercator projection."
+        "projections": {
```
HarelM, line 4606: I would remove these (projections, presets), as they are proprietary to this type and are not validated or checked, right?
If this is the case, and this is added only for documentation purposes, use the docs with markdown to extend how the docs will look.

src/util/projection.ts
```diff
@@ -0,0 +1,45 @@
...
+export function isProjectionValue(value: unknown): value is ProjectionT {
+    return Array.isArray(value) &&
+        value.length === 3 &&
+        isProjectionPrimitive(value[0]) &&
+        isProjectionPrimitive(value[1]) &&
+        typeof value[2] === 'number';
+}
+
+export function isProjectionPrimitive(value: unknown): value is ProjectionPrimitiveT {
+    return Object.keys(v8.projectionConfig.type.values.projections).includes(value as string);
```
HarelM, line 40: I now see that this is used here, is there an equivalent for this in the style spec?
If not, I would consider hardcoding it here, although I don't like to hardcode stuff, setting it in the v8.json file and parsing it in all kinds of places (think about the "i" in maputnik for example) is not a great experience.
I don't know... this is only a though...

HarelM:
I'm not sure I fully understand why we can't merge this work soon? 
Yes, there are other projections, we won't support those at the moment, we need to narrow down the scope to support the expression related to globe to mercator transition to allow configuring this explicitly instead of implicitly. 
Later on we might be able to support other use cases. 
Does that makes sense?

birkskyum:
I've been thinking a bit, and maybe instead of config object for _each_ projection, if we can just allow them to be string like this, with +proj + wkt format (or a shorthand string we can shap for it), then I think we're probably well enough:

```ts
const equalEarthProjection = 'PROJCS["World_Equal_Earth",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]],PROJECTION["Equal_Earth"],PARAMETER["false_easting",0],PARAMETER["false_northing",0],PARAMETER["longitude_of_center",0],PARAMETER["latitude_of_center",0],UNIT["Meter",1]]';
const webMercatorProjection = 'PROJCS["WGS_1984_Web_Mercator_Auxiliary_Sphere",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]],PROJECTION["Mercator_1SP"],PARAMETER["central_meridian",0],PARAMETER["scale_factor",1],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["Meter",1],AUTHORITY["EPSG","3857"]]';
const verticalPerspectiveProjection = '+proj=geos +lat_0=0 +lon_0=0 +h=35785831 +datum=WGS84 +units=m +no_defs';
```

On the GL JS implementation side, I believe we should consider using a stable library like proj4.js instead of reimplementing a coordinate transformation logic inside GL JS. It comes in at 93 kb
Screenshot 2024-11-10 at 18 42 32

Here's a small proj4.js demo i made, where I only add the lerp/animation logic, but the projection math is the proj4.js inside the browser client:
https://github.com/birkskyum/proj4js-demo

HarelM:
Yes, that's what I thought as well, besides the fact that the protection is happening in the shaders where we can't use proj4js I think this was my direction as well.
We still need to support an expression to transition between the projections, but defining the projections using WKT is what I was aiming for (we don't have to do it now I belive, but this is how I see the spec handles projections definitions).

birkskyum:
I think there might be some interesting ways to go about this:
https://github.com/rjw57/proj4webgl

I'm happy to explore anything that can keep the implementation of an advanced coor transform library outside the scope of GL JS. Could be quite the time sink.

HarelM:
I don't think we should solve all projections now. We should only make sure we support strings so if in the future we want to support arbitrary protection, we would not need to change the spec definition.
It's probably possible to do protection in the shaders like Jakub did for globe, but I won't try anything now, and I agree it is a huge lift.
Just make sure we can explicitly solve the definition of zoom transition between globe  and mercator.

birkskyum:
Good - I'll move on with our projection expression parsing, giving each level a zoom level string value for the expression.

kubapelc:
Some of my thoughts, sorry for posting so late in the discussion.

I'd like to point out this comment: https://github.com/maplibre/maplibre-style-spec/issues/568#issuecomment-1998113647

> Web Mercator is a deceptively simple projection. You can project back and forth with fairly simple functions and no iteration, no weird gridded offsets, that sort of thing. As such it can be unhelpful to design around. For example, doing the work in a shader is pretty easy, but much less so for something like UTM with a specialized ellipsoid and some strange offsets.

It will likely not be feasible to implement all projections in a shader that projects the geometry in realtime.

Speculation about a future solution: assuming a given projection would not change if the map is translated or zoomed (eg. globe is highly dependent on both, but mercator is simply translated or scaled), such projection could be implemented by projecting the geometry on the CPU in javascript and then drawing it with a relatively simple shader. But vertex buffers for both transitioning projections would need to be present and ready at the same time, requiring more processing to prepare them for each tile. This could even be done with WebGPU compute shaders for extra performance, otherwise it would slow down tile loading even more.

But as for WebGPU and compute shaders, keep in mind that at least on desktop, consumer GPUs only really work well in float32, and while float64 is supported, it is often many times slower than float32 (32x slower on current AMD and 64x slower on current Nvidia for example, looking at wikipedia TFLOPS numbers). Float64 would likely be needed for many projections. Even globe would benefit from it already, the transition to mercator is in place to hide float precision artifacts that result from only using float32.

> For reference, the globe costed 88kb, and some of that weight could be taken out.

I think most of globe code increase comes from refactoring everything to support multiple projections, making the camera controls work nicely on a sphere, adapting symbols and such... the actual projection code is fairly small. Hopefully a lot of that could be reused for future projections.

I personally think projection expressions make sense, even if only to allow customization of the globe-mercator transition.

birkskyum:
Thanks @kubapelc , that's very interesting points. I haven't seen the PROJJSON before - i don't know if it makes sense to have in the spec itself, of it we can just make a PROJ -> PROJJSON in the client though, and keep it a string here.

Camera Controls - this is going to be a bit tricky on the implementation side. When receiving some random PROJ string, we'll have to infer if it's i.e. globe controls or not. It's not something we have to have a plan for for the spec to move forward, but I can see quite a lot of gotchas with specifically for _random_ strings.

kubapelc:
> Camera Controls - this is going to be a bit tricky on the implementation side

This might not be that hard. Globe is special since it is a sphere. Mercator controls are already designed for movement around a plane, and it should not matter too much how the map data is projected onto that plane, so planar projections could reuse mercator camera.

src/util/projection.test.ts
```diff
@@ -0,0 +1,9 @@
+import {Projection} from './projection';
+
+describe('Projection class', () => {
+
+    test('should serialize to rgba format', () => {
```
HarelM, line 5: Probably not the right test description...

src/validate/validate.ts
```diff
@@ -46,6 +47,7 @@ const VALIDATORS = {
     'light': validateLight,
     'sky': validateSky,
     'terrain': validateTerrain,
+    'projectionConfig': validateProjectionConfig,
```
HarelM, line 50: Should this be a different validation than "regular" object?

birkskyum: Yes, 

validateProjectionConfig check a value of the type:

{ type: projection }

validateProjection check a value of the type:

projection

birkskyum: Or is there's a regular regular object fallback I can use, that just check the `type`, so we don't need this?

HarelM: I don't know, worth a check, if not, then this is OK as is.

birkskyum:
@HarelM , it would appear that I got the `interpolate-projection` working (!)

So, basically it'll provide the ["vertical-perspective", "mercator", 0.4] if between zoom levels, and if outside a transition, say the last top is zoom 10 = "mercator", and you're at zoom 12, then it's just returning "mercator", similar to step function.

I believe this is consistent with the other interpolation functions.

birkskyum:
I haven't tried to use it yet in MapLibre GL JS, but it's ready for a trial run.

I imagine it's the style->update() that's the best place to hook into.

birkskyum:
https://github.com/user-attachments/assets/4382040a-eacb-4b60-abba-188d10cbfba0

It works 🎉

Use `npm link` and added this in the top of the style->update()

```ts
    update(parameters: EvaluationParameters) {

        const exprestInt = createExpression(['interpolate-projection', ['linear'], ['zoom'], 10, 'vertical-perspective', 12, 'mercator'],v8)

        if (exprestInt.result === 'success') {
            console.log(exprestInt.value.evaluate({zoom: parameters.zoom}))

        } else {
            console.warn('Error parsing interpolate-projection')
        }
```

src/expression/definitions/interpolate.ts
birkskyum: Alternatively, we can do :

```ts
 switch (this.operator) {
            case 'interpolate':
 //  The line below this complains if typescript thinks this.type.kind can equal "projection"
                return interpolate[this.type.kind](outputLower, outputUpper, t);  
            case 'interpolate-hcl':
                return interpolate.color(outputLower, outputUpper, t, 'hcl');
            case 'interpolate-lab':
                return interpolate.color(outputLower, outputUpper, t, 'lab');
        }
```

src/reference/v8.json
birkskyum: They are actually validated right now because of

https://github.com/maplibre/maplibre-style-spec/pull/888/files#diff-22cb3d567c884a921b94b7964eecbaa9b22ee8b0be35ba1e7118b484b3c0dadaR129-R130

I can prefer this if it's not wanted.

src/util/interpolate.ts
birkskyum: I'm not sure still why we want to convert "mercator" to ["mercator", "mercator", 1], and then convert it back to "mercator" and use the value.

src/util/projection.ts
birkskyum: Alright, I'll remove the projections/primitives. It won't fly for WKT / +proj anyway, and it'll get out of sync with Native since it's not possible for us to make "sdk-support" on a value level.

src/reference/v8.json
birkskyum: Removed it

build/generate-docs.ts
```diff
@@ -218,6 +221,22 @@ function convertPropertyToMarkdown(key: string, value: JsonObject, keyPrefix = '
 
     markdown += '. ';
 
+    // Projection type:
+    const isProjection = value.type === 'projection' && value.values && !Array.isArray(value.values);
```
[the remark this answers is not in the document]
birkskyum: Screenshot 2024-11-13 at 03 31 53

removeing the presets / projections form v8, this goes away - there's a link to the Projection Type with all the details form types.md

src/expression/definitions/interpolate.ts
```diff
@@ -170,6 +173,10 @@ class Interpolate implements Expression {
 
         const outputLower = outputs[index].evaluate(ctx);
         const outputUpper = outputs[index + 1].evaluate(ctx);
+        
+        if (this.type.kind == 'projection') {
```
HarelM, line 177: I think this can be added to the below switch-case?

src/validate/validate_projectionconfig.test.ts
```diff
@@ -0,0 +1,49 @@
+import validateProjectionConfig from './validate_projectionconfig';
+import validateSpec from './validate';
+import v8 from '../reference/v8.json' with {type: 'json'};
+
+describe('Validate projection', () => {
```
HarelM, line 5: `projectionConfig`?

build/generate-docs.ts
```diff
@@ -180,6 +181,8 @@ function typeToMarkdownLink(type: string): string {
         case 'paint':
         case 'layout':
             return ` [${type}](layers.md#${type.toLocaleLowerCase()})`;
+        case 'projectionconfig':
+            return ' [projectionConfig](projection.md)';
```
HarelM, line 185: Is this really needed here?
This is used for the docs to allow link to an object type related to the expressions I think, and I'm not sure projectionConfig can be a return type of an expression, right?

HarelM:
This looks great!!! thanks!
I added a few minor comments, otherwise approved.

src/expression/definitions/interpolate.ts
birkskyum: A type guard is needed, so it turns the switch into:

```ts
        switch (this.operator) {
            case 'interpolate':
                if (this.type.kind !== 'projection') {
                    return interpolate[this.type.kind](outputLower, outputUpper, t);
                }
            case 'interpolate-hcl':
                return interpolate.color(outputLower, outputUpper, t, 'hcl');
            case 'interpolate-lab':
                return interpolate.color(outputLower, outputUpper, t, 'lab');
            case 'interpolate-projection': {
                return interpolate.projection(outputLower, outputUpper, t);
            }
        }
    ```

birkskyum: This is why.

Screenshot 2024-11-13 at 08 44 34

birkskyum: The stupid part is that this, or insertion of any of the other possible string values for kind, doesn't have a type problem

Screenshot 2024-11-13 at 08 51 59

birkskyum:
```ts
case 'interpolate':
      return interpolate[this.type.kind.toString()](outputLower, outputUpper, t);
  case 'interpolate-hcl':
      return interpolate.color(outputLower, outputUpper, t, 'hcl');
  ```
  
  A `kind.toString()` resolves it

build/generate-docs.ts
birkskyum: This is an overwrite so that the url of the main Projection page doesn't get a /projectionConfig url.

Screenshot 2024-11-13 at 09 04 55

If the line is removed this pops up
`WARNING -  Doc file 'root.md' contains a link 'projectionConfig.md', but the target is not found among documentation files.`

HarelM: Ah, I see...
I'm trying to think if there's a way to solve this elegantly, but I can't think of a good idea at the moment.

src/expression/definitions/interpolate.ts
HarelM: I think this removes the type guard from the `interpolate` object, which I'm not sure is what we want.
I generally try to avoid using `object[property]` as much as possible due to these limitation and the fact that it's harder to read the code this way.

HarelM:
Can you create a branch in this repo and merge this PR to this new branch so I'll see if I can try out some things in order to solve the parts that still annoys me?

src/expression/definitions/interpolate.ts
HarelM: I like this one better to be honest.

HarelM: By the way, if this is possible for other types, why do we need an extra case for interpolate-projection then?

HarelM:
Please also checkout the code coverage report, there are other stuff that are not covered there, which we need to think how to address, maybe out of scope for this PR.
But for example `interpolateFactory` is not used, maybe we should remove it instead of adding more logic to it...

build/generate-docs.ts
HarelM: I'm not sure this is a good idea, but in proj4js they use the term `ProjectionDefinition` for the values related to the projection:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/proj4/index.d.ts

I'm not sure what I think about it though - this is for the "internal" type, the one used in the expressions, not the root projection type.

birkskyum:
I've made a `projection-expression-collecting-branch` in this repo, and changed the `base` branch of this repo to it - is that what you meant?

HarelM:
Yes, thanks!

PR #899, opened 2024-11-13
Add projection expression syntax (#888)

birkskyum:
This is a merging branch for
- #888 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
See my additions here:
- https://github.com/maplibre/maplibre-style-spec/pull/901

birkskyum:
The only problem I really have before merging this is that the `interpolate-projection` ran in GL JS, but the same expression now with `interpolate` is breaking...

birkskyum:
> Wrt. Projection -> ProjectionDefinition

I think that's a great idea - will get to it.  "Projection" is overloaded by the class in geo/projection.

build/generate-style-spec.ts
```diff
@@ -38,6 +38,8 @@ function propertyType(property) {
                 return 'SkySpecification';
             case 'sources':
                 return '{[_: string]: SourceSpecification}';
+            case 'projection:':
+                return 'ProjectionDefinitionSpecification';
```
HarelM, line 42: Shouldn't this be `ProjectionSpecification`?

src/reference/v8.json
```diff
@@ -127,9 +127,15 @@
...
       "type": "projection",
-      "doc": "The projection configuration. **Note:** this definition is still experimental and is under development in maplibre-gl-js.",
+      "doc": "The projection configuration",
       "example": {
-        "type": "globe"
+        "type": [
+          "interpolate",
+          ["linear"],
+          ["zoom"],
+          10, "globe",
```
HarelM, line 136: `globe` is not the right example, right? should be `vertical-perspective`...?

src/util/interpolate.ts
```diff
@@ -23,10 +24,11 @@ export function isSupportedInterpolationColorSpace(colorSpace: string): colorSpa
  * @returns interpolation fn
  * @deprecated use `interpolate[type]` instead
  */
-export const interpolateFactory = (interpolationType: 'number'|'color'|'array'|'padding'|'variableAnchorOffsetCollection') => {
+export const interpolateFactory = (interpolationType: 'number'|'color'|'projectionDefinition'|'array'|'padding'|'variableAnchorOffsetCollection') => {
```
HarelM, line 27: Can this factory be removed? It is not used in the code as far as I know...?

src/util/projection.ts
```diff
@@ -0,0 +1,51 @@
+import {ProjectionDefinitionT, ProjectionDefinitionSpecification, PropertyValueSpecification} from '../types.g';
+
+export default class ProjectionDefinition extends Array {
```
HarelM, line 3: I'm not sure it makes sense to inherit from `Array`... methods like include/shift/pop are not relevant for this class...

src/util/projection.ts
```diff
@@ -0,0 +1,51 @@
...
+            return new ProjectionDefinition(input[0], input[1], input[2]);
+        }
+        if (typeof input === 'string') {
+            return new ProjectionDefinition(input, input, 1);
+        }
+        return undefined;
+    }
+}
+
+export function isProjectionDefinition(value: unknown): value is ProjectionDefinitionSpecification {
```
HarelM, line 33: Are all these exported methods still needed?

src/util/interpolate.ts
```diff
@@ -37,6 +39,10 @@ function number(from: number, to: number, t: number): number {
     return from + t * (to - from);
 }
 
+function projection(from: string, to: string, interpolation: number): ProjectionDefinition {
```
HarelM, line 42: Other method names math the type, consider changing this method badge as well.

src/util/interpolate.ts
birkskyum: i think we can remove the entire interpolate.projection() function, since it's a passthrough

birkskyum:
Screenshot 2024-11-14 at 00 15 23

it's working again , with

src/util/interpolate.ts
birkskyum: i've remove the function, we're fine with the default passthrough case here:

```ts
        switch (this.operator) {
            case 'interpolate':
                switch (this.type.kind) {
                    case 'number':
                        return interpolate.number(outputLower, outputUpper, t);
                    case 'color':
                        return interpolate.color(outputLower, outputUpper, t, 'rgb');
                    case 'array':
                        return interpolate.array(outputLower, outputUpper, t);
                    case 'padding':
                        return interpolate.padding(outputLower, outputUpper, t);
                    case 'variableAnchorOffsetCollection':
                        return interpolate.variableAnchorOffsetCollection(outputLower, outputUpper, t);
                    default:
                        return [outputLower, outputUpper, t];  // <-- New one
                }
            case 'interpolate-hcl':
                return interpolate.color(outputLower, outputUpper, t, 'hcl');
            case 'interpolate-lab':
                return interpolate.color(outputLower, outputUpper, t, 'lab');
        }
    }
```

src/util/projection.ts
birkskyum: managed to remove some. think the rest are

birkskyum:
I don't know if there's more to it.

We could cut down on the ProjectionDefinition types, by replacing it with more generic String / StringExpression // stringExpression kind of type - but that's all I can think of, and I believe it wouldn't affect users if this was done under the hood in the style spec typescript later on.

birkskyum:
There is one thing I'm wondering, and it's if we should allow returning

1. string
2. [string, string, number]
3. [expression, values for expression]. // interpolate | step

of if we should take away no. 2, and use a literal expression instead

1. string
2. [expression, values for expression] // interpolate | literal | step

birkskyum:
Maybe the array values in step could make sense to add

src/expression/definitions/interpolate.ts
```diff
@@ -170,10 +171,23 @@ class Interpolate implements Expression {
...
+                        return interpolate.number(outputLower, outputUpper, t);
+                    case 'color':
+                        return interpolate.color(outputLower, outputUpper, t, 'rgb');
+                    case 'array':
+                        return interpolate.array(outputLower, outputUpper, t);
+                    case 'padding':
+                        return interpolate.padding(outputLower, outputUpper, t);
+                    case 'variableAnchorOffsetCollection':
+                        return interpolate.variableAnchorOffsetCollection(outputLower, outputUpper, t);
+                    default:
```
HarelM, line 188: I'm not sure this should be a "default"...

src/expression/definitions/interpolate.ts
```diff
@@ -126,14 +126,15 @@ class Interpolate implements Expression {
...
             }
-
             const parsed = context.parse(value, valueKey, outputType);
             if (!parsed) return null;
             outputType = outputType || parsed.type;
             stops.push([label, parsed]);
         }
 
         if (!verifyType(outputType, NumberType) &&
+            !verifyType(outputType, StringType) &&
```
HarelM, line 136: This looks wrong code-wise - we should not support arbitrary string interpolation.

src/expression/expression.test.ts
```diff
@@ -629,3 +629,67 @@ describe('slice expression', () => {
...
 });
+
+describe('projection expression', () => {
+
+    let value: StyleExpression;
+    let response: any;
+
+    beforeEach(() => {
+        response = createExpression(
+            ['distance', {
```
HarelM, line 640: How is `distance` expression related to this change?

src/index.ts
```diff
@@ -67,6 +67,12 @@ export type StylePropertySpecification = {
     expression?: ExpressionSpecificationDefinition;
     transition: boolean;
     default?: VariableAnchorOffsetCollectionSpecification;
+} | {
+    type: 'projectionDefinition';
+    'property-type': ExpressionType;
+    expression?: ExpressionSpecificationDefinition;
+    transition: boolean;
+    default?: string;
```
HarelM, line 75: How would you represent a state "between" projections using just string?

src/util/projection.ts
HarelM, line 1: Consider changing the file name to match the class name

src/util/projection.ts
```diff
@@ -0,0 +1,28 @@
+export default class ProjectionDefinition {
+    constructor(from: string, to: string, transition: number){
+        this[0] = from;
```
HarelM, line 3: This is strange for typescript, why not use "this.from" etc?

birkskyum: it'll serialize as:
{from:this.from , to:this.to , t:this.t}

The current constructor will serialize as array
[from, to, t]

src/index.ts
birkskyum: I think the current default value is "mercator"

src/util/projection-definition.ts
HarelM, line 1: Other places in the code use "_" for file names I think...

src/expression/definitions/interpolate.ts
birkskyum: We do support arbitrary values for `step` , and I'm not sure of the discrepancy between the two.

src/util/projection.ts
HarelM: Serialization is defined in `toString` isn't it? I prefer to know which fields a class has when reading it, this array like thing is confusing, to me...

src/expression/definitions/interpolate.ts
HarelM: Let's not introduce string interpolation as part of this PR. If there's a need for it, let's add it in the future.

src/util/projection.ts
birkskyum: It only became "confusing Array like thing" when I removed the `extends Array` per your request. Before that it was an actual Array.

src/index.ts
HarelM: No, the default should be `mercator`. I just don't think the type of it should be `string`, but instead be `ProjectionDefinition` or `ProjectionDefinitionSpecification`.

src/util/projection.ts
HarelM: I still think it should be a regular class, I don't understand the value of having it look like an array internally, can you explain the added value of this?
```ts
export default class ProjectionDefinition {
  constructor(private from: string, private to: string, private transition: number) {}

  toString() {
    return ...
  }
  // maybe getter here, or simply make the properties public
  ...
  
}
```

src/index.ts
birkskyum: it's ProjectionDefinitionSpecification now

src/expression/definitions/interpolate.ts
birkskyum: OK, i'll attempt to add back the .projection() interpolation

src/util/projection.ts
birkskyum: If we can make it serialized correctly then it's fine with me to have the from / to / t

HarelM: Where is this serialization happening? What needs to be added in order to make it work like you would like it to work?

src/util/projection_definition.ts
HarelM, line 1: I saw a folder in the expressions called `types`, I saw other expression types there, maybe this should be there instead of here.
Another idea (out of scope for this PR) that came to my mind is that maybe we can generate the types.md file from this folder automatically using some tsdocs comments and some processing.

src/util/projection.ts
birkskyum: It's here

https://github.com/maplibre/maplibre-style-spec/blob/4ebf7975507637282b59b37a156cb024578f15ea/test/integration/expression/expression.test.ts#L60

birkskyum:
@HarelM  I'm struggling to get this working, and I believe it's because of the hard constraint that strings aren't allowed as interpolate values. I think we might have to either allow string to be interpolatable values, or abandon the otherwise minimal syntax below, that use strings as interpolate values:

`["interpolate", ["linear"], ["zoom"], 10, "+proj 1", 12 "+proj 2"]`

When it's being parsed, the "+proj 1" is being recognized as `kind: "string"`. How can we possibly tell the parser that this particular string is a `kind: "projectionDefinition"`?

Thus, I'm effectively trying to deliver string interpolate support, under the limitation that string is not an interpolatable type. This is proving quite difficult to work around. Feel free to have a go at it if you have ideas. Basically, all the solutions I come up with ends up boiling down to new ways of skipping the string interpolate check, that will otherwise throw.

src/expression/expression.test.ts
```diff
@@ -629,3 +629,43 @@ describe('slice expression', () => {
...
+            expect(response.value.evaluate({zoom: 5})).toStrictEqual(['vertical-perspective', 'mercator', 0.5]);
+            expect(response.value.evaluate({zoom: 10})).toBe('mercator');
+            expect(response.value.evaluate({zoom: 11})).toBe('mercator');
+        } else {
+            throw new Error('Failed to parse Step expression');
+        }
+    })
+
+    test('interpolate', () => {
+        const response = createExpression(['interpolate', ['linear'], ['zoom'], 8, 'vertical-perspective', 10, 'mercator']);
```
birkskyum, line 660: since we're working with strings as interpolate values, it's quite tricky to have a constraint that doesn't allow strings as interpolate values. The "vertical-perspective" will be parsed into `kind`: 'string'` rather than kind: projectionDefinition, and I don't know how I can pass the projectionDefinition information

HarelM: This is not different than color parsing from string. The kind should be coming from the spec definition.

test/integration/expression/tests/step/projection/step-array/test.json
```diff
@@ -0,0 +1,44 @@
+{
+    "propertySpec": {
+      "type": "string",
```
HarelM, line 3: This should be `projectionDefinition` I believe.

src/expression/expression.test.ts
HarelM: I haven't see color parsing in these tests, I'm guessing it is related to the fact the `kind` need to be set and this is done in the integration tests...

HarelM:
I'm working to move the types to the right place in the folder structure, to make things a bit more organized. Just FYI...

src/expression/expression.test.ts
HarelM: It's true that color has a way to not the expression engine that this part should be cast to type color.
But when using in the style you can simply use "red", "green" etc, this is similar to `mercator`.
We could in theory in the future add a `proj` operator like rgba, but I don't think we need it now.
The epxression engine knows the expected type based on the kind from the spec.
If you would try to do the same test with `createExpression(['interpolate', ['linear'], ['zoom'], 0, 'red', 10, 
'green']);` I believe it will fail like it would for expression.

birkskyum:
> I'm working to move the types to the right place in the folder structure, to make things a bit more organized. Just FYI...

Thanks, that does sounds great. Are you working on it in a separate cleanup PR, or on this branch?

HarelM:
Separate PR, I should be done in a few minutes...

HarelM:
See #903 - it has the switch expanded like in this PR, so this can be cleaner here, let me know if you have issues resolving conflicts.

birkskyum:
Update:
I'm half-ways through refactoring to named exports, in separate PR. I find it'll be faster to work when that's in order, because we rename a lot (Projection -> ProjectDefinition etc.), and a named export will auto-propagate these change to the import statements and keep things consistent, where a default export does no such thing, so typecheck can't help.


Will wrap that up and get back to this

HarelM:
I had the same observation lately when working with the code, this is needed indeed. Thanks!

src/expression/definitions/interpolate.ts
```diff
@@ -22,8 +22,7 @@ export type InterpolationType = {
     name: 'cubic-bezier';
     controlPoints: [number, number, number, number];
 };
-type InterpolatedValueType = NumberTypeT | ColorTypeT | PaddingTypeT | VariableAnchorOffsetCollectionTypeT | ArrayType<NumberTypeT>;
-
+type InterpolatedValueType = NumberTypeT | ColorTypeT | StringTypeT | ProjectionDefinitionTypeT | PaddingTypeT | VariableAnchorOffsetCollectionTypeT | ArrayType<NumberTypeT>;
```
HarelM, line 25: Not sure I would expect `StringTypeT` to be here, if this wasn't needed for color (which can be a string) I'm not sure I understand why we would need it now.

src/expression/types/projection_definition.test.ts
```diff
@@ -0,0 +1,9 @@
+import {ProjectionDefinition} from './projection_definition';
+
+describe('Projection class', () => {
```
HarelM, line 3: I would expect parse and interpolate to also be tested here.

src/expression/types/projection_definition.ts
```diff
@@ -0,0 +1,36 @@
+export class ProjectionDefinition {
+    public from
```
HarelM, line 2: You can add `public` modifier before every field in the constructor to save the need to write this.
But if you don't please specify the type here.

src/expression/values.ts
```diff
@@ -5,9 +5,10 @@ import {Formatted} from './types/formatted';
 import {Padding} from './types/padding';
 import {VariableAnchorOffsetCollection} from './types/variable_anchor_offset_collection';
 import {ResolvedImage} from './types/resolved_image';
-import {NullType, NumberType, StringType, BooleanType, ColorType, ObjectType, ValueType, CollatorType, FormattedType, ResolvedImageType, array, PaddingType, VariableAnchorOffsetCollectionType} from './types';
+import {NullType, NumberType, StringType, BooleanType, ColorType, ObjectType, ValueType, CollatorType, FormattedType, ResolvedImageType, array, PaddingType, VariableAnchorOffsetCollectionType, ProjectionDefinitionType} from './types';
 
 import type {Type} from './types';
+import {ProjectionDefinition} from './types/projection_definition';
```
HarelM, line 11: nit picking - move this import next to the other imports of color, resolvedImage etc...
Keep the `import type` to be separated and last.

src/index.ts
```diff
@@ -34,11 +34,12 @@ import {typeOf} from './expression/values';
 import {FormatExpression} from './expression/definitions/format';
 import {Literal} from './expression/definitions/literal';
 import {CompoundExpression} from './expression/compound_expression';
-import {VariableAnchorOffsetCollectionSpecification} from './types.g';
+import {ProjectionDefinitionSpecification, VariableAnchorOffsetCollectionSpecification} from './types.g';
```
HarelM, line 37: I would expect to see here more specifications imports like padding? color?
I think the `StylePropertySpecification` can be improved a bit to use those...?

HarelM: I've fixes that

src/expression/types/projection_definition.ts
HarelM: I've fixed that too now.

test/integration/expression/tests/step/projection/step-array/test.json
HarelM: Changing this solved the test failure.

src/expression/expression.test.ts
HarelM: I've fixed the tests by adding the relevant property spec parameter to help the expression engine understand that projection definition is expected.

HarelM:
I've fixed the tests and addressed the comments I found when reviewing.
Feel free to review this.
I think this is ready to be merged.
Let me know if you are able to use it in maplibre-gl-js.
Note that I have removed the serialization to array of `projectionDefinition` as I'm not sure where this is needed.
I do see that color does get serialized to array somehow (saw it in `test/integration/expression/tests/interpolate-hcl/linear/test.json` test), but I couldn't figure out where it happens.

birkskyum:
It's running fine in GL JS, with this syntax

```ts
const exprestInt = createExpression(['interpolate', ['linear'], ['zoom'], 10, 'vertical-perspective', 12, 'mercator'], v8.projection.type)
```

https://github.com/user-attachments/assets/14c74c44-7d9b-4095-986a-9f75f1d7e23f

HarelM (approved):
Great!

Issue #832, opened 2024-09-25
Vertical alignment in format expression


stanislawpuda-tomtom:
# Design Proposal: Add vertical alignment to format expression

## Motivation

[Format expression](https://maplibre.org/maplibre-style-spec/expressions/#format) can facilitate multiple text and image elements of different sizes. At the moment all of them are aligned to the font baseline. It is not possible to change the vertical alignment. Having a possibility to change the alignment, eg. center, would allow us to create rich labels.

If following design would be accepted I would like to make contributions to this repo and MapLibre GL JS with the implementation.

## Proposed Change

We would like to introduce `vertical-align` property to style override in `format` expression. `vertical-align` property can take three options:
- `bottom` *(default)*
- `center`
- `top`

#### Center

```JavaScript
"text-field": [
  "format",
  ["image", "image"], 
  { "vertical-align": "center" },
  " ",
  {},
  "BIG yg",
  { "font-scale": 1.2, "vertical-align": "center" },
  " ",
  {},
  "SMALL",
  { "font-scale": 0.8, "vertical-align": "center" }
],
```

![big_image_center](https://github.com/user-attachments/assets/4dd7a13f-62cc-4d0d-9d39-c73d67bc7009)
![small_image_center](https://github.com/user-attachments/assets/cd54deb5-f27f-4bd3-b82e-6eb8efd5a474)

#### Top

```JavaScript
"text-field": [
  "format",
  ["image", "image"], 
  { "vertical-align": "top" },
  " ",
  {},
  "BIG yg",
  { "font-scale": 1.2, "vertical-align": "top" },
  " ",
  {},
  "SMALL",
  { "font-scale": 0.8, "vertical-align": "top" }
],
```

![big_image_top](https://github.com/user-attachments/assets/7fad5463-f61e-4f23-9410-7083302c21fc)
![small_image_top](https://github.com/user-attachments/assets/ecebf3d4-a83b-4c11-b392-c386dc7eb567)

#### Bottom (*default*, current behaviour)

```JavaScript
"text-field": [
  "format",
  ["image", "image"], 
  {},
  " ",
  {},
  "BIG yg",
  { "font-scale": 1.2 },
  " ",
  {},
  "SMALL",
  { "font-scale": 0.8 }
],
```

![big_image_baseline](https://github.com/user-attachments/assets/435f8267-9348-4d07-957a-48bd71ef5f4e)
[...]

PR #900, opened 2024-11-13
Add `vertical-align` in format expression

stanislawpuda-tomtom:
Related issue #832.

This PR introduces vertical align property to format expression. It enables to specify how each section should be positioned in relation to biggest element in line. There are three possible options:
- "bottom" _default_: text baseline or image bottom are in line - this is current behaviour.
vertical-align-bottom
- "center": image center or text center are in line.
vertical-align-center
- "top": image top and text top are in line
vertical-align-top

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


stanislawpuda-tomtom:
I would like to cover questions from the meeting on Nov 13th:

### Using HTML
I did check, and it is possible to use HTML to display vertically aligned labels. This approach works for MapLibre GL JS, however it does not cover multiple platforms. There are more advantages of using format expression over custom HTML layer:
- much easier to use
- HTML markers don't support collision detection, as far as I know, which makes the implementation even more difficult
- better performance
- defined once in the style spec for multiple platforms

Format expression already gives possibility to create rich labels on the map, easily, without having to implement own custom solution. Having the possibility to align elements vertically in my opinion is good supplement to format expression. In some cases necessary to make full benefit from it.

### Sample use case
Our team created a sample mock how the labels look like with and without vertical alignment.

Before:
before

After:
after

lseelenbinder:
Thanks for taking the time to work through these pieces @stanislawpuda-tomtom!

In my opinion, those advantages are sufficient to merit including in the spec—especially for collision detection and performance reasons.

stanislawpuda-tomtom:
I have changed `baseline` to `bottom`.
I also modified one test to check if `vertical-align` is recognised for image section.

src/reference/v8.json
HarelM, line 1: Missing sdk support table I believe, are you planning on creating a PR for this in the native repo?

HarelM:
Are you sure about the places where you added the code?
It seems that equivalent format related code does not look the same.
I know this is an enum while the others are string, or number, but I find the code a bit weird...

stanislawpuda-tomtom:
I've added `vertical-align` to SDK support table.

Screenshot 2024-11-28 at 11 57 56

For the code, let me try to explain:

```JavaScript
let verticalAlign = null;
if (arg['vertical-align']) { // covers case, when `vertical-align` is defined with plain string, then we can check if matches enum
    if (typeof arg['vertical-align'] === 'string' && !VERTICAL_ALIGN_OPTIONS.includes(arg['vertical-align'] as VerticalAlign)) {
        return context.error(`\`vertical-align\` must be one of: 'bottom', 'center', 'top' but found '${arg['vertical-align']}' instead.`) as null;
    }

    verticalAlign = context.parse(arg['vertical-align'], 1, StringType); // in this case it returns expression, so we can't check if it matches enum, only if it is string
    if (!verticalAlign) return null;
}
```

src/reference/v8.json
stanislawpuda-tomtom: Unfortunately we won't be able to make PR to native :(.

For the table, it was added.

src/reference/v8.json
```diff
@@ -3226,10 +3226,10 @@
...
+        "doc": "Returns a `formatted` string for displaying mixed-format text in the `text-field` property. The input may contain a string literal or expression, including an [`'image'`](#image) expression. Strings may be followed by a style override object that supports the following properties:\n\n- `\"text-font\"`: Overrides the font stack specified by the root layout property.\n\n- `\"text-color\"`: Overrides the color specified by the root paint property.\n\n- `\"font-scale\"`: Applies a scaling factor on `text-size` as specified by the root layout property.\n\n- `\"vertical-align\"`: Align vertically text section or image in relation to the row it belongs to. Possible values are: \n\t- `\"bottom\"` *default*: text baseline or image bottom are in line.\n\t- `\"center\"`: image center or text center are in line.\n\t- `\"top\"`: image top and text top are in line.\n\n - [Change the case of labels](https://maplibre.org/maplibre-gl-js/docs/examples/change-case-of-labels/)\n\n - [Display and style rich text labels](https://maplibre.org/maplibre-gl-js/docs/examples/display-and-style-rich-text-labels/)",
```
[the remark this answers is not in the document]
stanislawpuda-tomtom: Fixed.

src/reference/v8.json
stanislawpuda-tomtom: We did cover MapLibre Style Spec, MapLibre GL JS and defined the design for the feature + use cases etc. 

My assumption, and later what I understand from the meeting was that it won't be a blocker for already made implementation. If (when) there will be an interest on MapLibre Native side, someone who will be interested in using the feature could cover the implantation.

stanislawpuda-tomtom:
@HarelM @lseelenbinder 
Is it possible to proceed with this PR? Is there something I can do, clarify, or is there any blocker?

src/expression/definitions/format.ts
```diff
@@ -69,10 +70,21 @@ export class FormatExpression implements Expression {
                     if (!textColor) return null;
                 }
 
+                let verticalAlign = null;
+                if (arg['vertical-align']) {
+                    if (typeof arg['vertical-align'] === 'string' && !VERTICAL_ALIGN_OPTIONS.includes(arg['vertical-align'] as VerticalAlign)) {
```
HarelM, line 75: Can this if be moved to the parse code to keep this as clean as the other if statements?

src/expression/definitions/format.ts
```diff
@@ -93,15 +105,23 @@ export class FormatExpression implements Expression {
...
             const evaluatedContent = section.content.evaluate(ctx);
             if (typeOf(evaluatedContent) === ResolvedImageType) {
-                return new FormattedSection('', evaluatedContent, null, null, null);
+                return new FormattedSection(
+                    '',
+                    evaluatedContent,
+                    null,
+                    null,
+                    null,
+                    section.verticalAlign ? section.verticalAlign.evaluate(ctx) : null
```
HarelM, line 114: How is this different than other properties? Why not passing null here?

src/expression/types/formatted.ts
```diff
@@ -1,19 +1,24 @@
 import type {Color} from '../../expression/types/color';
 import type {ResolvedImage} from '../types/resolved_image';
 
+export const VERTICAL_ALIGN_OPTIONS = ['bottom', 'center', 'top'] as const;
```
HarelM, line 4: Shouldn't this be defined in the v8 spec file and not here?

HarelM:
I've added details on my concerns regarding the current code. Let me know if this is clearer now.

stanislawpuda-tomtom:
@HarelM It is clear - I'll cover them. Thank you!

src/expression/definitions/format.ts
stanislawpuda-tomtom: I did try to move it to parse - however then it shouldn't be placed in parse fn itself, but I found that it is better to add `enum` validation, like for array in other places.

However, to make it possible there are multiple changes that has to be made. Moreover the validation for enum is a bit different, because other validations are based on types, and enum validation requires access to value.

You can see how it looks here: https://github.com/tomtom-forks/maplibre-style-spec/pull/2

After making this try I actually think that having it in `format.ts`, as it is now, is better and much simpler. Also in similar fashion the validation is implemented for `interpolate` expression ([link to the file](https://github.com/maplibre/maplibre-style-spec/blob/93a8cd30b84b815ea57d49172959038a57a86759/src/expression/definitions/interpolate.ts#L71)).

src/expression/definitions/format.ts
stanislawpuda-tomtom: This is the instantiation of `FormattedSection` for image. All other properties are null, because they only applies to text (`font` etc.), but `vertical-align` can be applied to both image and text section.

src/expression/types/formatted.ts
stanislawpuda-tomtom: I don't see a way how to do it - there is no other expression where this kind of definition lives in `v8.json` (or I can't find it). For `interpolate` expression type definition lives in the code (https://github.com/maplibre/maplibre-style-spec/blob/93a8cd30b84b815ea57d49172959038a57a86759/src/expression/definitions/interpolate.ts#L17).

stanislawpuda-tomtom:
@HarelM 

I answered your comments.

HarelM:
Thanks, I'll look into it in the coming days.

src/expression/definitions/format.ts
HarelM: Yeah, the other PR is too complicated, and other places are doing the validation in the relevant class, much like resolvedImage is being done here.

HarelM (approved):

HarelM:
@louwers can you please review the v8 changes to make sure this is acceptable by you as well?

src/reference/v8.json
```diff
@@ -3226,10 +3226,10 @@
...
+        "doc": "Returns a `formatted` string for displaying mixed-format text in the `text-field` property. The input may contain a string literal or expression, including an [`'image'`](#image) expression. Strings may be followed by a style override object that supports the following properties:\n\n- `\"text-font\"`: Overrides the font stack specified by the root layout property.\n\n- `\"text-color\"`: Overrides the color specified by the root paint property.\n\n- `\"font-scale\"`: Applies a scaling factor on `text-size` as specified by the root layout property.\n\n- `\"vertical-align\"`: Aligns vertically text section or image in relation to the row it belongs to. Possible values are: \n\t- `\"bottom\"` *default*: text baseline or image bottom are in line.\n\t- `\"center\"`: image center or text center are in line.\n\t- `\"top\"`: image top and text top are in line.\n\n - [Change the case of labels](https://maplibre.org/maplibre-gl-js/docs/examples/change-case-of-labels/)\n\n - [Display and style rich text labels](https://maplibre.org/maplibre-gl-js/docs/examples/display-and-style-rich-text-labels/)",
```
louwers, line 3229: I think the wording is unclear here for someone not familiar with the design proposal.

> image top and text top are in line.

What does image "image top" and "text top" mean here? Better wording would be _Align this section of the formatted string so that the top of this section is in line with the top of the other sections._ And similarly for the other descriptions.

src/reference/v8.json
```diff
@@ -3226,10 +3226,10 @@
...
         "example": {
```
louwers, line 3230: The example should be updated to include a `vertical-align` section.

Ideally we would include the images for the design proposal.

If that is difficult for some reason, please update the design proposal with the latest modifications (i.e. use `bottom` instead of `baseline` and link to it so people have some additional context and visuals.

louwers (commented):
Left two comments. LGTM otherwise.

src/reference/v8.json
stanislawpuda-tomtom: Good point, done. I simplified the text you've proposed a bit:
Screenshot 2024-12-23 at 07 01 35

src/reference/v8.json
stanislawpuda-tomtom: I have updated the example:
Screenshot 2024-12-23 at 07 23 35

I also referenced design proposal. It is updated to use `bottom` instead of `baseline`.
Screenshot 2024-12-23 at 07 23 27

For the images - technically I can do it (it works) however there are no images in Expression documentation at all. I'm not sure what is a pattern here. Also I think it is good that expressions documentation is concise, and I'm not sure if adding images won't change that. I think it will be the most valuable to add or update example after maplibre-gl-js implementation. What do you think?

stanislawpuda-tomtom:
@louwers 

Answered. Thank you for your comments.

zbigniewmatysek-tomtom:
@HarelM @louwers Could you take the last look? Once this is merged, we will be able to continue the work on the implementation PR ;)

HarelM:
@louwers need to approve this as he had some code review comments, I have approved it.

louwers (approved):
Thanks for the changes.

LGTM, but I still think some images would do the docs good. :)

stanislawpuda-tomtom:
@louwers @HarelM 
For the images we can do something like this:
Screenshot 2025-01-13 at 11 27 33
Let me know if this is fine for you, I'll make a commit.

Btw I'm not sure where I should put images. I can place them in `/docs/assets`. Those images which are in docs already one is in some fork branch, and other hosted under maplibre.org domain.

louwers:
Looks great!

I personally prefer using GitHub for image hosting. We can replace the images with examples from actual maps when this is implemented.

stanislawpuda-tomtom:
Done!

louwers:
@HarelM If you think it's better with the images, could you merge?

PR #901, opened 2024-11-13
Remove interpolate-projection

HarelM:
## Launch Checklist
@birkskyum I took a stab at using the interpolate for projection instead of introducing a new interpolation expression.

I'm still looking into understanding how color names are being parsed, but in general, I think this looks like a good direction.
Let me know what you think.

### DIFF
--- a/test/integration/expression/tests/zoom/interpolate-projection/test.json
+++ b/test/integration/expression/tests/zoom/interpolate-projection/test.json
@@ -1,6 +1,16 @@
 {
+  "propertySpec": {
+    "type": "projection",
+    "property-type": "data-constant",
+    "expression": {
+      "interpolated": true,
+      "parameters": [
+        "zoom"
+      ]
+    }
+  },
   "expression": [
-    "interpolate-projection",
+    "interpolate",
     [
       "linear"
     ],
@@ -8,9 +18,9 @@
       "zoom"
     ],
     0,
-    "vertical-perspective",
+    ["literal", ["vertical-perspective", "vertical-perspective", 1]],
     10,
-    "mercator"
+    ["literal", ["mercator", "mercator", 1]]
   ],
   "inputs": [
     [
--- a/test/integration/expression/tests/zoom/interpolate-projection/test.json
+++ b/test/integration/expression/tests/zoom/interpolate-projection/test.json
@@ -18,9 +18,9 @@
       "zoom"
     ],
     0,
-    ["literal", ["vertical-perspective", "vertical-perspective", 1]],
+    "vertical-perspective",
     10,
-    ["literal", ["mercator", "mercator", 1]]
+    "mercator"
   ],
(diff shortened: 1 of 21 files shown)


test/integration/expression/tests/zoom/interpolate-projection/test.json
```diff
@@ -1,16 +1,26 @@
...
+    "interpolate",
     [
       "linear"
     ],
     [
       "zoom"
     ],
     0,
-    "vertical-perspective",
+    ["literal", ["vertical-perspective", "vertical-perspective", 1]],
```
HarelM, line 21: This is how I made the test work - calling literal on the project array allows converting the array to a `Projection`.
I'm not sure if this is great, would love to see what you think.

HarelM: Actually, it also works with regular strings when the interpreter knows which type it should output (projection in this case).

HarelM:
Edit: I think it is working as expected now according to the tests.

test/integration/expression/tests/zoom/interpolate-projection/test.json
birkskyum: regular strings?

I'd like to be able to define

    10,
    "vertical-perspective"
    12, 
    "mercator"
    
Is that possible?

HarelM: Yes, yes, other tests are doing exactly that.

HarelM: I don't think there's any validation on the strings right now though... But I'm sure it won't be a big issue to add validation...

birkskyum:
I like it a lot! Nice that it's going away the `interpolate-projection`. 

If we in any way can find a way to have just [step, [zoom], number, string, number, string] I think it would be worthwhile.

HarelM:
I can merge this to the other branch and you could test this. I think it should work, shouldn't it?

HarelM:
`[number, string, number, string]` is not an expression that is supported and I don't want to invent more expressions. So I think this is a good middle ground.

test/integration/expression/tests/zoom/interpolate-projection/test.json
birkskyum: there is no validation since I took out the .presets / .projections - it's for later.

What's special about this test, that makes it need the ["mercator", "mercator", 1] when the other tests doesn't?

HarelM: Nothing, I just wanted to see that `literal` supports `projection` casting.

HarelM: To be honest, I started with that to see that this would work and afterward saw that regular strings also works.

PR #902, opened 2024-11-13
Tweak - Use strings instead of strings parsed to projection arrays

birkskyum:
@HarelM 

There's still no `interpolate-projection`, but this small extension to your work makes the:

```
  "interpolate",
      [
        "linear"
      ],
      [
        "zoom"
      ],
      8,
      ["vertical-perspective","vertical-perspective",1],
      10,
      ["mercator","mercator",1]
```

Into this, which I find is a lot cleaner:

```
  "interpolate",
      [
        "linear"
      ],
      [
        "zoom"
      ],
      8,
      "vertical-perspective",
      10,
      "mercator"
```

For some reason, I just can't see why it's better for the API to use `["vertical-perspective","vertical-perspective",1]` to describe the zoom stops, when the last element is ignored, and the other duplicated. 

Just to double-check, do you prefer the arrays, even though they hold no additional information than a primitive string?

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/test/integration/expression/tests/zoom/interpolate-projection/test.json
+++ b/test/integration/expression/tests/zoom/interpolate-projection/test.json
@@ -18,9 +18,9 @@
       "zoom"
     ],
     0,
-    ["literal", ["vertical-perspective", "vertical-perspective", 1]],
+    "vertical-perspective",
     10,
-    ["literal", ["mercator", "mercator", 1]]
+    "mercator"
   ],
   "inputs": [
     [
--- a/src/expression/definitions/coercion.ts
+++ b/src/expression/definitions/coercion.ts
@@ -124,7 +124,7 @@ class Coercion implements Expression {
             case 'resolvedImage':
                 return ResolvedImage.fromString(valueToString(this.args[0].evaluate(ctx)));
             case 'projection':
-                return Projection.parse(this.args[0].evaluate(ctx));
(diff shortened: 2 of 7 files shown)


test/integration/expression/tests/zoom/interpolate-projection/test.json
```diff
@@ -18,9 +18,9 @@
       "zoom"
     ],
     0,
-    ["literal", ["vertical-perspective", "vertical-perspective", 1]],
+    "vertical-perspective",
```
HarelM, line 21: Does this mean literal is no longer supported?

birkskyum: Yes, it makes very little sense to me why we want to support literals, unless we want to allow transitions from one intermediate:

 `["vertical-perspective", "mercator", 0.2]` 
 
 to another
 
  `["equal-earth", "utm23", 0.5]` 
  
  which I think is _too_ flexible

HarelM:
I don't prefer the array, I just think it makes sense to support them, isn't it?

test/integration/expression/tests/zoom/interpolate-projection/test.json
HarelM: Fair enough.

PR #903, opened 2024-11-15
Move types, clean imports, move interpolate method

HarelM:
## Launch Checklist

This PR is mainly a refactoring of exiting code.
There were some classes that were placed in utils instead of types folder.
I moved the code around to fix that.
Also removed the deprecated interpolation factory.
I moved the imports to be at the begging of the file where I saw that this was not the case.


HarelM:
@birkskyum I believe you are familiar with most of the code changes here.
Let me know if something here is not clear.

src/util/interpolate-primitives.test.ts
```diff
@@ -0,0 +1,20 @@
+import {interpolateArray, interpolateNumber} from './interpolate-primitives';
+
+describe('interpolate', () => {
+
+    test('interpolate number', () => {
+        expect(interpolateNumber(-5, 5, 0.00)).toBe(-5.0);
+        expect(interpolateNumber(-5, 5, 0.25)).toBe(-2.5);
```
birkskyum, line 7: I this because we can't make a Number.interpolate()?

src/util/interpolate-primitives.ts
```diff
@@ -0,0 +1,9 @@
+export function interpolateNumber(from: number, to: number, t: number): number {
```
birkskyum, line 1: https://github.com/maplibre/maplibre-gl-js/blob/3d318e80220d083a960051e8601b107ce3f4ee77/src/util/util.ts#L210-L218

birkskyum: Will try to bench the two lerps against each other at some point - for the fun of it :)

birkskyum (approved):
Thanks, this helped

PR #904, opened 2024-11-16
Named exports

birkskyum:
I left the config files rollup/jest/eslint, but everything else is named exports. I've sometimes just mapped the name, so it's just the import line that changes, without affecting the re-exports, or other logic.

Since this is strictly only an internal change, I don't know if a changelog bullet is needed. 
## Launch Checklist

 - [X} Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/expression/types/color.ts
```diff
@@ -56,10 +56,10 @@ class Color {
...
     }
 
-    static black: Color;
-    static white: Color;
-    static transparent: Color;
-    static red: Color;
+    static black = new Color(0, 0, 0, 1);
+    static white = new Color(1, 1, 1, 1);
+    static transparent = new Color(0, 0, 0, 0);
+    static red = new Color(1, 0, 0, 1);
```
birkskyum, line 62: @HarelM , this is the only logic that changed - I didn't understand why the static properties were added under

bin/gl-style-validate.ts
```diff
@@ -2,7 +2,7 @@
 
 import minimist from 'minimist';
 import rw from 'rw';
-import validate from '../src/validate_style';
+import {validateStyle as validate} from '../src/validate_style';
```
HarelM, line 5: Can these renames be removed as well?

src/expression/types/color.ts
HarelM: This is probably from the time before typescript, where this was the way to achieve it using javascript.

HarelM: Looks better now, thanks!

bin/gl-style-validate.ts
birkskyum: Yes, the src/index.ts I imagine will be breaking if I remove, but I can look into the rest

src/util/extend.ts
```diff
@@ -1,4 +1,4 @@
-export default function extendBy(output: any, ...inputs: Array<any>) {
+export function extendBy(output: any, ...inputs: Array<any>) {
```
HarelM, line 1: I wonder if this can be removed now and we can use some typescript functionality

HarelM:
I think it's a big enough change to add to the changelog, even if it's only refactoring. 
I would also look into removing "import * from..." to finish up the cleaning.
Looks great, thanks!!

src/util/extend.ts
HarelM: Keep it for now then, it's out of scope for this PR...

HarelM (approved):

birkskyum:
The only `import * as` that are left are in test files, looping over exports or used in spyOn

Screenshot 2024-11-16 at 20 29 40

PR #905, opened 2024-11-17
Regression - Fix export of `latest`

birkskyum:
it's necessary to also have a default export on latest, otherwise it's not in index.d.ts

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 import v8Spec from './reference/v8.json' with {type: 'json'};
 const v8 = v8Spec as any;
-import {latest} from './reference/latest';
+import latest from './reference/latest';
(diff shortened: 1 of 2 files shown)


HarelM (approved):
Does this needs a test to cover this regression?

birkskyum:
@HarelM I actually did make a test for this, in he build suite.

In `style-spec.test.ts` I added

```ts
import {latest as latestInBundle} from '../../dist/index';

test('"latest" entry point should be defined', async () => {        
    expect(latestInBundle).toBeDefined();
});
```

The problem here is that it's testing our d.ts file, which uses standard ESM import/export, and Jest doesn't support that, so I get:

"unexpected token 'export'"

Screenshot 2024-11-17 at 11 54 51

PR #906, opened 2024-11-17
Serialize projectionDefinition to arrays

birkskyum:
This is how it would look to serialize to arrays like color does.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/expression/types/projection_definition.ts
+++ b/src/expression/types/projection_definition.ts
@@ -7,10 +7,11 @@ export class ProjectionDefinition {
         this.from = from;
         this.to = to;
         this.transition = transition;
+        
     }
 
     static interpolate(from: string, to: string, t: number) {
-        return new ProjectionDefinition(from, to, t);
+        return new ProjectionDefinition(from, to, t).toJSON();
     }
 
     static parse(input?: any): ProjectionDefinition {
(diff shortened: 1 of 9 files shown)


src/expression/types/projection_definition.ts
```diff
@@ -7,10 +7,11 @@ export class ProjectionDefinition {
         this.from = from;
         this.to = to;
         this.transition = transition;
+        
     }
 
     static interpolate(from: string, to: string, t: number) {
-        return new ProjectionDefinition(from, to, t);
+        return new ProjectionDefinition(from, to, t).toJSON();
```
HarelM, line 14: I think interpolate should return the same type of object and not a string...

HarelM:
I'm failing to see the motivation behind this PR to be honest...

src/expression/types/projection_definition.ts
birkskyum: Okay,

Just to clarify, this does return an array, not a string.

birkskyum:
@HarelM , thanks for being honest with it, I'm fine with either.

> I do see that color does get serialized to array somehow (saw it in test/integration/expression/tests/interpolate-hcl/linear/test.json test), but I couldn't figure out where it happens.

My main motivation here was to ensure that we could freely decide on API design, if we serialize to arrays or class objects, instead of having the choice made for us due to technical limitations.

PR #908, opened 2024-11-17
Add build test of `latest`

### LINKED ISSUE #907: Add build test for checking export of `latest`
Related to:
- #905 

We should have a test in style-spec.test.ts like below to ensure the `latest` object is exported:

import {latest as latestInBundle} from '../../dist/index';

test('"latest" entry point should be defined', async () => {        
    expect(latestInBundle).toBeDefined();
});

We have a Jest 29 problem here, because it's a test of the index.d.ts file, which uses standard ESM import/export, and Jest doesn't support that, so I get:

"unexpected token 'export'"

image

We have two options - we can swap for Vitest now, or wait till Jest 30 is out, which is [in progress](https://github.com/jestjs/jest/milestone/15), and try with that.

birkskyum:
Closes #907 

This requires jest 30 / vitest to work

### DIFF
--- a/test/build/style-spec.test.ts
+++ b/test/build/style-spec.test.ts
@@ -1,5 +1,6 @@
 import {readdir} from 'fs/promises';
 import {latest} from '../../src/reference/latest';
+import {latest as latestInBundle} from '../../dist/index';
 import fs from 'fs';
 
 const minBundle = fs.readFileSync('dist/index.mjs', 'utf8');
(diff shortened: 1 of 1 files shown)


HarelM:
@birkskyum is this still relevant or can I close this PR?

birkskyum:
I had forgotten about it - i guess now we have vitest, which has the ESM support needed, I'd like to see if i can make it work.

birkskyum:
because of vitest, we can now add this test

PR #951, opened 2024-12-22
Add `semiliteral` operator for expressions inside arrays and objects

### LINKED ISSUE #950: Design Proposal: Allow creating arrays with expressions
# Design Proposal: Allow creating arrays with expressions

## Motivation

A number of layer properties take arrays as their values. For example, see `text-offset` and `text-variable-anchor-offset`. The only way to produce an array in the existing expression system is to either get it from a feature with `get`, or create it with `literal`. Neither of these allow calculating the elements of the array with math expressions, though you can select between different literal arrays or properties with conditional `match` / `case` expressions.

For arrays of just numbers (let's call them vectors), one can work around this for certain use cases by interpolating between multiple vectors, and doing your math on the interpolation input. MapLibre Compose does this to simulate multiplying the magnitude of a `text-offset` (to convert between SP and EM units).

For arrays containing other types, not even that workaround is available (example: `text-variable-anchor-offset`). MapLibre Compose was unable to work around this.

Some use cases that could benefit from expressions in array elements:

- x and y components of an offset based on different feature properties (see this old "bug" report: https://github.com/maplibre/maplibre-style-spec/issues/545)
- calculating an offset with `sin` and `cos`, such as to offset in the direction of some angle
- scaling the magnitude of an offset, for example maplibre-compose scales properties provided in EM units in order to support Compose's TextUnit type which can be in either SP or EM units

## Proposed Change

I propose a new function in the style spec, `[]`, that takes any number of expression arguments and returns an array of all those arguments. The syntax would look like:

```json
{
  "text-offset": ["[]", ["get", "label_x"], ["get", "label_y"]]
}
```

This is in contrast to `literal`, which does _not_ evaluate its arguments as expressions.

## API Modifications

Just the new function `[]`. Also perhaps a similar function `{}` for objects.

No migration needed for the main proposal; this is new functionality.

Some alternatives in the "Rejected Alternatives" section below are backwards incompatible and would need a migration if they're selected.

## Rejected Alternatives

- **Update `literal` to evaluate its elements as expressions**: this would be backwards compatible in cases of nested arrays. The only place in the style spec that uses nested arrays is `text-variable-anchor-offset`, whose value looks like `["top", [1, 2], "bottom", [3, 4]]`.
- **Add an optional boolean parameter to `literal` to evaluate its elements as expressions**: I think this is confusing to the reader, as it's not immediately apparent what `true` or `false` mean when reading a style spec. Also, my impression (haven't read the code so perhaps wrong) is that `literal` is part of the expression parser, and not part of expression evaluation like other functions.
- **Just evaluate elements of any array that doesn't start with a string**: If you want the first element of such an array to be a literal string, you'd have to escape it with something like `concat`. Feels inconsistent.
-  **Double brackets syntax like `[[ ["get", "x"], ["get", "y"] ]]`**: Syntax is ambiguous in case of single element arrays where the first element is an array of strings (looks like a function call). Would need some sort of escape like the above alternative.
- **New function as proposed, but alternative names**: I considered various alternative names before landing on `[]`:
  - Looking for existing functions in the spec that take some input and return a new object, we have `collator` that creates a Collator, `format` that creates a Formatted , and `image` that creates a ResolvedImage. So, the natural convention might be `array` to create an Array, but that function name is already taken for a function to assert that the input is an array (similar to `string`, `number`, `boolean`, and `object`)
  - We have `to-color`, `to-string`, etc, which all take a single value and convert it to the target type (some with fallbacks). The name `to-array` is available, but the convention doesn’t quite match what we want.
  - `rgb` and `rgba` are closest in concept to what we want (evaluate some arguments as expressions, construct a result object from it) but of course that naming convention doesn't have an obvious analogue for arbitrary arrays.
  - So, I think a new naming convention is necessary, and `[]`/`{}` seems reasonable.

sargunv:
## Description

The purpose of this PR is to add support for producing arrays with elements calculated by expressions. It's a draft implementation of https://github.com/maplibre/maplibre-style-spec/issues/950.

I implemented it as a single `semiliteral` operator mirroring the `literal` syntax, like other lisps have quasiquote and quote. Happy to restructure this into `[]/{}` or some other name as discussed in that design proposal.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/expression/expression.ts
```diff
@@ -18,7 +18,7 @@ export interface Expression {
 
 export type ExpressionParser = (args: ReadonlyArray<unknown>, context: ParsingContext) => Expression;
 export type ExpressionRegistration = {
-    new (...args: any): Expression;
+    prototype: Expression;
```
sargunv, line 21: tweaked this because my `Semiliteral` class is `abstract` and therefore doesn't have a constructor

src/reference/v8.json
```diff
@@ -2847,6 +2847,24 @@
...
+          "syntax": {
+            "method": ["JSON object or array"],
+            "result": "array | object"
+          },
+          "value": ["semiliteral",[["get", "label_x"], ["get", "label_y"]]]
+        },
+        "group": "Types",
+        "sdk-support": {
+          "basic functionality": {
+            "js": "https://github.com/maplibre/maplibre-style-spec/issues/950",
```
louwers, line 2862: The idea is to link to issues in the MapLibre GL JS and MapLibre Native repo so that people can track progress on the implementation.

HarelM:
Yes, I think that was the last action item on this topic.

HarelM:
@sargunv any updates on this?

src/reference/v8.json
```diff
@@ -3160,6 +3160,24 @@
...
       },
+      "semiliteral": {
+        "doc": "Provides an array or object value, evaluating expressions in the elements of the array or values of the object.",
+        "example": {
+          "syntax": {
+            "method": ["JSON object or array"],
+            "result": "array | object"
+          },
+          "value": ["semiliteral",[["get", "label_x"], ["get", "label_y"]]]
+        },
```
CommanderStorm:
```suggestion
        "syntax": {
          "overloads": [
            {
              "parameters": ["json_object"],
              "output-type": "object"
            },
            {
              "parameters": ["json_array"],
              "output-type": "array"
            }
          ],
          "parameters": [
            {
              "name": "json_object",
              "type": "JSON object"
            },
            {
              "name": "json_array",
              "type": "JSON array"
            }
          ]
        },
        "example": ["semiliteral", [["get", "label_x"], ["get", "label_y"]]],
```

src/reference/v8.json
```diff
@@ -3160,6 +3160,24 @@
...
+        "group": "Types",
+        "sdk-support": {
+          "basic functionality": {
+            "js": "https://github.com/maplibre/maplibre-style-spec/issues/950",
+            "android": "https://github.com/maplibre/maplibre-style-spec/issues/950",
+            "ios": "https://github.com/maplibre/maplibre-style-spec/issues/950"
```
CommanderStorm:
```suggestion
            "js": "https://github.com/maplibre/maplibre-gl-js/issues/7933",
            "android": "https://github.com/maplibre/maplibre-native/issues/4401",
            "ios": "https://github.com/maplibre/maplibre-native/issues/4401"
```

louwers:
Thanks @CommanderStorm

src/expression/definitions/semiliteral.ts
```diff
@@ -0,0 +1,111 @@
...
+        this.type = type;
+    }
+
+    static parse(args: ReadonlyArray<unknown>, context: ParsingContext): Expression {
+        if (args.length !== 2)
+            return context.error(
+                `'semiliteral' expression requires exactly one argument, but found ${args.length - 1} instead.`
+            ) as null;
+
+        if (!isValue(args[1])) return context.error('invalid value') as null;
```
HarelM: I think it can be useful to point out what value is invalid...?

src/reference/v8.json
```diff
@@ -3160,6 +3160,40 @@
           }
         }
       },
+      "semiliteral": {
+        "doc": "Provides an array or object value, evaluating expressions in the elements of the array or values of the object.",
+        "syntax": {
+          "overloads": [
+            {
+              "parameters": ["json_object"],
```
HarelM: When is this relevant? MVT doesn't support it, and MLT still doesn't support it. Is this relevant to geojson?

HarelM:
Thanks @CommanderStorm for picking this up!
I'm not sure I fully follow the object input/output idea, can you elaborate on this use case?
Aslo, did you check that this really works with gl-js and actually solves the issues that were brought up there that this spec change solves (i.e. sanity check)?

Otherwise, this looks good.

src/reference/v8.json
CommanderStorm: Yes, `literal` also supports this.

src/expression/definitions/semiliteral.ts
CommanderStorm: See literal, that is how we do this.

https://github.com/maplibre/maplibre-style-spec/blob/faaa9fe3b4f543decacb4015a88c8ef125348490/src/expression/definitions/literal.ts#L23

CommanderStorm:
> can you elaborate on this use case

Consistency with `literal`. it is supported there, so this should support it too.

HarelM:
This is more "data driven" so passing a static json object is different than getting it from a feature. If this was the same as literal we wouldn't need this new style spec property. 
I prefer not to add it if there's currently no way to use it, we can always add it in the future...

src/expression/definitions/semiliteral.ts
CommanderStorm: the only way this can actually happen is a non-json value sneaks in, so a function or things..

CommanderStorm:
You can use it, see the third parameter to get or has or such, but I removed it if you insist.

HarelM (approved):

HarelM:
I'm not sure what should be done with the following comment:
- https://github.com/maplibre/maplibre-style-spec/pull/951#discussion_r1895082611

But from my point of view this can be merged and a new version can be released.
If you want to wait for the other PRs to be merged that fine too.

src/expression/expression.ts
```diff
@@ -21,7 +21,7 @@ export type ExpressionParser = (
     context: ParsingContext
 ) => Expression;
 export type ExpressionRegistration = {
-    new (...args: any): Expression;
+    prototype: Expression;
```
CommanderStorm:
```suggestion
    new (...args: any): Expression;
```

sargunv:
thanks for driving this @CommanderStorm and @HarelM ! Apologies for letting it languish for so long

louwers:
@sargunv Do you want to do the Native impl?

PR #1028, opened 2025-03-01
Fix `validate_object` crashing when object prototype keys used in style's objects

cxcorp:
From #1025

Fixes an unhandled error occurring when validated objects (e.g. sources) had keys that were found in `Object.prototype`. This caused certain truthyness checks to incorrectly pass, as e.g. `object["__proto__"]` always returns a truthy value (unless object has a null prototype or otherwise non-`Object` prototype).

This caused unhandled errors such as:

```
TypeError: validateElement is not a function
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at Object.validateSource [as source] (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_source.ts:32:22)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:96:42)
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:99:23)
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:99:23)
    at validateStyleMin (node_modules/@maplibre/maplibre-gl-style-spec/src/validate_style.min.ts:35:28)
    at src/fuzz.test.ts:175:11
```

```
TypeError: Cannot convert undefined or null to object
    at valueOf (<anonymous>)
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at Object.validateSource [as source] (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_source.ts:32:22)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:96:42)
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:99:23)
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
    at validate (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate.ts:99:23)
[...]

Issue #1025, opened 2025-02-24
Multiple ways to cause uncaught error in validator instead of returning array of validation errors


cxcorp:
I made a [rudimentary fuzzer](https://github.com/cxcorp/maplibregl-style-fuzzer) to fuzz the `validateStyleMin` function with intentionally out-of-spec styles, e.g. broken layers or sources, and discovered 10+ unique styles that cause the validator to throw an uncaught error instead of returning an array of validation errors.

Some issues stem from systematically using untrusted property keys to access e.g. lookup objects, some stem from not checking that e.g. the objects in an array are actually objects instead of `null`, some a bit more edge-casy (e.g. objects that override `toString` and thus crash when attempting to stringify the object for an error message).

**I would like to ask if `maplibre-style-spec` considers uncaught errors to be a problem or not.** If all consumers wrap the validation function in a `try-catch` and assume that sources can be trusted, there is no problem and there is no need for me to go produce more detailed bug reports or fixes. If they do not, and the libraries DO try to protect against malicious source JSONs, then I would be happy to go into further detail and try to provide some fixes in the form of PRs. Please let me know.

Example style (notice the additional key `__proto__.` in the `ne2_shaded` source):

```json
{
  "version": 8,
  "metadata": {
    "mapbox:autocomposite": false,
    "maputnik:renderer": "mbgljs",
    "openmaptiles:version": "3.x"
  },
  "sources": {
    "ne2_shaded": {
      "maxzoom": 6,
      "tileSize": 256,
      "tiles": [
        "https://tiles.openfreemap.org/natural_earth/ne2sr/{z}/{x}/{y}.png"
      ],
      "type": "raster",
      "__proto__.": 6
    },
    "openmaptiles": {
      "type": "vector",
      "url": "https://tiles.openfreemap.org/planet"
    }
  },
  "sprite": "https://tiles.openfreemap.org/sprites/ofm_f384/ofm",
  "glyphs": "https://tiles.openfreemap.org/fonts/{fontstack}/{range}.pbf",
  "layers": [
    {
      "id": "background",
      "type": "background",
      "paint": {
        "background-color": "#f8f4f0"
      }
    }
  ]
}
```

Causes the following error:

```
TypeError: validateElement is not a function
    at validateObject (node_modules/@maplibre/maplibre-gl-style-spec/src/validate/validate_object.ts:38:32)
[...]

PR #1036, opened 2025-03-04
Fix prototype keys causing crash in color parsing and invalid colors in color coercion expressions

cxcorp:
From #1025

This PR fixes two issues:
1. When parsing color strings (e.g. `text-color: "red"`), `parse_css_color.ts` attempted to look up named colors from an object literal by doing `namedColors[input]`. If the color string was `"__proto__"` or `"toString"` or some other key in the `Object` prototype, the value was truthy, and caused an unhandled error to be thrown because the value was destructured.
   https://github.com/maplibre/maplibre-style-spec/blob/91175b30b548f27527ec9d3ac09b73fed31cf699/src/expression/types/parse_css_color.ts#L40-L44
2. When a `Coercion` expression was used to coerce an expression's string value to a color (e.g. `to_color`), it used `EvaluationContext.parseColor()` which checked if the string had already been parsed and cached. The cache lookup was done via `this._parseColorCache[input]`, where `input` comes from the user controlled style. As a result, the input was never parsed as the value was truthy, and the value from the prototype was returned and became the expression's value.
   https://github.com/maplibre/maplibre-style-spec/blob/91175b30b548f27527ec9d3ac09b73fed31cf699/src/expression/evaluation_context.ts#L16
   https://github.com/maplibre/maplibre-style-spec/blob/91175b30b548f27527ec9d3ac09b73fed31cf699/src/expression/evaluation_context.ts#L48-L53

Fixing issue 1 revealed issue 2, as `EvaluationContext.parseColor()` used `Color.parse()` which used `parseCssColor()`. `Color.parse()` is also used in `function` expressions.

The fixes for the issues were:
1. For `parseCssColor()`, since we probably want the convenience of using an object literal to initialize it, I opted to use the previously introduced (#1028) `getOwn()` utility to ensure that only the literal's keys are tested.
2. For `EvaluationContext.parseColor()`, the ` _parseColorCache` was previously initialized to an empty object literal `{}`. Since there was no need for the ergonomics of using a literal, I replaced it with `Object.create(null)` which creates an object without a prototype (note: `{}` is syntactic sugar for `Object.create(Object.prototype)`).

The ideal solution for this would be to systematically use a `Map` if the key is user-controlled, but I did not find any uses of it from the codebase and assumed that the library targets ES5 compatibility.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/expression/evaluation_context.ts
+++ b/src/expression/evaluation_context.ts
@@ -13,16 +13,14 @@ export class EvaluationContext {
     availableImages: Array<string>;
     canonical: ICanonicalTileID;
 
-    _parseColorCache: {[_: string]: Color};
+    _parseColorCache: Map<string, Color>;
 
     constructor() {
         this.globals = null;
         this.feature = null;
         this.featureState = null;
         this.formattedSection = null;
-        // the cache keys are user controlled (from the source JSON), so
-        // avoid prototype pollution by creating a record with a null prototype
-        this._parseColorCache = Object.create(null) as {[_: string]: Color};
+        this._parseColorCache = new Map<string, Color>();
         this.availableImages = null;
         this.canonical = null;
     }
@@ -48,9 +46,10 @@ export class EvaluationContext {
     }
 
     parseColor(input: string): Color {
-        let cached = this._parseColorCache[input];
+        let cached = this._parseColorCache.get(input);
         if (!cached) {
-            cached = this._parseColorCache[input] = Color.parse(input);
+            cached = Color.parse(input);
+            this._parseColorCache.set(input, cached);
         }
         return cached;
     }
--- a/src/expression/evaluation_context.ts
+++ b/src/expression/evaluation_context.ts
@@ -20,7 +20,9 @@ export class EvaluationContext {
         this.feature = null;
         this.featureState = null;
         this.formattedSection = null;
-        this._parseColorCache = {};
+        // the cache keys are user controlled (from the source JSON), so
+        // avoid prototype pollution by creating a record with a null prototype
+        this._parseColorCache = Object.create(null) as {[_: string]: Color};
         this.availableImages = null;
         this.canonical = null;
     }
--- a/test/integration/style-spec/tests/bad-color.output.json
+++ b/test/integration/style-spec/tests/bad-color.output.json
@@ -10,5 +10,17 @@
   {
     "message": "layers[1].paint.fill-outline-color: color expected, array found",
     "line": 26
+  },
+  {
+    "message": "layers[2].paint.fill-color: color expected, \"__proto__\" found",
+    "line": 35
+  },
+  {
+    "message": "layers[2].paint.fill-outline-color.stops[0][1]: color expected, \"valueOf\" found",
+    "line": 39
+  },
+  {
+    "message": "layers[2].paint.fill-outline-color.stops[1][1]: color expected, \"__proto__\" found",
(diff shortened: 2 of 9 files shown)


src/expression/evaluation_context.ts
```diff
@@ -20,7 +20,9 @@ export class EvaluationContext {
         this.feature = null;
         this.featureState = null;
         this.formattedSection = null;
-        this._parseColorCache = {};
+        // the cache keys are user controlled (from the source JSON), so
+        // avoid prototype pollution by creating a record with a null prototype
+        this._parseColorCache = Object.create(null) as {[_: string]: Color};
```
cxcorp, line 25: The current TypeScript configs didn't seem to care whether this `any` was casted or not, but I added it so that it doesn't need to be fixed in the future if the TS configs are ever made stricter in the future.

test/integration/style-spec/tests/bad-color.output.json
```diff
@@ -10,5 +10,17 @@
   {
```
cxcorp, line 10: I noticed that there were both `*.output-api-supported.json` and `*.output.json` files in the tests, but was unable to find references to `output-api-supported` or `api-supported` in the codebase, other than a changelog entry saying `--mapbox-api-supported` had been removed from `gl-style-validation` in 14.0.0. Are the `.output-api-supported.json` files still used, or subject for removal in a later PR?

src/expression/evaluation_context.ts
HarelM: If you think a Map object would be a better fit here (looks like it, since we are talking about cache) please change the code accordingly.

test/integration/style-spec/tests/bad-color.output.json
HarelM: I can't find anything related to this in the history of the repo, I'm not sure they were ever used, and also from a very shallow test I made these looks duplicates of the output files, so I would recommend removing them.

test/integration/style-spec/tests/bad-color.output.json
```diff
@@ -10,5 +10,17 @@
...
     "message": "layers[1].paint.fill-outline-color: color expected, array found",
     "line": 26
+  },
+  {
+    "message": "layers[2].paint.fill-color: color expected, \"__proto__\" found",
```
HarelM, line 15: Should the message beyter describe the problem? Is there a word missing after the "found"?

HarelM: Ignore this, my mistake...

src/expression/evaluation_context.ts
cxcorp: Sure, I think it's the best choice here, but was hesitant to introduce it as the first use in this repo. I thought it might be a ES5 compatibility thing, but on further investigation the codebase is already using ES6 features like `Set` and `Number.isNaN` so it's probably just a matter of this being a long-lived codebase.

Changed to use a `Map` in c52614560a3954ab939200ee98d6e309451b902d.

test/integration/style-spec/tests/bad-color.output.json
cxcorp: Sure. Remove in this PR or shall I make a separate PR?

HarelM: A separate PR seems like the right approach.

Issue #886, opened 2024-11-04
Design Proposal: Global map state


hiddewie:
# Design Proposal: Global map state

## Motivation

Split off from https://github.com/maplibre/maplibre-style-spec/issues/516#issuecomment-2453478541, and originally posted / requested in https://github.com/maplibre/maplibre-gl-js/issues/4964.

The goal is to store some global map state on the `Map` object, which can be used in the style. This makes it easier to let the user configure the map for certain preferences, which can then dynamically be used in the style to modify the appearance of the map.

My concrete use case is supporting a map *theme*, where the user selects a theme and this impacts the look and feel of the web application, in particular the features on the map.

## Proposed Change

Introduce a function `map-state` which takes a single string argument which is the key.

The return value is the value in the global map state, if present, otherwise `null`.

Examples:
```js
paint: {
  'text-color': ['map-state', 'color'],
}
```

Handling unset key in the map state
```js
paint: {
  'text-color': ['coalesce', ['map-state', 'color'], 'red']
}
```

The map state will function as a key-value dictionary, so the values in the map state contains string keys and string, numeric or boolean values. We could choose to support e.g. arrays as values as well.

The `map-state` function should have the same semantics as `feature-state` which uses data set on a feature.

## API Modifications

The expression `map-state` will be supported.

## Migration Plan and Compatibility

No migration needed, this is new functionality.

The feature should not be complex to implement in the specification and client libraries because:
- the function `feature-state` already exists, and this function is similar although it takes the state from a different place
- the map state is a new feature, so this will not conflict with existing features.

Probably the biggest complexity is managing the global map state changing its value, and recomputing the style for the impacted features on the map.

## Rejected Alternatives

- https://github.com/maplibre/maplibre-gl-js/issues/4964#issuecomment-2453139681: Changing the style dynamically during loading of the map: this makes the map style highly coupled with the client using the style. In particular, the client application has to walk every style expression, and match the expression against the replacements with the configured global values from the user configuration.
- https://github.com/maplibre/maplibre-gl-js/issues/4964#issuecomment-2453480099: using `let` expressions and dynamically transforming the style. This still couples the client with the map style, and still requires walking the expressions in the map style to dynamically replace the `let` values with the configured values.
- https://github.com/maplibre/maplibre-style-spec/issues/516: Design proposal to add arbitrary functions to the map. That proposal would replace this proposal, but the scope is much bigger and more complex to implement than this feature.
[...]

PR #1044, opened 2025-03-11
global-state expression

zbigniewmatysek-tomtom:
This PR introduces:
* new expression `global-state` that retrieves a property value from the new root property `state`.
* new root property `state` , where the default value can be set. This eliminates need to put `coalesce` in all places where `global-state` is used.

The expression is similar to `get` and such be interpreted by renderers as such:
* when referenced property is not set, it must return `null`
* referenced property values are treated as primitives, so there is no need to wrap `global-state` in `literal` if it returns an array.

Renderer implementations can expose methods such as setGlobalStateProperty(name, value) for changing the state properties at runtime.

Use cases enabled or simplified by this extension:
* Setting map language (without a plugin or SDK wrapper code)
* POI categories filtering (without using layer.setFilter() that relies on the filter content to work)
* Single style with multiple variants (light/dark mode, browsing/navigation) instead of multiple styles.

Related issues:
https://github.com/maplibre/maplibre-style-spec/issues/886
https://github.com/maplibre/maplibre-gl-js/issues/4964

## Launch Checklist

 - [x] Link to related issues.


src/expression/definitions/global_state.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        if (!globalState || Object.keys(globalState).length === 0) return null;
+
+        const key = this.input.evaluate(ctx);
+
+        if (!isValidNativeType(key, ['string'])) {
+            throw new RuntimeError(`Global state property must be string, but found ${valueToString(key)} instead,`);
+        }
+
+        return globalState[key];
```
HarelM, line 43: I've seen some guards to protect this kind of invocation in recent PRs, can you check if this needs to be applied here (getOwn)?

src/reference/v8.json
```diff
@@ -3286,6 +3286,24 @@
...
+        "example": {
+          "syntax": {
+            "method": ["string"],
+            "result": "value"
+          },
+          "value": ["global-state", "someProperty"]
+        },
+        "sdk-support": {
+          "basic functionality": {
+            "js": "TBD",
```
HarelM, line 3301: Please add relevant issues link here, see other places related to web-native parity.

src/expression/definitions/global_state.ts
zbigniewmatysek-tomtom: Changed to getOwn.

build/generate-docs.ts
```diff
@@ -517,6 +518,11 @@ function createExpressionsContent() {
     fs.writeFileSync(`${BASE_PATH}/expressions.md`, content);
 }
 
+function createSchemaContent() {
+    const content = fs.readFileSync('build/schema.md', 'utf-8');
```
HarelM, line 522: Is this really needed? We have other .md files in the docs folder...

package.json
```diff
@@ -32,7 +32,7 @@
     "compile": "tsc",
     "lint": "eslint",
     "typecheck": "tsc --noEmit",
-    "prepare": "npm run generate-style-spec"
+    "prepare": "npm run generate-style-spec && npm run generate-typings && npm run build"
```
HarelM, line 35: Don't forget to remove this when this is ready for review.

src/reference/v8.json
```diff
@@ -97,6 +97,32 @@
         }
       }
     },
+    "state": {
+      "type": "state",
+      "default": {},
+      "doc": "The initial state of the style. This object where keys are property names and value a [`schema`](https://maplibre.org/maplibre-style-spec/schema) definition. State property value can be retrieved with the [`global-state`](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression.\n\n State is required when `global-state` is used in the style",
```
HarelM, line 103: I'm not sure this is well explained. is "schema definition" a new type? should it be in the types.md file instead of its own file?
Also this is not written well I believe: "...and value a schema definition."

build/schema.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+### number
+
+The `number` type is used for any numeric type, either integers or floating point numbers.
+
+```json
+{  
+    "type": "number",
+    "minimum": 0, // optional
```
HarelM, line 28: Are we sure this is what we want to start supporting? Schema parsing and validating is complicated, we are already doing it for the style json, I wouldn't want to introduce a different "language", if possible...

build/schema.md
```diff
@@ -0,0 +1,79 @@
+# Schema
+
+The value for `state` properties. Defines the default state property value, and its type.
+
+The schema definition follows a subset of [JSON Schema](https://json-schema.org/) but is more strict in terms of what properties are required.
```
HarelM, line 5: See me other comment about using a different "language".

build/generate-style-spec.ts
```diff
@@ -315,6 +317,10 @@ export type DataDrivenPropertyValueSpecification<T> =
     | CompositeFunctionSpecification<T>
     | ExpressionSpecification;
 
+
+export type StateSpecification = Record<string, SchemaSpecification>;
+export type SchemaSpecification = Record<string, unknown>;
```
HarelM, line 322: I think this can be improved if there's a schema definition for type, default etc, can't it?

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+        case 'array': {
+            return validateArray(value, key);
+        }
+        default: {
+            if (value.type !== undefined) {
+                return [
+                    new ValidationError(
+                        'type',
+                        value.type,
+                        'expected string, number or boolean'
```
HarelM, line 53: array too?

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+                    ),
+                ];
+            }
+        }
+    }
+
+    return validateEnum(value, key);
+}
+
+function validateString(schema: Record<string, unknown>, key: string) {
```
HarelM, line 63: Don't we have these kinds of validations already as part of the style spec validation code...?

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+        });
+
+        errors.push(...itemErrors);
+    }
+
+    return errors;
+}
+
+function validateBoolean(schema: Record<string, unknown>, key: string) {
+    if (typeof unbundle(schema.default) != 'boolean') {
```
stanislawpuda-tomtom, line 217:
```suggestion
    if (typeof unbundle(schema.default) !== 'boolean') {
```

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+import {getType} from '../util/get_type';
+import {deepUnbundle, unbundle} from '../util/unbundle_jsonlint';
+import {isObjectLiteral} from '../util/is_object_literal';
+
+interface ValidateSchemaOptions {
+    key: string;
+    value: unknown;
+}
+
+export function validateSchema({
```
stanislawpuda-tomtom, line 11: I would reconsider naming there. Schema is very broad, and when I see this function, I really have no idea what it validates, especially in package like style spec. Maybe `GlobalStateDefinition`? Or sth like that?

stanislawpuda-tomtom: I would expect unit tests whole content of this file.

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+function validateString(schema: Record<string, unknown>, key: string) {
+    if (typeof unbundle(schema.default) !== 'string') {
+        return [
+            new ValidationError(`${key}.default`, schema.default, 'string expected'),
+        ];
+    }
+
+    return [];
+}
+function validateNumber(schema: Record<string, unknown>, key: string) {
```
stanislawpuda-tomtom, line 72: Can't schema property have a more specific type?

test/integration/style-spec/tests/global-state-missing-definition.input.json.json
```diff
@@ -0,0 +1,21 @@
+{
```
stanislawpuda-tomtom, line 1: This file has double `.json` file name extension.

stanislawpuda-tomtom: Also for the test naming, we should keep `state` or `global-state`.

src/reference/v8.json
```diff
@@ -97,6 +97,32 @@
...
+          },
+          "default": ["CCS", "CHAdeMO", "Type2"]
+        },
+        "minPreferredChargingSpeed": {
+          "type": "number",
+          "minimum": 0,
+          "default": 50
+        }
+      },
+      "sdk-support": {
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: It is visible in Root:

image

src/validate/validate_schema.ts
zbigniewmatysek-tomtom: At the moment of validation, it is unknown.

src/validate/validate_schema.ts
zbigniewmatysek-tomtom: Fixed

src/validate/validate_schema.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (!isObjectLiteral(value)) {
+        return [
+            new ValidationError(
+                key,
+                value,
+                `object expected, ${getType(value)} found`
+            ),
+        ];
+    }
+    if (value.type === undefined && value.enum === undefined) {
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: I aligned it with JSON schema where type is not required for enum.

src/validate/validate_global_state_expression.ts
```diff
@@ -0,0 +1,26 @@
...
+    return [];
+}
+
+export function findGlobalStateExpression(expression: Expression): GlobalState {
+    if (expression instanceof GlobalState) {
+        return expression;
+    }
+
+    let result = null;
+    expression.eachChild((child) => {
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: Fixed

src/expression/expression.test.ts
```diff
@@ -666,6 +665,34 @@ describe('projection expression', () => {
...
+        expect(response.result).toBe('error');
+        expect(response.value[0]).toBeInstanceOf(ExpressionParsingError);
+        expect((response.value[0] as ExpressionParsingError).message).toBe('Expected 1 argument, but found 2 instead.');
+    });
+    test('evaluates a global state property', () => {
+        const response = createExpression(['global-state', 'foo']);
+        if (response.result === 'success') {
+            expect(response.value.evaluate({globalState: {foo: 'bar'}, zoom: 0}, {} as Feature)).toBe('bar');
+        } else {
+            throw new Error('Failed to parse GlobalState expression');
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: This is by convention, this condition tests the test definition itself - take a look at other tests in this file.

src/reference/v8.json
```diff
@@ -3286,6 +3312,24 @@
           }
         }
       },
+      "global-state": {
+        "doc": "Retrieves a property value from the [state](https://maplibre.org/maplibre-style-spec/root/#state) property. Default values must be defined in the `state` property.\n\n - [Setting global state](https://maplibre.org/maplibre-gl-js/docs/examples/global-state/)",
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: Ad. 1) I removed the default mention.
Ad 2) The description is factual, it is taken from another place in the style. There will be a renderer method to set the style value. Where the actual value is stored (being that in the style or another object in the renderer) is an implementation detail, as far as the user is concerned, it is set in the style and taken from the style.

src/validate/validate_schema.ts
zbigniewmatysek-tomtom: Fixed

build/generate-style-spec.ts
zbigniewmatysek-tomtom: Done

src/reference/v8.json
```diff
@@ -97,6 +97,32 @@
         }
       }
     },
+    "state": {
+      "type": "state",
+      "default": {},
+      "doc": "The initial state of the style. This object where keys are property names and value is a [`schema`](https://maplibre.org/maplibre-style-spec/schema) definition. State property value can be retrieved with the [`global-state`](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression.\n\n State is required when `global-state` is used in the style",
```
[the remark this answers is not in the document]
stanislawpuda-tomtom: It is better :).

src/expression/expression.test.ts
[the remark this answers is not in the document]
stanislawpuda-tomtom: Ok.

src/validate/validate_state.test.ts
```diff
@@ -0,0 +1,23 @@
+import {validateState} from './validate_state';
+
+describe('Validate state', () => {
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: Added tests

src/validate/validate_schema.ts
zbigniewmatysek-tomtom: Added tests for validate_state

src/expression/definitions/global_state.ts
```diff
@@ -0,0 +1,44 @@
...
+        if (args.length !== 2)
+            return context.error(`Expected 1 argument, but found ${args.length - 1} instead.`) as null;
+
+        const key = args[1];
+
+        if (typeof key !== 'string') {
+            return context.error(`Global state property must be string, but found ${typeof args[1]} instead.`) as null;
+        }
+
+        if (!key) return null;
```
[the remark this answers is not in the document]
zbigniewmatysek-tomtom: Fixed

build/schema.md
zbigniewmatysek-tomtom: Answered in https://github.com/maplibre/maplibre-style-spec/pull/1044#discussion_r2024346465

build/schema.md
zbigniewmatysek-tomtom: "schema" keywords were removed from the docs, https://github.com/maplibre/maplibre-style-spec/pull/1044#discussion_r2024346465

build/generate-docs.ts
zbigniewmatysek-tomtom: I've updated this a little - now v8 description is concatenated with a partial describing the state structure.

src/validate/validate_schema.ts
zbigniewmatysek-tomtom: Yes and no :)

First, representing this structure in v8 schema was problematic, for example "type" keyword is restricted in v8 and in state it's simply a property name with a fixed value.
Secondly, my validators are referencing other properties, or are more complex and would require custom validators anyway - like enum, array, or number with minimum and maximum.

I tried to use built in validation system at first, but it seemed not fit for the job without major adjustments. 

I welcome suggestions from you and others more experienced with this code base.

zbigniewmatysek-tomtom:
@HarelM thanks for initial comments! The PR is now ready for review.

src/validate/validate_schema.ts
HarelM: To be honest, I would like to avoid introducing JSON schema as part of this feature.
I would like to know what the limitations were of using the current type system?

build/state_data_types_partial.md
```diff
@@ -0,0 +1,75 @@
+## Data types
```
HarelM, line 1: Does it also support expressions?

src/validate/validate_global_state_expression.test.ts
```diff
@@ -0,0 +1,69 @@
...
+                'in',
+                ['get', 'payment_type'],
+                ['global-state', 'ev_preferred_emsp']
+            ]
+        ]
+    ];
+
+    const parsedExpression = createExpression(expression);
+
+    if (parsedExpression.result === 'error') {
```
HarelM, line 54: Logic is not recommended in tests.
Better to use `expect` instead.

src/validate/validate_global_state_expression.test.ts
```diff
@@ -0,0 +1,69 @@
+import {createExpression} from '..';
+import {findGlobalStateExpressionKeys} from './validate_global_state_expression';
+
+test('findGlobalStateExpressionKeys extracts all global state keys', () => {
```
HarelM, line 4: Use `describe` for the test suite please.

src/validate/validate_global_state_expression.ts
```diff
@@ -0,0 +1,22 @@
+import {getOwn} from '../util/get_own';
+import {Expression} from '../expression/expression';
+import {GlobalState} from '../expression/definitions/global_state';
+import {ValidationError} from '../error/validation_error';
+
+export function validateIfDefinedInGlobalState(key: string, options: any) {
```
HarelM, line 6: I didn't see a unit test that cover this method, did I miss it?

src/validate/validate_state.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    test('Should return error if type is not an object', () => {
```
HarelM, line 4: Please split the test into several test and use the test name to describe the different scenarios.

src/validate/validate_state.test.ts
```diff
@@ -0,0 +1,149 @@
...
+            expect(errors[0].message).toBe('state.someNumber.default: must be greater than or equal to 2');
+        });
+        test('should return error if minimum is not a number', () => {
+            const errors = validateState({
+                key: 'state',
+                value: {someNumber: {type: 'number', default: 1, minimum: '2'}}
+            });
+            expect(errors).toHaveLength(1);
+            expect(errors[0].message).toBe('state.someNumber.minimum: must be a number');
+        });
```
HarelM, line 62: Nit-picking: better having a newline after each test, for better readability.

src/validate/validate_state.test.ts
```diff
@@ -0,0 +1,149 @@
...
+            const errors = validateState({
+                key: 'state',
+                value: {someNumber: {type: 'number', default: 1, maximum: '2'}}
+            });
+            expect(errors).toHaveLength(1);
+            expect(errors[0].message).toBe('state.someNumber.maximum: must be a number');
+        });
+    });
+
+    describe('string', () => {
```
HarelM, line 73: I'm not sure this is a good name for a test "suite".

src/validate/validate_state_property.ts
```diff
@@ -0,0 +1,214 @@
...
+import {getType} from '../util/get_type';
+import {deepUnbundle, unbundle} from '../util/unbundle_jsonlint';
+import {isObjectLiteral} from '../util/is_object_literal';
+
+interface ValidateStatePropertyOptions {
+    key: string;
+    value: unknown;
+}
+
+export function validateStateProperty({
```
HarelM, line 11: Is it possible that this method will return a list of validation errors? I see it only returns an array of a single error, am I missing anything?

src/validate/validate_state_property.ts
```diff
@@ -0,0 +1,214 @@
...
+                ];
+            }
+        }
+    }
+
+    return validateEnum(value, key);
+}
+
+function validateString(schema: Record<string, unknown>, key: string) {
+    const defaultValue = unbundle(schema.default) as string;
```
HarelM, line 61: Is `unbundle` needed in all these places?

src/validate/validate_expression.ts
```diff
@@ -42,5 +43,11 @@ export function validateExpression(options: any): Array<ValidationError> {
         }
     }
 
-    return [];
+    const globalStateExpressionKeys = new Set<string>()
+
+    findGlobalStateExpressionKeys(expressionObj, globalStateExpressionKeys);
+
+    return Array.from(globalStateExpressionKeys).map((globalStateExpressionKey: string) => {
```
HarelM, line 50: Is there a better way to iterate over this set? This looks a bit cumbersome...

src/reference/v8.json
```diff
@@ -97,6 +97,32 @@
...
+          "type": "number",
+          "minimum": 0,
+          "default": 50
+        }
+      },
+      "sdk-support": {
+        "basic functionality": {
+          "js": "https://github.com/maplibre/maplibre-gl-js/issues/4964",
+          "android": "https://github.com/maplibre/maplibre-native/issues/3302",
+          "ios": "https://github.com/maplibre/maplibre-native/issues/3302"
```
louwers, line 122: Please include some additional information in the Native ticket. Can just be copy pasted, but issues should be self-contained.

build/state_data_types_partial.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+## Default values
+
+All data types can have optional `default` property. The value of the `default` property will be validated against the data type during style validation.
+
+
+## Annotations
+
+Properties can be annotated with optional `title` and `description` keywords. A "title" will preferably be short, whereas a "description" will provide a more lengthy explanation of the purpose of state property.
```
louwers, line 75: We don't have this for sources and layers, what is the use-case for introducing it here?

louwers: Aha, read your comment, Maputnik...

src/validate/validate_global_state_expression.test.ts
zbigniewmatysek-tomtom: Changed to `fail`. Check is needed, otherwise TS is complaining.

src/validate/validate_global_state_expression.test.ts
zbigniewmatysek-tomtom: Done

src/validate/validate_global_state_expression.ts
zbigniewmatysek-tomtom: Refactored end tested the exported function.

src/validate/validate_state.test.ts
zbigniewmatysek-tomtom: Those tests were redundant, I removed them.

src/validate/validate_state.test.ts
zbigniewmatysek-tomtom: Fixed

src/validate/validate_state.test.ts
zbigniewmatysek-tomtom: Improved naming

src/validate/validate_state_property.ts
zbigniewmatysek-tomtom: Array validator can return multiple errors, others return array for consistency.

I've added a test for that path:
image

src/validate/validate_expression.ts
zbigniewmatysek-tomtom: Refactored, I think it's much better now

build/state_data_types_partial.md
zbigniewmatysek-tomtom: Not as the part of this proposal. 

For now I `global-state` to work similarly to `get`, meaning returned value will be interpreted as JSON types:
>  The literal operator is not necessary if the array is returned from a sub-expression, e.g. ["in", 1, ["get", "myArrayProp"]].

It seems feasible to support expressions, but I did not investigate that.

build/state_data_types_partial.md
zbigniewmatysek-tomtom: Make sense, descriptions can be put in metadata, I removed this section.

src/validate/validate_state_property.ts
zbigniewmatysek-tomtom: typeof checks fail otherwise. I think it's safer to cast to primitive types explicitly and` if type is already primitive, unbundle` internally will just return it:

image

louwers (changes requested):
A style is a machine-readable description of the visual appearance of a map. In my opinion application-level concerns such as modifying a style should be handled with application code.

I understand that this feature is intended to ease the development of styles. You can quickly modify variables which are used in various places in a style with tools like Maputnik so see how it impacts the visual appeal. However, we do not need this feature as part of the style spec for this. We can develop some sort standardized style template similar to what @sjg-wdw suggests in [this comment](https://github.com/maplibre/maplibre-style-spec/issues/886). It does not make sense to add features to the style spec just to make authoring styles easier, we should focus on make better tools for authoring styles instead.

Another use case for this feature the ability to author different 'versions' of styles. But a common pattern for this already exists: create some sort of framework that outputs a style, and set this output as a new style

```
style = StyleBuilder(styleTemplate)
  .lightMode()
  .showBikeLanes()
  .hideTrafficlights.build();
map.setStyle(style)
```

If it is inefficient to set a new style like this, we should work on diffing a style with a new incoming style and making that more efficient. 

The existing `feature-state` functionality already mixes application-level concerns with style level concerns. I think a better way to implement that would have been to modify features directly, and rely on `get`. In any case, just because we have one feature like this does not mean we have to go down this path further...

What we discussed during the TSC Meeting was a new `[global-state, default]` expression. If I recall correctly a new root property was not discussed. In any case, I'm still not convinced. Let's discuss some more if we should really add this, or if the goals of this functionality can indeed better be solved by other means than via a style spec extension.

zbigniewmatysek-tomtom:
@louwers Thanks for the write up! Let me explain our motivations in more detail and why having style building layer was not selected by us as the optimal solution, and why we choose to make this contribution.

One of the main reasons for adding global-state was to decouple style updates from the software that uses those styles, making integration easier.

1. Let’s say we want to support filtering EV POIs based on charger type. Filter expressions are complex, and if we implement this on the client side, the SDK or client code would need to handle parsing these expressions and updating the relevant parts of the runtime - often changing the structure of the expression in the process.

2. Now, imagine a new minor version of the style is released. This update adds a new POI category and lowers the priority of an existing one. It also changes the structure of the filter expression. In this case, the client-side logic would probably need to be updated or at least tested before the update can be pushed to clients. **This isn’t ideal, as it slows down the update process and adds integration overhead for each minor style update**.

Another reason for having this feature in the renderer and the style spec instead of client side code / SDK, is that **renderer is a single piece of code common for most of the platforms**. We have SDKs for Android, iOS and automotive written in distinct technologies. Even if we develop common component for building the style, it requires integration on all platforms. When we have this logic in the renderer we are closer to "build once, ship everywhere".


@sjg-wdw Great to see the renderer maintainer perspective! While I am not familiar with the maplibre native architecture, I will try to address some of the concerns you mentioned based on what I know about gl-js.

In maplibre there are two main types of runtime style modification functions (simplifying):
* Those that you can call every frame and are cheap and can be done via shaders, e.g.: setting paint properties
* Computationally expensive functions that require re-building geometry and are usually done is a separate thread, e.g.: setting layout properties and filter.

The `global-state` can be used in various places in the style and it's location determines if this is cheap or expensive. In the simplest implementation calling `setGlobalState(property, value)` would trigger sources reload + repaint as setting new style do.

Renderers _may_ built optimization around it, e.g.: via tracking in where expressions are used. Yes, this increases complexity. It still won't be that performant to use it for animations. This is not the objective to make it such. As long as the user knows the tradeoff and it does not affect other features, it is acceptable.

**I believe we need to weight the renderer complexity versus the simplicity of adoption and impact on the ecosystem.** 

Dynamic features that MapLibre has are one of the reasons why this standard became so successful. There will be always a dilemma of choosing between complexity, maintenance concerns and new features. Having seen the reception from community, I am confident that proposed direction solves common problem of many developers building products on top of MapLibre.

HarelM:
@zbigniewmatysek-tomtom can you elaborate a bit more as I'm not sure I fully understand why bart's suggestions to use tools to update the style "from out side" doesn't solve the same problem? What am I missing?

Let's say that Maputnik now supports a yaml format (since you can use "variables" in yaml) and when you decide to export it, it will generate a json file can be used with maplibre (this is more in the realm of how to implement this, but it's a bit more concrete for the discussion).
So you have a yaml file that you maintain and when you are happy with the changes you export it to json and upload it to be used by the clients.
Wouldn't that be enough for this use case?
I know that you won't be able to change it the same way in the render, but you mentioned that you want those separated, didn't you?

zbigniewmatysek-tomtom:
Our end goal is to deliver a dynamic map experience in the automotive context, where map style needs to be manipulated at runtime to react to a set of input (and such input cannot be addressed by a finite set of precomputed styles). Such input can be:
* user preferences (only display information that are relevant to a user in a specific context, depending on the task they need to perform)
* time/day settings
* vehicle parameters
* stage of the driving journey (e.g.: show selected information, like parking solutions or charging solutions, only at arrival)

Delivering multiple styles is not sustainable from maintenance standpoint, but also a huge overhead for integrators that need to manage them properly. Especially since these styles need to evolve through time, factoring in data source changes, specific customers variants. The multiple styles approach is inefficient, fragile and prone to error.

zbigniewmatysek-tomtom:
For context, we are already managing dozens of style variants, including browsing, navigation, light, dark, satellite, regional versions for Korea and Japan, renderer-specific, customer-specific, and their various combinations.

Adding more static variants is not a path we want to pursue.

Runtime style or expression manipulation, like what the style builder does, is prone to errors and difficult to verify with so many different variants.

In contrast, the proposed solution doesn’t modify the style expression or properties on the client side. Instead, it only changes the data passed to the evaluation context in a verifiable way, thanks to type checks.

zbigniewmatysek-tomtom:
@louwers Our apps and clients apps use online styles which get updated frequently. For example a client map use style version `22.3*` and will automatically get minor version updates without updating the software.

The "style builder" works for simple cases like changing the visibility or in situations where the contract is clear. When you manipulate complex expressions in the runtime or make changes across entire style (light/dark mode switch), this becomes fragile. A simple style update; adding a new layer, updating expression may break the "style builder" logic. I mentioned this in in https://github.com/maplibre/maplibre-style-spec/pull/1044#issuecomment-2775170227 Things that could be pushed to customers automatically, suddenly become breaking changes and requires integration effort.

Style builder needs to consider things like layer ids, layer order, expression structure (order, nesting). Please note, that style content may be different for different variants!

Another reason mentioned before is the portability. In TomTom we have 3 different stacks (excluding web). From what I can see, MapLibre Navigation SDK is also written in distinct technologies - Java/Kotlin for Android and Swift in iOS. If you have this implemented at the renderer level, those products will be immediately able to adopt it. 😉

zbigniewmatysek-tomtom:
@louwers 

> With the global state approach old clients will be tied to a possibly outdated set of variable names.

Could you elaborate on this? It seems to be also the case with `get` and it would be the same with templates?

> I am not suggesting that you should patch a style at runtime. 

Sorry, I didn’t realize you were talking about a template-based approach for the style builder - I missed that part from the earlier post.

We have considered the template system before proposing this solution.
* Our API is built around serving styles compatible with the style specification. Such styles are supported TomTom SDKs and MapLibre renderers. Following the standard brings value for both us and the customers.
* We are not in unique positions, clients in this ecosystem expect to be served styles from online APIs.
* Introducing a new style format for usage on the client requires changes across the entire stack - from the API adjustments, and style validation tooling, map editing software to SDKs and client code.
* We estimate the cost of such change to be one magnitude greater than implementing global-state solution. 
* Having 2 style formats will be confusing for the users. It brings new compatibility issues. 

> Or is reparsing and re-evaluating a style itself a blocker with the current implementation(s)
* Not a blocker, but templates do not seem to have any performance benefits to the proposed approach. The only optimization that can be done is style diffing. The new expression leaves room for more refined optimizations if needed.

 > In fact, I think forward compatibility is much easier to achieve with the style builder approach because you have much more control.
* Too much flexibility is also a liability.  It makes the solution hard to verify and integrate. Ideally we want styles to be as much statically testable as possible and reduce the chance of producing invalid output.

HarelM:
Let's use the next monthly meeting to complete the discussion around this.
I think the points made here are sound, from both perspectives, and we need to further understand what our options are going forward.

1ec5:
Unfortunately, the discussion got split between this PR and #886, but I’d like to elaborate on https://github.com/maplibre/maplibre-style-spec/issues/886#issuecomment-2719051890 in case I’m unable to make the next meeting:

Before there was a runtime styling API, Mapbox insisted that developers publish multiple styles, one for each variant that the application could switch to. Any developer who cared about localization would publish one for each language their application supported or each that their audience would likely speak. Developers painstakingly juggled multiple subtle variations on the same style, making mistakes in the process, or came up with postprocessing scripts to generate styles from a template. I think this is essentially the suggestion in https://github.com/maplibre/maplibre-style-spec/pull/1044#issuecomment-2778834470. While some developers were able to cope with this approach, overall it resulted in few developers supporting localization. Soon, users came to expect other personalizations, such as [dynamic text sizes](https://developer.apple.com/documentation/uikit/scaling-fonts-automatically) (popular among older users). Each additional personalization meant a combinatorial explosion of styles to publish.

Once the runtime styling API came along, developers quickly started to use this API as a sort of scripting technology. The native map SDKs and navigation SDKs have added built-in APIs like [`localizeLabels`](https://github.com/maplibre/maplibre-native/blob/d6938987577a2ccb6e50695dbf119223caef6adb/platform/darwin/src/MLNStyle.mm#L585-L607) and [`toggleTraffic`](https://github.com/maplibre/maplibre-navigation-ios/blob/f91faffa1aa03be16fc588dc5076b7811d1c5166/MapboxNavigation/MLNMapView.swift#L70-L77) that manipulate the style generically according to some common patterns. On the Web, [mapbox-gl-language](https://github.com/mapbox/mapbox-gl-language/), [mapbox-gl-traffic](https://github.com/mapbox/mapbox-gl-traffic/), and later [maplibre-gl-dates](https://github.com/OpenHistoricalMap/maplibre-gl-dates/) fill similar needs. Half a dozen times now, I’ve had to reimplement the same pattern for iterating through the entire style, [applying the same transformation to each layer](https://github.com/osm-americana/openstreetmap-americana/blob/c4829f2c91099c3b5619fff829a390133ea08a53/src/constants/label.js#L80-L134), [taking care not to transform a given property value too many times](https://github.com/OpenHistoricalMap/maplibre-gl-dates/blob/98c640785d6382de9a4022d27bc83b2d7c902cdd/index.js#L239-L249) or blow away any customization by some other add-on. I’ve considered adding it to the GL JS examples since it’s so unintuitive.

There has to be a better way, but I don’t see how templating would address this use case. In general, a style author should not need to know in advance which language or font size the end user’s operating system is set to. That would allow them to tailor the typography for a particular configuration, but it shouldn’t be a baseline requirement, or most won’t bother and users won’t benefit. A particularly circumspect style author could make sure to wrap absolutely every expression they write in a `let` expression that has slots for `language` etc., but that’s extremely error-prone and an unnecessary burden on style authors. Instead, the style author should be able to factor out all the occurrences of `name` into a `localized_name` global variable that can be manipulated at runtime. This could be useful even if nothing changes at runtime, just for ensuring that important inputs can change easily as needed.

zbigniewmatysek-tomtom:
@HarelM @louwers Please take another look 😉 

As discussed on the call:
* Removed "types"
* Removed requirement that property must be defined in `state` if it's referenced by the `global-state`
* Kept `state` with optional `default`
* Updated docs
* I am going to move the "types" and validations to a separate proposal, once this one is merged

PR description was updated to reflect the changes.

HarelM (approved):
LGTM, THANKS!

HarelM:
I think the last comment that was not addressed is the following:
https://github.com/maplibre/maplibre-style-spec/pull/1044#discussion_r2024582078
Around making the native issue self-descriptive.

zbigniewmatysek-tomtom:
@HarelM updated native issue: https://github.com/maplibre/maplibre-native/issues/3302

src/reference/v8.json
```diff
@@ -97,6 +97,26 @@
         }
       }
     },
+    "state": {
+      "type": "state",
+      "default": {},
+      "doc": "An object defining global state properties of the map style and their default values. State property value can be retrieved with the [`global-state`](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression.",
```
louwers, line 103:
```suggestion
      "doc": "An object used to define default values when using the [`global-state`](/maplibre-style-spec/expressions/#global-state) expression.",
```

Reworded a bit for clarity.

src/reference/v8.json
```diff
@@ -3303,6 +3323,24 @@
           }
         }
       },
+      "global-state": {
+        "doc": "Retrieves a property value from the [state](https://maplibre.org/maplibre-style-spec/root/#state) property. Returns `null` if the property is not defined",
```
louwers, line 3327:
```suggestion
        "doc": "Retrieves a property value from global state that can be set with platform-specific APIs. Defaults can be provided using the [`state`](/maplibre-style-spec/root/#state) root property. Returns `null` if no value nor default value is set for the retrieved property.",
```

The current wording makes it seem that this property *only* retrieves values from the `state` property.

louwers (approved):
Please see my comments for suggestions on slightly rewording the documentation for clarity.

Otherwise LGTM.

hiddewie:
@zbigniewmatysek-tomtom Incredible work! Much appreciated how you move the proposal from https://github.com/maplibre/maplibre-gl-js/issues/4964 forward!

HarelM:
@zbigniewmatysek-tomtom can you please address the last few comments, so I'll be able to release a new version of the style-spec package?

HarelM:
@louwers feel free to push the merge button once you finish the review. 
LGTM.

Issue #868, opened 2024-10-21
Flexible font define in style


alanchenboy:
# Design Proposal: Flexible font define in style

## Motivation

MapLibre inherits the font pbf design from Mapbox, this is a limit to rendering complex text. We must let the core reach the font file to finish shaping and rendering complex text like Indic, Khmer Burmese, etc.

## Proposed Change

The design looks like CSS unicode-range, 

  ```
 "font-faces": [
        {
            "text-font": ["Noto Sans Regular"],
            "font-files": [
                {
                    "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
                    "unicode-range": [
                        "U+1780-17FF"
                    ]
                },
                {
                    "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
                    "unicode-range": [
                        "U+1000-109F", "U+A9E0-0xA9FF", "U+AA60-0xAA7F"
                    ]
                },
                {
                    "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
                    "unicode-range": [
                        "U+0900-097F", "U+A8E0-0xA8FF"
                    ]
                }
                
            ]
        }
    ],
```
full style:
https://alanchenboy.github.io/harfbuzzResource/hindi.json

## API Modifications

A new feature for complex font rendering, it's transparent for the map users.  

Here is an example of how to render complex text on the native side.  
https://github.com/maplibre/maplibre-native/pull/1439.  
## Migration Plan and Compatibility

N/A

## Rejected Alternatives

N/A

HarelM:
Thanks for taking the time to open this design proposal.
Can you clarify the relation between this property and `glyphs`? It is not clear when one should be selected over the other.
Also, will using this require some configuration in the rendering part? special build? a plugin?
We don't have other style properties that has external requirements in order to work as far as I know...

louwers:
We would use Harfbuzz for text shaping in MapLibre Native by default, so no special plugins would be required.

HarelM:
Wouldn't this require a special build? Or is the default build comes with Harfbuzz and you can reduce build size if you need, but this would be the special build - i.e without it?

louwers:
No, we would enable it by default, although it would be possible to drop out if you have binary size constraints and you make your own builds.

[...]

PR #1068, opened 2025-03-26
Relax requirement for `glyphs` when using `text-field`

### LINKED ISSUE #1045: Extended TinySDF text rendering
# Design proposal: Extended TinySDF text rendering[^funding]

## Motivation

In the style specification, the root [`glyphs`](https://maplibre.org/maplibre-style-spec/root/#glyphs) property points to a series of server-side glyph PBFs. This property is technically optional, but required if any symbol layer has a [`text-field`](https://maplibre.org/maplibre-style-spec/layers/#text-field) property.

The glyph PBF format has several limitations, including:

* Glyph PBFs require special tooling instead of the much more widespread tooling for custom fonts that is available on each platform. They need to be hosted somewhere, as opposed to system-installed fonts or Web fonts that can be loaded from an existing CDN.
* Downloading glyph PBFs wastes bandwidth when the surrounding webpage or application already has to download the same fonts in a different format. It wastes either storage space or time when downloading offline maps, as the client needs to download all the possible glyph PBFs or scan all the downloaded tiles for `text-field`s to determine which glyph PBFs to download.
* If the specified glyph PBFs don’t load for some reason, the user is unable to see the text at all: maplibre/maplibre-gl-js#4563.
* Expanding [text rendering support beyond the Basic Multilingual Plane](https://github.com/maplibre/maplibre-gl-js/issues/2307) would require a webmaster to generate and host inordinately more glyph PBFs.
* An assumed one-to-one correspondence between Unicode codepoints and font glyphs does not hold in writing systems that require contextual text shaping or ligatures. This will hamper future efforts to support additional writing systems and improve support for right-to-left scripts.[^one-to-one]
* Some specialized clients lack a viable path to supporting custom fonts in MapLibre stylesheets: qgis/QGIS#46588.

> [!NOTE]
> “[About Text Rendering in MapLibre](https://github.com/wipfli/about-text-rendering-in-maplibre/)” has a more detailed background of MapLibre’s text rendering architecture, which has heavily influenced the text-related portions of the style specification.

## Proposed change

Leaving `glyphs` unset would no longer be an error if the stylesheet contains a text layer. When this is the case, the client would be responsible for rendering the text however it can. The client would interpret [`text-font`](https://maplibre.org/maplibre-style-spec/layers/#text-font) as a fallback list of font names, which can be influenced by application-specific configuration options external to the stylesheet.

For its part, MapLibre would extend the existing TinySDF path for local text rendering to cover every codepoint, not just ideographic code blocks. The `text-font` values would be fed into TinySDF to influence font selection; the details differ by platform.

## API modifications

`glyphs` would no longer be required under any circumstance, not even when a symbol layer has a `text-field` property.

In the absence of `glyphs`, the `text-font` property would be interpreted as a fallback list of font names, instead of a concatenated fontstack name. This has long been the case in [MapLibre Native for iOS](https://maplibre.org/maplibre-native/ios/latest/documentation/maplibre-native-for-ios/customizing_fonts#Client-side-fonts) in its local text rendering codepath. However, with this change, other clients would adopt the same behavior for consistency.

## Migration plan and compatibility

Existing styles can continue to specify `glyphs` as before.

To benefit from improved performance, robustness, and (in the future) language support, a style author can omit the `glyphs` property. They would need to ensure that any `text-font` property names fonts that exist in the current environment, such as system fonts. Custom fonts would require additional setup depending on the platform:

* On the Web, add an `@font-face` ruleset to the surrounding webpage’s CSS stylesheet and wait until [`document.fonts.onloadingdone`](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/loadingdone_event) to load the MapLibre style.
* On iOS, bundle the font with the application or [load it as an on-demand resource](https://developer.apple.com/documentation/xcode/configuring-custom-fonts) then register it with Core Text.
* On Android, bundle the font with the application or [load it as a downloadable font](https://developer.android.com/develop/ui/views/text-and-emoji/downloadable-fonts).

Otherwise, any text will be rendered using a last-resort font. This font will not look spectacular, but it will be a clear usability improvement over the current behavior of failing to render the surrounding tile: maplibre/maplibre-gl-js#4563.[^last-font]

Extended TinySDF will be implemented in stages, starting with GL JS. In the meantime, developers may be reluctant to adopt extended TinySDF unless each platform is capable of loading a `glyphs`-less stylesheet without errors. For interim compatibility, MapLibre Native will be modified so that `glyphs` defaults to the following URL template:

```
https://demotiles.maplibre.org/font/Open%20Sans%20Regular,Arial%20Unicode%20MS%20Regular/{range}.pbf
```

This URL template is based on the default value of `text-font`, Open Sans Regular followed by Arial Unicode MS Regular. This fontstack has been available from demotiles.maplibre.org since maplibre/demotiles#1 and is being used in the wild. It appears as [an example](https://maplibre.org/maplibre-style-spec/root/#glyphs) in the style specification and various other MapLibre documentation, and it has also spread through copy-pasting from one stylesheet to another.[^arial] The URL template ignores any explicitly set `text-font`, because only a few common fontstacks have been uploaded to the demo; any unavailable fontstack will result in the tile failing to load, which is worse than falling back to a generic font.

Later, once extended TinySDF support is ported to MapLibre Native on Android and iOS, this fallback value can be removed in favor of local rendering. Ultimately, there would be no guarantee of the resulting rendering looking consistent pixel for pixel across platforms, due to font variation that’s out of our control. Developers who need precise cross-platform consistency should continue to specify `glyphs` and rely on server-rendered fonts.

These are all implementation details. From the style specification’s standpoint, the client would simply need to render the text somehow.

### Protocol

Via maplibre/maplibre-gl-js#2555, [maplibre-local-glyphs](https://github.com/spatialillusions/maplibre-local-glyphs/) implements a protocol that produces individual glyphs locally on the fly, similar to this proposal. While this is a workable stopgap, it is platform-specific and does not provide a path to addressing some of the limitations inherent to the glyph PBF format, such as the codepoint–glyph equivalence.

### Root font definitions

My proposal overlaps with #868, which would introduce a new root property to register TrueType files and map them to codepoint ranges. I find it problematic that this proposal would have the stylesheet specify a font URL upfront. An iOS developer may prefer to bundle the fonts with the application or [load it as an on-demand resource](https://developer.apple.com/documentation/xcode/configuring-custom-fonts) from a URL that the style author would not know in advance, but which definitely would not be compatible with the Web version of the same application.

Moreover, the proposal’s choice of TrueType font URLs is problematic for the Web. Some foundries such as MonoType and Adobe only license TrueType fonts for desktop use; only .woff files are provided for Web use, and they [make developers use specific CSS or JavaScript code](https://helpx.adobe.com/fonts/using/webfont-licensing.html#web-sites) to load it. On the other hand, popular sources of open source fonts, such as Google Fonts, don’t provide direct access to download a font file; you have to load the font using [a remote CSS stylesheet](https://fonts.google.com/selection/embed).

Under my proposal, loading fonts becomes an application responsibility. Any font definitions that MapLibre Native might eventually require would be added to MapLibre Native’s public API rather than the style specification. This keeps the overall workflow straightforward and aligned with standard platform development practices.

### Positioned glyph font

For improved support of certain writing systems, @wipfli has implemented a [positioned glyph font](https://oliverwipfli.ch/devanagari-in-the-protomaps-basemap-with-a-positioned-glyph-font-for-maplibre-2024-06-30/) architecture that reuses the existing server-rendered font pipeline. Unlike in a conventional setup, the glyph PBFs contain preshaped glyphs at custom indices, and the vector tiles contain what appear to be strings of gibberish but are actually matching glyph indices. The advantages of this solution are that it can be deployed to all platforms without changing MapLibre on each platform, and the graphical quality is superior to text rendered by TinySDF on the client side.

Unfortunately, this approach requires special software to generate both the glyph PBFs and vector tiles. The choice of glyphs is dependent on existing OSM data, whereas MapLibre would ideally remain dataset independent to a large degree. The tight coupling between glyph PBFs and vector tiles also forces a tight coupling to a specific font face. This tradeoff may be acceptable to some language communities that are currently poorly served by MapLibre, but other proprietary map libraries are capable of the same language support plus custom fonts, so we should aim for a more straightforward solution as well.

## Proof of concept

1ec5/maplibre-gl-js#1 contains a proof of concept implementation of this and some other major changes that I have yet to formally propose. Based on this branch, [a proof of concept](https://bdon.github.io/maplibre-text/variable-fonts/) by @bdon demonstrates using a custom variable font in an otherwise standard MapLibre map, something that would be infeasible in the current font architecture or any of the alternative proposals. This proposal is a first step in taking some of the ideas in these proofs of concept and reimplementing them in a state suitable for release, accounting for backwards compatibility.

[^funding]: This work will be supported by [an NLnet grant](https://nlnet.nl/project/Protomaps/).
[^one-to-one]: MapLibre’s text segmentation behavior badly needs revisiting, but nothing about the style specification pertains to text segmentation, so I’ve left this aspect out of the proposal. We can track that work in client issues such as maplibre/maplibre-gl-js#2384 maplibre/maplibre-gl-js#3215. There isn’t an inherent requirement to coordinate fixes for this problem across all the platforms.
[^last-font]: Currently, the style specification defines the default value of `text-font` as Open Sans followed by Arial Unicode MS. An unwritten caveat is that ideographic text rendering already eschews these fonts in favor of other platform-specific selections. Should we modify the style specification to hedge about the default behavior in this case?
[^arial]: We would want to wean ourselves off this dependency later, due to poor aesthetic quality and what some perceive to be a questionable legal situation: maplibre/maplibre-gl-js#4820 https://github.com/maplibre/maplibre-style-spec/issues/311#issuecomment-2717319680.

1ec5:
Adjusted the `glyphs` documentation to relax the informal requirement for this property when `text-field` occurs in a symbol layer. The documentation now matches the machine-readable specification in classifying `glyphs` as optional. Also clarified the interpretation of `text-font` when `glyphs` is set versus unset. When it’s unset, the array is interpreted as a cascading fallback list. This is how it already works on each platform when setting the ideographic font, but now it‘s explicit and coupled to another part of the style specification.

Fixes #1045.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] ~~Include before/after visuals or gifs if this PR includes visual changes.~~
 - [ ] ~~Write tests for all new functionality.~~
 - [ ] ~~Post benchmark scores.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- `glyphs` is now optional even if a symbol layer specifies `text-field`; if it is unset, `text-font` is interpreted as a fallback font list ([#1068](https://github.com/maplibre/maplibre-style-spec/pull/1068))
 
 ### 🐞 Bug fixes
 - Fix RuntimeError class, make it inherited from Error ([#983](https://github.com/maplibre/maplibre-style-spec/issues/983))
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -534,6 +534,11 @@ function createMainTopics() {
                 content += convertPropertyToMarkdown(subKey, subValue as JsonObject);
             }
         }
+        if (value['sdk-support']) {
+            content += '\n---\n';
+            content += sdkSupportToMarkdown(value['sdk-support']);
+        }
+
         fs.writeFileSync(`${BASE_PATH}/${key}.md`, content);
     }
 }
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -183,8 +183,20 @@
     },
     "glyphs": {
       "type": "string",
-      "doc": "A URL template for loading signed-distance-field glyph sets in PBF format.\n\nIf this property is set, any text in the `text-field` layout property is displayed in the font stack named by the `text-font` layout property based on glyphs located at the URL specified by this property. Otherwise, font faces will be determined by the `text-font` property based on the environment.\n\nThe URL must include:\n\n - `{fontstack}` - When requesting glyphs, this token is replaced with a comma separated list of fonts from a font stack specified in the `text-font` property of a symbol layer. \n\n - `{range}` - When requesting glyphs, this token is replaced with a range of 256 Unicode code points. For example, to load glyphs for the Unicode Basic Latin and Basic Latin-1 Supplement blocks, the range would be 0-255. The actual ranges that are loaded are determined at runtime based on what text needs to be displayed.\n\nThe URL must be absolute, containing the [scheme, authority and path components](https://en.wikipedia.org/wiki/URL#Syntax).",
-      "example": "https://demotiles.maplibre.org/font/{fontstack}/{range}.pbf"
+      "doc": "A URL template for loading signed-distance-field glyph sets in PBF format.\n\nIf this property is set, any text in the `text-field` layout property is displayed in the font stack named by the `text-font` layout property based on glyphs located at the URL specified by this property. Otherwise, font faces will be determined by the `text-font` property based on the local environment.\n\nThe URL must include:\n\n - `{fontstack}` - When requesting glyphs, this token is replaced with a comma separated list of fonts from a font stack specified in the `text-font` property of a symbol layer. \n\n - `{range}` - When requesting glyphs, this token is replaced with a range of 256 Unicode code points. For example, to load glyphs for the Unicode Basic Latin and Basic Latin-1 Supplement blocks, the range would be 0-255. The actual ranges that are loaded are determined at runtime based on what text needs to be displayed.\n\nThe URL must be absolute, containing the [scheme, authority and path components](https://en.wikipedia.org/wiki/URL#Syntax).",
+      "example": "https://demotiles.maplibre.org/font/{fontstack}/{range}.pbf",
+      "sdk-support": {
+        "basic functionality": {
+          "js": "0.0.16",
+          "android": "0.1.1",
+          "ios": "0.1.0"
+        },
+        "omit to use local fonts": {
+          "js": "https://github.com/maplibre/maplibre-gl-js/issues/3302",
+          "android": "https://github.com/maplibre/maplibre-native/issues/165",
+          "ios": "https://github.com/maplibre/maplibre-native/issues/165"
+        }
+      }
     },
     "transition": {
       "type": "transition",
@@ -1889,7 +1901,7 @@
         "Open Sans Regular",
         "Arial Unicode MS Regular"
       ],
-      "doc": "Fonts to use for displaying text. If the `glyphs` root property is specified, this array is joined together and interpreted as a font stack name. Otherwise, it is interpreted as a cascading fallback list of font names.",
+      "doc": "Fonts to use for displaying text. If the `glyphs` root property is specified, this array is joined together and interpreted as a font stack name. Otherwise, it is interpreted as a cascading fallback list of local font names.",
       "requires": [
         "text-field"
       ],
@@ -1903,6 +1915,11 @@
           "js": "0.43.0",
           "android": "6.0.0",
           "ios": "4.0.0"
+        },
+        "local fonts": {
+          "js": "https://github.com/maplibre/maplibre-gl-js/issues/3302",
+          "android": "https://github.com/maplibre/maplibre-native/issues/165",
+          "ios": "https://github.com/maplibre/maplibre-native/issues/165"
         }
       },
       "expression": {
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -183,7 +183,7 @@
     },
     "glyphs": {
       "type": "string",
-      "doc": "A URL template for loading signed-distance-field glyph sets in PBF format. \n\nThe URL must include:\n\n - `{fontstack}` - When requesting glyphs, this token is replaced with a comma separated list of fonts from a font stack specified in the text-font property of a symbol layer. \n\n - `{range}` - When requesting glyphs, this token is replaced with a range of 256 Unicode code points. For example, to load glyphs for the Unicode Basic Latin and Basic Latin-1 Supplement blocks, the range would be 0-255. The actual ranges that are loaded are determined at runtime based on what text needs to be displayed.\n\nThis property is required if any layer uses the `text-field` layout property. The URL must be absolute, containing the [scheme, authority and path components](https://en.wikipedia.org/wiki/URL#Syntax).",
+      "doc": "A URL template for loading signed-distance-field glyph sets in PBF format.\n\nIf this property is set, any text in the `text-field` layout property is displayed in the font stack named by the `text-font` layout property based on glyphs located at the URL specified by this property. Otherwise, font faces will be determined by the `text-font` property based on the environment.\n\nThe URL must include:\n\n - `{fontstack}` - When requesting glyphs, this token is replaced with a comma separated list of fonts from a font stack specified in the `text-font` property of a symbol layer. \n\n - `{range}` - When requesting glyphs, this token is replaced with a range of 256 Unicode code points. For example, to load glyphs for the Unicode Basic Latin and Basic Latin-1 Supplement blocks, the range would be 0-255. The actual ranges that are loaded are determined at runtime based on what text needs to be displayed.\n\nThe URL must be absolute, containing the [scheme, authority and path components](https://en.wikipedia.org/wiki/URL#Syntax).",
       "example": "https://demotiles.maplibre.org/font/{fontstack}/{range}.pbf"
     },
     "transition": {
@@ -1889,7 +1889,7 @@
         "Open Sans Regular",
         "Arial Unicode MS Regular"
       ],
-      "doc": "Font stack to use for displaying text.",
+      "doc": "Fonts to use for displaying text. If the `glyphs` root property is specified, this array is joined together and interpreted as a font stack name. Otherwise, it is interpreted as a cascading fallback list of font names.",
       "requires": [
         "text-field"
       ],


HarelM (approved):

HarelM:
LGTM. Waiting for @louwers review as well.

louwers:
MapLibre Native seems to break down pretty bad when `glyphs` is missing.

image

No graceful degradation like MapLibre GL JS. 🫤

image

louwers (changes requested):
I am not in favor of this change, I think it is better to always require people to set `glyphs` (even if it is just a fallback font) and add APIs to override the glyphs property with custom font loading. This offers much better backward compatibility for future styles.

That said, I don't want to be the sole blocker if other people don't have an issue with this. That said there is something that I think we should definitely change before merging this:

We document what versions of MapLibre GL JS and MapLibre Native support which features, and which behavior. We should add a support matrix to this entry with 'optional when using fonts'. There should be a ❌ for Native along with a tracking issue, to make people aware that when they author styles without a `glyphs` property, all bets are off if that style is used with MapLibre Native (see comment above). Of course, users of older MapLibre GL JS versions need to understand too what version of MapLibre GL JS they need to use to take advantage of 'font loading from the environment'.

---

Again, I believe that font loading should be a responsibility of the style author. A style should incorporate everything that is needed to show the map. If the PBF format is too limited, we should devise and agree upon another cross-platform font format. Hand waving and just telling people to make sure fonts are available in 'the environment' is inconvenient for users and may split the ecosystem.

1ec5:
> MapLibre Native seems to break down pretty bad when glyphs is missing.

Would you accept a PR to MapLibre Native that implements maplibre/maplibre-gl-js#4564?

> If the PBF format is too limited, we should devise and agree upon another cross-platform font format. Hand waving and just telling people to make sure fonts are available in 'the environment' is inconvenient for users and may split the ecosystem.

If locally rendered fonts are problematic, then they’re already problematic with respect to CJK, because the developer only needs to name a font “in the environment”, if anything. The named font doesn’t have to be bundled and doesn’t even have to exist; the implementation falls back to a system font on both [iOS](https://github.com/maplibre/maplibre-native/blob/3fc93a0b024e34514dafcbb424db93593ff540be/platform/darwin/core/local_glyph_rasterizer.mm#L111-L117) and [Android](https://github.com/maplibre/maplibre-native/blob/3fc93a0b024e34514dafcbb424db93593ff540be/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/text/LocalGlyphRasterizer.java#L54).[^android] Locally rendered fonts have been enabled by default for CJK since GL JS v0.0.16, Android v5.3.0, and iOS v3.7.3. Any ecosystem split should’ve been apparent among CJK-serving developers by now.

> We document what versions of MapLibre GL JS and MapLibre Native support which features, and which behavior. We should add a support matrix to this entry with 'optional when using fonts'. There should be a ❌ for Native along with a tracking issue, to make people aware that when they author styles without a glyphs property, all bets are off if that style is used with MapLibre Native (see comment above). Of course, users of older MapLibre GL JS versions need to understand too what version of MapLibre GL JS they need to use to take advantage of 'font loading from the environment'.

Sure, I’ve added a compatibility table for `glyphs` and added to the existing compatibility table for `text-font`, linking to maplibre/maplibre-gl-js#3302 and maplibre/maplibre-native#165.

[^android]: Based on the behavior of [`Typeface.create()`](https://developer.android.com/reference/android/graphics/Typeface#create(android.graphics.Typeface,%20int)).

louwers:
Thanks! Looks like we still need to add some code to make sure the table renders (or add a table to the Markdown string):

image

> Locally rendered fonts have been enabled by default for CJK 

Sure, but it's a bit of a hack the way it's implemented, I think you agree. That's my whole point, font faces are an important part of the look and feel of maps. If we're going to rely on the environment for Western fonts too, we will have even more cases where the best we will be able to do is use a fallback font instead of the font intended by the style author.

> Would you accept a PR to MapLibre Native that implements https://github.com/maplibre/maplibre-gl-js/pull/4564?

Yes, but what font would you use?

What I would really like is if we could use web fonts with MapLibre Native too (and embed web fonts directly in a style). Something like [this](https://developer.chrome.com/blog/memory-safety-fonts) looks promising (we can compile [Rust code](https://maplibre.org/maplibre-native/docs/book/rust.html) into MapLibre Native now).

1ec5:
> Looks like we still need to add some code to make sure the table renders

I ran `WATCH=1 npm run generate-docs` per the readme and I’m seeing both compatibility tables. It should work just like for the `centerAltitude` root property.

SDK Support table under the glyphs heading

SDK Support table under the text-font heading, with an extra row for local fonts

The command didn’t generate anything else for me to check in. Can you try again locally?

>> Locally rendered fonts have been enabled by default for CJK
> 
> Sure, but it's a bit of a hack the way it's implemented, I think you agree. That's my whole point, font faces are an important part of the look and feel of maps. If we're going to rely on the environment for Western fonts too, we will have even more cases where the best we will be able to do is use a fallback font instead of the font intended by the style author.

It’s a matter of degree. Just yesterday, I was explaining the current text rendering system to some of the folks who maintain the AARoads Wiki, which proudly uses MapLibre ([example](https://wiki.aaroads.com/wiki/A1_(Uganda))). They were in disbelief that it relies on spritesheets and glyph atlases, which they regard as a video game hack from the ’80s. Rest assured, I’m not seeking to revamp WebGL’s native text drawing functionality. 😅

I agree that working directly with actual font data and a fit-for-purpose HarfBuzz would be superior to, essentially, trafficking in screenshots of text. I will make sure anything I do in this area facilitates rather than impeding work towards that goal. But in my view, the change I’m proposing is essential for both local font and complex text support in MapLibre. Complex text support requires us to change how we segment the text, so that each glyph in the atlas represents a string of arbitrary length rather than a single character. This is fundamentally incompatible with the existing glyph PBF format.

Going forward, we can let the style author choose between traditional glyph PBFs and “handwavy” fonts from the local environment, based on their priorities. But if we insist on relegating local fonts to a runtime override, style authors will have to juggle two different interpretations of `text-font` simultaneously. Practically speaking, the local option means they’ll include a generous selection of font names, to catch as many available fonts as possible on each platform, which will be wildly incompatible with the traditional comma-separated fontstack interpretation. We shouldn’t force the style author to upload a hideous set of Code2000 glyphs with a fontstack name of `Euphemia%20UCAS,Euphemia,Aboriginal%20Sans,OskiBlackfoot,Code2000,Everson%20Mono%20Unicode`,[^wiktionary] just to align it to a reasonably comprehensive local fallback list for indigenous Canadian place names.

If I may, MapLibre Native has been on a journey to support complex writing systems the right way for over two years, with little to show to end users so far. Meanwhile, I’ve demonstrated that extended SDF can address that need at least functionally. I support the HarfBuzz effort and want it to succeed, but I also think developers and users should benefit from something tangible in the meantime. In seeking visual perfection, we should not lose sight of the fact that a beautiful, consistent font is useless if the text is missing and potentially offensive if the text is left garbled.

>> Would you accept a PR to MapLibre Native that implements https://github.com/maplibre/maplibre-gl-js/pull/4564?
> 
> Yes, but what font would you use?

I’m offering to implement the full feature in both repositories, without a hard-coded fallback font. This means the style author would choose the fonts based on what they provide or what’s likely to be available on the platform, as they’re already used to doing. I had previously floated the idea of a hard-coded fallback to buy myself some time before committing to the native implementation, but at this point there’s little point in holding off on a full implementation. As a practical matter, I probably would still stage complex text segmentation in GL JS first, but that shouldn’t be an interoperability concern.

> What I would really like is if we could use web fonts with MapLibre Native too (optionally also embed web fonts directly in a style). Something like [this](https://developer.chrome.com/blog/memory-safety-fonts) looks promising (we can compile [Rust code](https://maplibre.org/maplibre-native/docs/book/rust.html) into MapLibre Native now). Because parity between MapLibre GL JS and MapLibre Native in terms of how a particular style looks is one of our aims.

A typical MapLibre-powered application consists of more than a map. Very often, a developer needs to apply a corporate font across all three platforms. The Web almost always uses a different font asset than the mobile platforms, distributed through different means, even for a visually identical typeface. There are both technical and legal reasons for this, as I pointed out in https://github.com/maplibre/maplibre-style-spec/issues/868#issuecomment-2721877794. This isn’t a new problem, so let’s take advantage of the solutions that developers are already using for their applications.

We can explore Web fonts for the native platforms separately. If extended SDF ships on Android or iOS without Web font support, it just means developers who are already adequately served by glyph PBFs will stick to glyph PBFs, while those who need complex text will settle for system fonts as a last resort.

[^wiktionary]: [More where that came from](https://en.wiktionary.org/wiki/MediaWiki:Gadget-LanguagesAndScripts.css).

louwers:
Thanks for taking the time to so argue this change so extensively.

> Can you try again locally?

Ah right, it shows up in the 'root properties' page, but I was looking at http://localhost:8000/maplibre-style-spec/glyphs/

I will have a look how we can add it there too.

> They were in disbelief that it relies on spritesheets and glyph atlases, which they regard as a video game hack from the ’80s. 

Lol. To be fair, they probably weren't using SDFs for glyphs in the 80's?

> I’m offering to implement the full feature in both repositories, without a hard-coded fallback font.

That would be great. I still believe that font distribution should be part of styles to make sure works reliably everywhere. There is quite a lot of variability in terms of what fonts are available in the Native world. But that can still be addressed with other style spec proposals.

louwers (approved):

1ec5:
Thanks for fixing the other page; didn’t realize it was there too.

#868 does make sense from a certain perspective. A webpage is to a CSS stylesheet defining `@font-face` as a map in the webpage is to a MapLibre stylesheet defining font locations. In the native Web stack, font substitution makes this tenable. Hopefully we can do both, in which case the self-contained fonts will take precedence.

Issue #1067, opened 2025-03-25
Design Proposal: Hypsometric Tint


NathanMOlson:
# Design Proposal: Hypsometric Tint

## Motivation

https://github.com/maplibre/maplibre-gl-js/issues/5666

I want to display elevation using hypsometric tints (color mapping). https://en.wikipedia.org/wiki/Hypsometric_tints. This would behave like GDAL's `gdaldem color-relief` https://gdal.org/en/stable/programs/gdaldem.html#modes. 

## Proposed Change

I propose a new layer type called `color-relief` (adopting the naming convention from gdaldem).

`color-relief` would have a `paint` option `color-relief-color`, which would use an `interpolate` expression to describe the color ramp between elevation (in meters above sea level) and color.

A new parameter `elevation` would be added to the `interpolate` operator. `elevation` would only be able to be used with the `color-relief-color` property.

For example, 

```
layers: [
    {
        "id": "elevation",
        "source": "terrainSource",
        "type": "color-relief",
        "paint": {
            "color-relief-color": ["interpolate",
                ["linear"],
                ["elevation"],
                0,"black",
                8000,"#FFFFFF"
            ]
        }
    }
]
```

would display the elevation from terrainSource colored from black at 0m above sea level to white at 8000m above sea level.

## API Modifications

No change to the API.

## Migration Plan and Compatibility

No change to existing functionality. To use the new functionality, create a `color-relief` layer.

## Rejected Alternatives

Another option would be to add a paint option `color-relief` to the `hillshade` layer. This would allow a tighter integration between hillshade and hypsometric tint, which are often used in conjunction. This could potentially lead to a more performant implementation.

Alternatively, this functionality could be provided by a new `method` in the existing `hillshade` layer, similar to #1066. Hypsometric tint is different from the other hillshade methods because it uses absolute elevation, as opposed to slope and aspect.

HarelM:
See my comment in the linked maplibre-gl issue.
Regarding performance considerations, I would leave it to be a secondary consideration, the first one should be readability. 
If we can't make it performant enough and still readable we should consider the alternatives, but I'm not sure this will be the case here. 
I tend to say that it's not really hillshading, much like 3D terrain isn't hillshading.
A different layer sounds to me like it might be a better solution.

HarelM:
This proposal implicitly adds `elevation` parameter to `interpolate` operator, right?
[...]

PR #1089, opened 2025-04-18
Color Relief

NathanMOlson:
Implements https://github.com/maplibre/maplibre-style-spec/issues/1067.
[GL JS issue](https://github.com/maplibre/maplibre-gl-js/issues/5666)
[Native issue](https://github.com/maplibre/maplibre-native/issues/3408)

This PR adds a new layer type `color-relief`. `color-relief` has a paint option `color-relief-color`, which uses an interpolate expression to describe the color ramp between elevation (in meters above sea level) and color.

This PR adds a new parameter `elevation` to the `interpolate` operator. `elevation` can only be used with the `color-relief-color` property, as described in the documentation.

image

 - [X] Link to related issues.
 - [X] Include before/after visuals or gifs if this PR includes visual changes.

Documentation looks like this:
cr_docs


HarelM:
LGTM, @louwers can you take a look?

louwers (commented):
Where does it get the elevation data from?

What's sea level?

What's the advantage of this property over raster color-relief images?

The default color shows up weird in the docs.

@sjg-wdw What do you think about this? Basically adds a new layer type with `elevation` expression.

image

HarelM:
This can be achieved in maplibre-gl-js today using a plugin (and a lot of code repetition).
Using the json you won't be able to define a general purpose function (at least I'm having a hard time thinking how this should be defined), so, much like other things, it stays specific. 

BTW, a wild idea, is it possible to define a color method that will draw line that look like contours? i.e. transparent most of the time and  solid color between short elevation when it's "round"?
I know it won't solve the general contour need (as it would need label and other stuff), but might be an easy way to add something to the map that looks similar to contours.

NathanMOlson:
> BTW, a wild idea, is it possible to define a color method that will draw line that look like contours? i.e. transparent most of the time and solid color between short elevation when it's "round"?

Sort of, but it would really want to be a function of zoom. Here's an example using 100m contour spacing from z10 to z14:

z10
z11
z12
z13
z14

NathanMOlson:
Here's a request for a different raster operation: [invert colors](https://github.com/maplibre/maplibre-gl-js/issues/5468)

louwers:
Thanks for your input @sjg-wdw. I agree that this proposal feels too specific, and that we should favor something more generic and extensible. I am not really qualified to come up with a better API though.

NathanMOlson:
The proposed `color-relief` layer takes elevation as input and outputs a color raster. This is the same behavior as the existing `hillshade` layer, so I think any generalization should also encompass hillshade.

GDAL's [`gdaldem`](https://gdal.org/en/stable/programs/gdaldem.html) is a tool that takes elevation as input and outputs a color (or grayscale) raster. This provides a generalization that makes sense to me: it includes hillshade and color-relief, as well as slope, aspect, Terrain Ruggedness Index, Topographic Position Index, and roughness. This is the sort of generalization that I think could work in the style spec.

Maybe something like 

```
layers: [{
  type: 'dem-view',
  algorithm: 'hillshade' | 'color-relief' | 'slope' | 'aspect' | ...
}]
```

GDAL's "shaders" (the processing algorithms for all the different outputs) are all independent and don't share code, so this generalization scheme doesn't really satisfy @sjg-wdw's desire for a reusable shader.

For color-relief, hillshade, and the other `gdaldem` algorithms, I don't think presenting imaging processing steps to the user in the style spec is the right thing to do. That feels like it is pushing implementation details to the user that they shouldn't have to worry about.

HarelM:
I don't think we have plans to break existing styles, so hillshade spec will be kept as is for the foreseeable future.
@sjg-wdw @louwers do you have a concrete proposal how to make this generalized? Because saying it can be generalized and actually nailing generalization are two different things and I want to move this conversation forward.

sjg-wdw:
> GDAL's "shaders" (the processing algorithms for all the different outputs) are all independent and don't share code, so this generalization scheme doesn't really satisfy @sjg-wdw's desire for a reusable shader.

Well that's a separate problem, I suppose.  But if you structure the style the way you're suggesting then it gets easier to do that in the renderer.

sjg-wdw:
> I don't think we have plans to break existing styles, so hillshade spec will be kept as is for the foreseeable future.
> @sjg-wdw @louwers do you have a concrete proposal how to make this generalized? Because saying it can be generalized and actually nailing generalization are two different things and I want to move this conversation forward.

@NathanMOlson's proposal looked good to me.  I might call the type 'elevation' or something similar.  The 'algorithm' keyword looks good and then just add a 'color' option to that.  The algorithm should give us enough information to know what the 'color' is.

As for the other things I suggested, like a convolution filter, no need to add that now.  Let someone come along and implement it.  At least there's a place for it.

And for the image variant, it does suggest a way to set up that as well.  No need to make that part of this spec change, but it suggests the way forward there too.

HarelM:
I think the discussion should be at: #1067.
I tend to favor option 1 to be honest (part of hillshade).

NathanMOlson:
@louwers Are there any changes you want to see here?

src/reference/v8.json
```diff
@@ -4187,6 +4218,26 @@
           }
         }
       },
+      "elevation": {
+        "doc": "Gets the elevation of a pixel (in meters above the vertical datum reference of the `raster-dem` tiles) from a `raster-dem` source. Can only be used in the `color-relief` property.",
```
louwers, line 4222: I think this should be can only be used in the `color-relief-color` property of a `color-relief` layer.

src/reference/v8.json
```diff
@@ -7633,6 +7685,50 @@
...
+        "interpolated": true,
+        "parameters": [
+          "zoom"
+        ]
+      },
+      "property-type": "data-constant"
+    },
+    "color-relief-color": {
+      "type": "color",
+      "doc": "Defines the color of each pixel based on its elevation. Should be an expression that uses `[\"elevation\"]` as input.",
```
louwers, line 7713: Is it an error not to use `elevation`?

Could you provide one or two examples? I think this is more clear then.

louwers (approved):

HarelM:
Can you add a screenshot of the docs to the initial comment of this PR?

src/reference/v8.json
NathanMOlson: Done.

src/reference/v8.json
NathanMOlson: > Is it an error not to use `elevation`?

I haven't added any enforcement mechanisms in the spec, but the GL JS implementation won't work if the value doesn't use elevation...

> Could you provide one or two examples? I think this is more clear then. And helpful for ChatGPT.

Done. I added one example, is there a way to add two? I tried but the JSON validator was unhappy with the duplicate field.

NathanMOlson: Is there a way to enforce this? I modelled it after `heatmap-color`, which has this in its doc string: "Should be an expression that uses `[\"heatmap-density\"]` as input." As far as I can tell the spec does not have an enforcement mechanism for this.

louwers: Looks good, thanks.

PR #1112, opened 2025-05-02
Improve expression syntax documentation

### LINKED ISSUE #724: Step syntax can include colors
The syntax for `step` is

```
["step", step, number, number, number, ...]: number
```

according to the documentation.

However, the step value can be a color too (or anything really?). E.g.

```
["step", ["get", "rank"], 0, "red", 10, "green", 20, "blue", 30, "purple", 40, "yellow"]
```

@HarelM Could you confirm?

https://maplibre.org/maplibre-style-spec/expressions/#step

sruenwg:
Closes #1102
Closes #724

This changes the expression syntax documented on the Expressions page to follow the patterns proposed in #1102.
Some incorrectly documented syntaxes were fixed in the process.

`match` before:
image
`match` after:
image

`in` before:
image
`in` after:
image

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -10,7 +10,7 @@ import jsonStringify from 'json-stringify-pretty-compact';
 const BASE_PATH = 'docs';
 
 type JsonExpressionSyntax = {
-    variants: {
+    overloads: {
         parameters: string[];
         'output-type': string;
     }[];
@@ -140,8 +140,8 @@ function sdkSupportToMarkdown(support: JsonSdkSupport): string {
  */
 function expressionSyntaxToMarkdown(key: string, syntax: JsonExpressionSyntax) {
     let markdown = '\nSyntax:\n';
-    const codeBlockLines = syntax.variants.map((variant) => {
-        return `[${[`"${key}"`, ...variant.parameters].join(', ')}]: ${variant['output-type']}`;
+    const codeBlockLines = syntax.overloads.map((overload) => {
+        return `[${[`"${key}"`, ...overload.parameters].join(', ')}]: ${overload['output-type']}`;
     });
     markdown += `${codeBlockMarkdown(codeBlockLines.join('\n'), 'js')}\n`;
     for (const parameter of syntax.parameters ?? []) {
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -531,7 +531,8 @@ function createExpressionsContent() {
             const parameterLines = (syntax.parameters ?? []).map((param) => {
                 let line = `- \`${param.name}\``;
                 if (param.type) {
-                    line += `: *${param.type}*`;
+                    const type = param.type.replaceAll('<', '&lt;').replaceAll('>', '&gt;');
+                    line += `: *${type}*`;
                 }
                 if (param.description) {
                     line += ` — ${param.description}`;
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -9,6 +9,18 @@ import jsonStringify from 'json-stringify-pretty-compact';
 
 const BASE_PATH = 'docs';
 
+type JsonExpressionSyntax = {
+    variants: {
+        parameters: string[];
+        'output-type': string;
+    }[];
+    parameters?: {
+        name: string;
+        type?: string;
+        description?: string;
+    }[];
+}
+
 type JsonSdkSupport = {
     [info: string]: {
         js?: string;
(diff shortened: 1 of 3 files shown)


build/generate-docs.ts
```diff
@@ -511,9 +511,37 @@ function createExpressionsContent() {
             }
             content += `\n### ${key}\n`;
             content += `${value.doc}\n`;
-            value.example.syntax.method.unshift(`"${key}"`);
-            content += `\nSyntax:\n${codeBlockMarkdown(`[${value.example.syntax.method.join(', ')}]: ${value.example.syntax.result}`, 'js')}\n`;
-            content += `\nExample:\n${codeBlockMarkdown(`"some-property": ${formatJSON(value.example.value)}`)}\n`;
+
+            content += '\nSyntax:\n';
```
HarelM, line 515: Should this be exported to a method to take care of everything related to the syntax part?

src/reference/v8.json
```diff
@@ -2837,13 +2837,32 @@
     "values": {
       "let": {
         "doc": "Binds expressions to named variables, which can then be referenced in the result expression using `[\"var\", \"variable_name\"]`.\n\n - [Visualize population density](https://maplibre.org/maplibre-gl-js/docs/examples/visualize-population-density/)",
-        "example": {
-          "syntax": {
-            "method": ["string", "value", "expression"],
-            "result": "value"
-          },
-          "value": ["let", "someNumber", 500, ["interpolate", ["linear"], ["var", "someNumber"], 274, "#edf8e9", 1551, "#006d2c"]]
+        "syntax": {
```
HarelM, line 2840: I think that the fact that this was added here as a "top level property" requires it to be removed somewhere when "minimizing" this v8 file, can you please check this?

src/reference/v8.json
```diff
@@ -2837,13 +2837,32 @@
...
+          "variants": [
```
HarelM, line 2841: I'm not sure I fully understand the difference between variants and parameters, can you elaborate what each means or rename it to be more clear?

build/generate-docs.ts
```diff
@@ -511,9 +511,37 @@ function createExpressionsContent() {
...
+            const syntax: {
+                variants: {
```
HarelM, line 517: It would be better to define these types elsewhere instead of "inline".

build/generate-docs.ts
```diff
@@ -511,9 +511,37 @@ function createExpressionsContent() {
...
+                    name: string;
+                    type?: string;
+                    description?: string;
+                }[];
+            } = value.syntax;
+            const codeBlockLines = syntax.variants.map((variant) => {
+                return `[${[`"${key}"`, ...variant.parameters].join(', ')}]: ${variant['output-type']}`;
+            });
+            content += `${codeBlockMarkdown(codeBlockLines.join('\n'), 'js')}\n`;
+            const parameterLines = (syntax.parameters ?? []).map((param) => {
```
HarelM, line 531: Consider a simple for look instead of a map call since the conversion is to a line that later on gets added, so a for feels a bit more readable, to me at least.

HarelM:
Thanks for this great effort!
I've added a few comments, nothing major.
Once addressed, I'll merge this.

src/reference/v8.json
sruenwg: I meant for `variants` to list the different syntax variants for the operator, and for `parameters` to describe each parameter which appears within those syntax variants.

For example, there are three ways to form an expression with the 'array' operator: with 1, 2, or 3 parameters.
It makes most sense to me to call these `variants` of the syntax for 'array' expressions, hence `"array".syntax.variants`.
``` jsonc
// 'array' syntax
{
  "variants": [
    {
      "parameters": ["value"],
      "output-type": "array"
    },
    {
      "parameters": ["type", "value"],
      "output-type": "array<type>"
    },
    {
      "parameters": ["type", "length", "value"],
      "output-type": "array<type, length>"
    }
  ],
  "parameters": [
    {
      "name": "value",
      "type": "any"
    },
    {
      "name": "type",
      "type": "\"string\" | \"number\" | \"boolean\"",
      "description": "The asserted type of the input array."
    },
    {
      "name": "length",
      "type": "number literal",
      "description": "The asserted length of the input array."
    }
  ]
}
```


`syntax.parameters` is separate from `syntax.variants` since the variants tend to share parameters (e.g. 'value' is shared across all the variants above, and 'type' is shared between the latter two), and I'd like to only define these parameters once.

I personally feel these names are pretty consistent with the intended usage, but please let me know if there are different names you would prefer.

src/reference/v8.json
sruenwg: If this is referring to `minifyStyleSpec` used in the Rollup config, I believe nothing has to be changed since that function [already removes the entire `expression_name` object](https://github.com/maplibre/maplibre-style-spec/blob/b4d89780ac588e6d34d88ffac903d6c792a8fcd4/build/rollup_plugin_minify_style_spec.ts#L17) (which contains all the parts I changed) from the v8 JSON.

sruenwg:
Thanks for the comments @HarelM!
Please check the updated version + my comments whenever you are able to, thanks again 🙇

src/reference/v8.json
HarelM: Maybe "overloads"? variants wasn't intuitive for me... But if overloads isn't intuitive for you then variants is ok...

HarelM:
Just to be on the safe side, although this doesn't change the logic, let's add a changelog entry if anyone is relaying on the current v8 content.

src/reference/v8.json
sruenwg: Ah 'overloads' definitely works with me too. Will rename it 👍

PR #1128, opened 2025-05-13
Improve expression types

sruenwg:
Fixes/improves typings for expressions.
Added typechecked unit tests corresponding to each change made to the type definition. (Tests are not exhaustive for each operator — I mostly added tests only for behaviour that is relevant to the change in typing.)
Replaced the `compileTimeCheck`s in `feature_filter.test.ts` with typecheck tests in `expression.test.ts` using Vitest's `expectTypeOf`.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


sruenwg:
Some questions I have before this is ready for a full review (mainly for @HarelM):

1. I'd like to hear your thoughts on the pattern of typechecked unit tests I've used here, and whether you see it as an improvement over `compileTimeCheck`.
To test the types, I believe we want to be able to easily compare the result of the TS typecheck against the result of the style spec parsing/evaluation — the `compileTimeCheck` pattern does not evaluate the expression and hence makes this difficult.
I've instead made it so `createExpression` in `expression.test.ts` performs type-checking so that the typecheck and parsing/evaluation occur in one place:
   - an expression's unit test that returns an error would often be marked with `// @ts-expect-error` to denote that TS would successfully reject that input
   - an expression's unit test that successfully parses and evaluates would almost never be marked with `// @ts-expect-error`

   I also prefer this pattern because:
   - an expression's set of unit tests usually covers many edge cases, and these would all help to verify the type's correctness
   - unit test authors would encounter TS errors for new features, hopefully encouraging more up-to-date types

2. Should the types strictly be more permissive than the parser/evaluator implementation?
This was briefly mentioned in https://github.com/maplibre/maplibre-gl-js/pull/1390 and while I agree for the most part, it does create situations like allowing `['let', 'a', 0, 'b', 1]`, `['concat']`, or `['+']` which may not be ideal.
The types for variadic expressions like these are currently all over the place, so before updating these types I'd like to confirm whether this is indeed the pattern we wish to follow.

HarelM:
The "hardcoded" types code exists since there wasn't an easy solution to generate the expression types from the spec.
If this is no longer the case, there's no need for the tests if the expression generation is automatic I believe.
I might be wrong though, as those tests would be needed to make sure the generated or hand written typescript is correct.
But we don't test that the rest of the generated code is actually generated properly, we did this for the expression since we wanted to make sure the code we write is properly defining existing expressions.

Regarding `// @ts-expect-error` - does it fail in CI in case of a change in the types? I don't mind removing `compileTimeCheck` if there's a typescript built in way to check it.
The problem I see for this is that it's not clear that this is part of a test but looks like someone didn't want to fix the types and added this comment, so I'm not sure I'm totally on board with this approach...

sruenwg:
Sorry, I don't think I quite get what you were trying to say in the first paragraph.

> If this is no longer the case, there's no need for the tests if the expression generation is automatic

To the best of my knowledge, there's still no easy way to generate the types from the implementation (hence the type updates in this PR).

> we did this for the expression since we wanted to make sure the code we write is properly defining existing expressions

I definitely agree the expression types should be tested especially since they are handwritten (hence I am hoping there could be a way to test them more extensively).

---

Yes, `// @ts-expect-error` expects an error, so `tsc` fails when the type definitions are changed such that the subsequent line no longer actually produces a TS error.
I find this feature perfect for our situation since the [TS release notes for `// @ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) describes an example almost identical to what we are facing.

> it's not clear that this is part of a test but looks like someone didn't want to fix the types

While I don't agree it's unclear, I do agree this introduces the risk of people adding these comments out of laziness.
Three options come to mind to address this:
1. Simply use the old pattern. I will copy all the test expressions added in this PR over to `compileTimeCheck` for typechecking the valid expressions.
2. Split `createExpression` into `createValidExpression(expression: ExpressionSpecification, ...)` and `createInvalidExpression(expression: any, ...)` in the test file. Since we care less about whether the type manages to reject invalid expressions, using an untyped `createInvalidExpression` for invalid expressions means there would be no need for `// @ts-expect-error`.
3. Split `createExpression` into `createValidExpression` and `createInvalidExpression`, but make both functions typed and write a custom lint rule to disallow `// @ts-expect-error` on `createValidExpression`s (though even without the lint rule I feel most people would hesitate to put `@ts-expect-error` on a `createValidExpression`).

Please let me know your thoughts and any additional suggestions you might have.
For reference, my preference would be 3 > 2 > 1 as I see value in documenting what instances the type succeeds in rejecting.

HarelM:
Thanks for the link related to `// @ts-expect-error`.
Since I'll be the one reviewing this repo for the near future, I can watch out for incorrect typescript comments.
Regarding 1 2 or 3, then I guess we can switch to using `// @ts-expect-error` for the tests, and also for other places.
I would recommend keeping only a "valid" create expression method so that we know the types are checked always and we can use `ts-ignore` if we can't properly define it in typescript or `ts-expect-error` if the expression is incorrect since this is what the test is checking.

What I meant to ask is, given the new syntax v8 definition you added, can we generate the types automatically from the v8 file, and if not, what can we do to automate maybe a big part of it and keep the rest hardcoded or something?

Otherwise, I'm good with the content of this PR.

HarelM:
I'm planning on releasing a version in a day or two, any chance to push this through?

sruenwg:
Sorry, hadn't had time to work on this since.

> given the new syntax v8 definition you added, can we generate the types automatically from the v8 file

That does sound like a nicer way to maintain the expression types…
Assuming no one is really waiting on this PR, I'd rather explore that approach (generating types from v8) a bit than rushing to get this current PR merged. Is that alright with you?

HarelM:
Sure, no problem.

sruenwg:
Had some time to think about this… I believe the only hard/messy part about generating the expression types would be to figure out a clean way to translate the more complex parameter types into TS.

For the simpler parameter types, the simplest approach is probably to just define a mapping from the v8 types to TS types:
```ts
const v8TypeToTsType = {
  'any': 'null | boolean | number | string | ExpressionSpecification',
  'array': 'ExpressionSpecification',
  'number': 'number | ExpressionSpecification',
  'string': 'string | ExpressionSpecification',
  'number literal': 'number',
  'string literal': 'string',
  ...,
};
```
We *could* then extend this to also include the more complex types, but this certainly does not feel ideal:
```ts
const v8TypeToTsType = {
  'number | array<number> | color | array<color> | projection': 'ExpressionSpecification | number | (number | ExpressionSpecification)[] | ColorSpecification | (ColorSpecification | ExpressionSpecification)[] | ProjectionDefinitionSpecification',
  ...,
};
```
I mostly feel uncomfortable about the fact that these complex types are often one-off types specific to a particular expression, so I think I would prefer it if we hardcoded the complex types at the parameter level in v8.json, if that makes sense. The idea I have is to change `parameter.type` to `parameter['docs-type']` and create an optional `parameter['ts-type']` field which would hold the raw TS type string for complex types.

So to answer this:
> what can we do to automate maybe a big part of it and keep the rest hardcoded or something?

The overall structure of the expression (parameter positions, arity, optionality) should be able to be derived from `syntax.overloads`, and the TS type for simple parameters could be determined via `v8TypeToTsType[parameter['docs-type']]`.
For complex types, until a better solution is suggested I think it would be reasonable to hardcode the `ts-type` and construct the TS type via `parameter['ts-type'] ?? v8TypeToTsType[parameter['docs-type']]`.

Please let me know if this is/isn't close to what you had in mind.

---

For reference, some other ideas which might be cleaner but felt like overkill:
- Defining the types in v8 with a JSON Schema-like syntax and parsing that to build both the type strings for the docs and the TS types.
  This would make v8 even more verbose and hard to understand for devs unfamiliar with JSON Schema while also introducing complex (brittle) parser logic.
- Using string interpolation to mark non-TS portions of the type for replacement, e.g. `{ "type": "{{number}} | {{array<{{number}}>}} | {{color}} | {{array<{{color}}>}} | {{projection}}" }`
  Again, the parser would introduce more complex code to maintain for little benefit in return.

HarelM:
I think your direction is good, and balances readability and complexity.
I also think that you can compare the output to existing hardcoded solution is if it makes sense and does the job well. and then decide if this is better or worse than current solution.

Regarding json schema stuff, I thought in the past to convert the entire v8 to json schema, and generate a parser, validator and typescript types from it.
I'm not sure it's possible given the complexity of the expressions definitions, but I might be wrong, it's definitely not an easy task, even with AI tools.
Generally speaking, the fact that we defined a language of our own in this v8 file is a bad use of everyone's time, but this is what we currently have, so we can always decide if we want to continue the current solution or rewrite that.
I'm good with either direction, continuing with current solution is always faster in the short term...

Bottom line, carry on!

sruenwg:
Thanks so much for the guidance and feedback.

> I also think that you can compare the output to existing hardcoded solution is if it makes sense and does the job well

Will definitely do.

---

Regarding JSON Schema:

> the fact that we defined a language of our own in this v8 file is a bad use of everyone's time

I generally agree with this and would prefer relying on existing strategies/formats like JSON Schema when possible.

> I thought in the past to convert the entire v8 to json schema

I'm not too familiar with the v8 properties aside from `expression_name`, but from what I could gather, many portions of v8 already sort of resemble JSON Schema but with many additional properties.
Some issues that come to mind if we wish to convert everything to JSON Schema:
- The style spec types are not JSON types.
  This can be dealt with by modifying the standard metaschema.
- Many fields have no JSON Schema equivalents, e.g. `sdk-support`, `property-type`, `requires`.
  Again, these can be supported by extending the standard metaschema.
- We have an `enum` type documented alongside types like `color` and `padding`, so I assume we want to use the `"type": "enum"` syntax similar to those other types. However, this is similar to the standard `"enum": [...literal values...]` keyword, which we cannot use since the standard `enum` does not allow for descriptions on each enum value. We *could* use `"oneOf": [{ "const": "a", "description": "..." }, ...]`, but this might make it unclear that we wish to document this property's type as `enum`.
- For expressions, afaik the main issue lies in JSON Schema not providing a way to define:
  - repeated array item pairs, e.g. the [stop input, stop output] pairs in `["interpolate", interpolation_type, input, stop_1_input, stop_1_output, ..., stop_n_input, stop_n_output]` (see https://github.com/json-schema-org/json-schema-spec/issues/1323), or
  - a parameter after a set of variable-length parameters, e.g. the fallback in `["match", input, label_1, output_1, ..., label_n, output_n, fallback]`.

Depending on how we tackle the above issues, I suspect not much of v8 may actually change (aside from expressions) even if we switch over what parts we can to a more JSON Schema-like structure.
Personally, perhaps what seems more valuable is simply having a JSON Schema which documents and validates the structure of v8.

> and generate a parser, validator and typescript types from it

Can you elaborate what you meant here by generating a parser and validator from the v8 file? Do you just mean writing parser logic (to generate the docs and TS types) and a validator (i.e. a JSON Schema to validate v8.json)?
There's a chance I'm completely misunderstanding the design you have in mind…

sruenwg:
Also, I noticed the repo recently switched to using Vitest, and it seems Vitest supports a unit testing-like setup for testing types: https://vitest.dev/guide/testing-types.html
Wondering whether you agree this would be better than any of the options we'd considered.

> The problem I see for this is that it's not clear that this is part of a test

Relying on Vitest's `expectTypeOf` would certainly solve this ↑ issue.

> we want to be able to easily compare the result of the TS typecheck against the result of the style spec parsing/evaluation

This ↑ was my main motivation for the `// @ts-expect-error` idea, but I believe this comparison would be easy enough with Vitest if we simply placed the `expectTypeOf` test close to the corresponding `expect` test.

Everything seemed to work as expected when trying some basic tests locally.
The docs recommend using `test-d.ts` files for type testing, but I believe placing the type tests alongside regular tests in `test.ts` files should also be fine if we don't care too much about speed:
image

Here is the basic idea:
https://github.com/sruenwg/maplibre-style-spec/commit/425ae1ce4a56111d8a2fed53ef92ec1c32472f4b

HarelM:
The vitest type tests looks good, how do you test something that is invalid?
In any case, feel free to use those instead.

HarelM:
What I meant to say around json schema is that instead of the v8.json we will have a json schema file that defines the spec.
This new json scheme file(s) can be used with a parser/validator to make sure the a style.json is following the json schema defintitions (which are basically the style spec), and from that schema file(s) we will be able to generate typescript types, that are basically similar to what we do using the v8 file now, but as far as I know this is already a "solved" problem - meaning there's already a tool to convert from json schema definitions to typescript types.
What we did here is basically define a language/schema using v8.json file, this language/schema is then parsed into typescript definitions and we wrote (inherited) a validator as well, all of those are custom code which we maintain.
In theory, a json schema should be able to facilitate all of the above - validation and types, but I'm not sure how easy it would be to do that, if at all is possible, and from what I tried for a short period of time it's not easy, especially this part, with the definitions of the expressions.
So, I'm not sure it's worth the investigation unless you want to research it because it's a challenge :-).

We can continue with the effort here regardless of if and when a json schema will replace this v8 and the code around it.
I hope it's clearer now, if not please let me know what's not clear.

sruenwg:
That makes sense, thank you for the clarification.

I don't think I'll pursue the idea here since I'm fairly certain the expressions simply can't be expressed with standard JSON Schema due to the problems I mentioned several messages ago, so at minimum the expressions would require custom code anyway (which the point of the JSON Schema approach would be to try to avoid) for both style validation and TS type generation.
I'll keep the idea in the back of my head though and perhaps experiment a bit in my free time…

For this current PR, now that I better understand the desire to minimize custom code, I'd like to hold off on introducing even more custom code via `v8TypeToTsType` and keep using hardcoded types for now, if that is alright.

---

Please let me know if this PR look fine to merge.
To sum up the changes:
- Fixed/improved expression types to better match the underlying implementation
- Added typecheck tests using Vitest's `expectTypeOf` to verify that the above changes indeed bring the types closer to the implementation
- Replaced the `compileTimeCheck`s in `feature_filter.test.ts` with typecheck tests in `expression.test.ts`

src/expression/expression.test.ts
```diff
@@ -2075,6 +2076,7 @@ describe('interpolation expressions', () => {
             expect(response.value.evaluate({zoom: 10}).values).toEqual([Color.parse('black'), Color.parse('white')]);
         });
         test('type accepts expression which interpolates between non-literal color array outputs', () => {
+            // eslint-disable-next-line
```
sruenwg, line 2079: These comments are here since ESLint warns that the `obj` const is only used as a type, however this is exactly what we want to test since this `const obj = ...` syntax mimics how users may group arguments together in a variable.
```
// We want to test the type against usages like this, where TS would infer obj as having type {'colors-8': string[]; 'colors-10': string[]}
const obj = {'colors-8': ['white', 'black'], 'colors-10': ['black', 'white']};

// Something like this would stop ESLint from complaining, but this is an exact type (not widened by TS) and hence is not what we want to test
type Obj = {'colors-8': ['white', 'black'], 'colors-10': ['black', 'white']};
```

build/generate-style-spec.ts
```diff
@@ -157,18 +157,18 @@ export type CollatorExpressionSpecification =
 
 export type InterpolationSpecification =
     | ['linear']
-    | ['exponential', number | ExpressionSpecification]
-    | ['cubic-bezier', number | ExpressionSpecification, number | ExpressionSpecification, number | ExpressionSpecification, number | ExpressionSpecification]
+    | ['exponential', number]
```
HarelM, line 160: Does this mean exponential and the others don't support data driven expressions or am I missing anything?

sruenwg: Yes, I believe that's how the validation currently works.
The implementation is here:
https://github.com/maplibre/maplibre-style-spec/blob/4370302a42e247093f01f98a9cf6a656fd242728/src/expression/definitions/interpolate.ts#L75-L78
and the test I added to verify it expects a literal:
https://github.com/sruenwg/maplibre-style-spec/blob/cb2e6a96ef6f6aa0a41b70481dc006dd5c2041b6/src/expression/expression.test.ts#L1674-L1681

HarelM: Ok, cool, we can always change it back I guess...

src/expression/expression.test.ts
```diff
@@ -403,27 +428,545 @@ describe('Distance expression', () => {
...
+        expectTypeOf<['array', [1, 2, 3]]>().not.toExtend<ExpressionSpecification>();
+        expectTypeOf<['array', 'number', [1, 2, 3]]>().not.toExtend<ExpressionSpecification>();
+        expectTypeOf<['array', 'number', 3, [1, 2, 3]]>().not.toExtend<ExpressionSpecification>();
+
+        expectTypeOf<['array', ['literal', []]]>().toExtend<ExpressionSpecification>();
+        expectTypeOf<['array', 'number', ['literal', [1, 2, 3]]]>().toExtend<ExpressionSpecification>();
+        expectTypeOf<['array', 'number', typeof Number.MAX_SAFE_INTEGER, ['get', 'arr']]>().toExtend<ExpressionSpecification>();
+    });
+    test('requires either "string", "number", or "boolean" as the asserted type', () => {
+        const invalidResponses = [
```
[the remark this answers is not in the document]
sruenwg: Sure, I'm okay with either approach so I'll break these down into individual tests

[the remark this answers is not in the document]
HarelM: Great. thanks!

build/generate-style-spec.ts
```diff
@@ -178,47 +178,48 @@ export type ExpressionSpecification =
...
-        ExpressionInputType | ExpressionInputType[], ExpressionInputType | ExpressionSpecification,
-        ...(ExpressionInputType | ExpressionInputType[] | ExpressionSpecification)[], // repeated as above
-        ExpressionInputType | ExpressionSpecification]
-    | ['within', unknown | ExpressionSpecification]
-    | ['distance', unknown | ExpressionSpecification]
+    | ['match', string | number | ExpressionSpecification,
+        string | number | string[] | number[], null | ExpressionInputType | ExpressionSpecification,
+        ...(string | number | string[] | number[] | null | ExpressionInputType | ExpressionSpecification)[], // repeated as above
+        null | ExpressionInputType | ExpressionSpecification]
+    | ['within', GeoJSON.GeoJSON]
```
melitele, line 215:
```
[!] (plugin typescript) RollupError: [plugin typescript] src/source/tile.test.ts (51:86): @rollup/plugin-typescript TS2322: Type '{ type: string; coordinates: number[][][]; }' is not assignable to type 'GeoJSON<Geometry, { [name: string]: any; }>'.
  Type '{ type: string; coordinates: number[][][]; }' is not assignable to type 'MultiPolygon'.
    Types of property 'type' are incompatible.
      Type 'string' is not assignable to type '"MultiPolygon"'.
src/source/tile.test.ts (51:86)

51         tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['within', polygon]});
                                                                                        ~~~~~~~
```

sruenwg: I believe the options here are either to:
1. get the user to type the variable as a `GeoJSON.GeoJSON` (more work on user side, more informative types), or to
2. loosen the typing (less work on user side, less informative types).

In this case I think I prefer option 1 since:
- any alternative to `GeoJSON.GeoJSON` feels too uninformative and non-self-documenting (e.g. `unknown`), and
- I assume that in TS contexts this variable would more often be typed as `any` (when pulled from some untyped external data) or `GeoJSON.GeoJSON` (when pulled from some data structure containing a GeoJSON) rather than written out by hand like in the test you ran.

Please let me know your thoughts.
The `'ExpressionSpecification type supports common variable insertion patterns'` test in `expression.test-d.ts` shows how this issue was dealt with elsewhere, in case you were curious. We should probably add this case there once we figure out how we want the typing to behave.

melitele: Thank you for the explanation. The type is fine. My point is that this is a breaking change (albeit trivial to fix per option 1) and it isn't clear from the PR description that that is intentional. The _user_ here is `maplibre-gl-js` module.

PR #1158, opened 2025-06-06
Convert Vitest expression unit tests to JSON test fixtures

sruenwg:
Converts the expression tests in `src/expression/expression.test.ts` to JSON test fixtures in `test/integration/expression` where the majority of expression tests already live.

This PR is meant to address [this comment](https://github.com/maplibre/maplibre-style-spec/pull/1128#issuecomment-2915370037) regarding expression functionality being tested via two patterns and often duplicating tests.

I refactored `test/integration/expression.test.ts` to aid myself in understanding the file, but please let me know if you would like this reverted.
No functionality was changed.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


sruenwg:
No worries. In the meantime I'll look into what tests I missed to cause the drop in coverage…

test/integration/expression/expression.test.ts
```diff
@@ -7,180 +7,220 @@ import {
...
-            const evalOk = compileOk && deepEqual(result.outputs, expected.outputs, DECIMAL_SIGNIFICANT_FIGURES);
-            
             try {
                 expect(compileOk).toBeTruthy();
             } catch {
                 throw new Error(`Compilation Failed:\nExpected ${JSON.stringify(expected.compiled)}\nResult   ${JSON.stringify(result.compiled)}`);
             }
-            
+
+            const resultOutputs = (result as any).outputs;
```
HarelM, line 112: Can we improve this to not use "as any"?

test/integration/expression/expression.test.ts
```diff
@@ -7,180 +7,220 @@ import {
...
+                }
 
-                const fname = path.join(__dirname, expressionTestFileName);
-                fs.writeFileSync(fname, JSON.stringify(fixture, null, 2));
+                fs.writeFileSync(fixturePath, JSON.stringify(fixture, null, 2));
                 return;
             }
 
-            const expected = fixture.expected;
+            const expected = fixture.expected as FixtureResult;
```
HarelM, line 103:
```suggestion
            const expected: FixtureResult = fixture.expected;
```
Would this work?

test/integration/expression/tests/slice/string-two-indexes/test.json
HarelM, line 1: Why was this file changed?

HarelM:
I think this PR is a bit big and I think I'm getting lost here reviewing it...
I would advise to do the following:
1. A separate PR to refactor `test/integration/expression/expression.test.ts` - as this is only type changes etc, right?
2. A PR to move tests from typescript to json (i.e should only have added json files and reduce tests in the expressions.tests.ts file that has some expression tests)
3. A PR to change the content of some of the current tests as right now it's not clear why these tests were changed.

Sorry for the extra work, but this will help me focus my questions according to the relevant content of the PR.
Thanks for all the hard work around this!

test/integration/expression/expression.test.ts
sruenwg: I recall settling on this after considering:
1. `(result as FixtureSuccessResult).outputs`: This felt misleading since `result` is actually `FixtureSuccessResult | FixtureErrorResult` (and the existing code looked like it also wanted to handle unexpected `outputs` existing on `FixtureErrorResult`s)
2. `result['outputs']`: This felt like a bit of a silent hack since this compiles whereas `result.outputs` doesn't

though in hindsight I see that `as any` is hacky in its own way. Would you prefer `result['outputs']` here?

test/integration/expression/expression.test.ts
sruenwg: Unfortunately no.
image

The issue here is that `fixture.expected` is `FixtureResult | undefined`, so I did consider `const expected = fixture.expected!` but settled on `as FixtureResult` since I couldn't find the `!` pattern used anywhere else and it seemed we were fairly lax on using `as` assertions.

test/integration/expression/tests/slice/string-two-indexes/test.json
sruenwg: Ah sorry, I tweaked some existing tests so there's some consistency among similar tests within the same test folder, if that makes sense.
So for the `slice` expression tests, `array-one-index`, `array-two-indexes`, `string-one-index`, `string-two-indexes` are now aligned and contain the same/analogous set of tests.
This was to make it easier for me to check that each file was tested with the same set of input cases, and I figured future maintainers would also have an easier time following the tests if they were organized this way.

test/integration/expression/expression.test.ts
HarelM: Fair enough.

test/integration/expression/expression.test.ts
HarelM: Would a union type of `FixtureSuccessResult | FixtureErrorResult` would help here?

sruenwg:
Apologies, I definitely could have organized my commits better to make reviewing easier.

> I would advise to do the following:
> 
> 1. A separate PR to refactor `test/integration/expression/expression.test.ts` - as this is only type changes etc, right?
> 2. A PR to move tests from typescript to json (i.e should only have added json files and reduce tests in the expressions.tests.ts file that has some expression tests)
> 3. A PR to change the content of some of the current tests as right now it's not clear why these tests were changed.

Might take me a while but I can definitely do that.
I suspect PR 3 might get messy regardless but we'll see… I will at least make sure to document the rationale behind each refactor in bite-size commits so the PR is easier to follow.

test/integration/expression/expression.test.ts
sruenwg: No, TS would complain that `FixtureErrorResult` has no `outputs` property.
(`result` here has type `FixtureResult = FixtureSucessResult | FixtureErrorResult`, so this would be the same as giving no type annotation at all)

sruenwg:
Closing this PR in favour of separate PRs as described above.

PR #1172, opened 2025-06-11
Improve typings in expression integration test file

sruenwg:
Improves typings in `expression.test.ts` along with some minor refactoring (mostly to aid with type inference).
These changes were originally a part of https://github.com/maplibre/maplibre-style-spec/pull/1158 but were moved to this separate PR as suggested in https://github.com/maplibre/maplibre-style-spec/pull/1158#issuecomment-2953737315.

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/test/integration/expression/expression.test.ts
+++ b/test/integration/expression/expression.test.ts
@@ -181,8 +181,9 @@ function getCompilationSuccessResult(expression: StylePropertyExpression): Compi
 
 function evaluateExpression(inputs: FixtureInput[], expression: StylePropertyExpression): EvaluationOutput[] {
     const type = getStylePropertyExpressionType(expression);
+    const outputs: EvaluationOutput[] = [];
 
-    return inputs.map((input) => {
+    for (const input of inputs) {
         const {availableImages, canonicalID} = input[0];
         const {featureState, geometry, id, properties} = input[1];
 
@@ -212,15 +213,16 @@ function evaluateExpression(inputs: FixtureInput[], expression: StylePropertyExp
             if (type.kind === 'color') {
                 value = [value.r, value.g, value.b, value.a];
             }
-            return value;
+            outputs.push(value);
         } catch (error) {
-            return {
+            outputs.push({
                 error: error.name === 'ExpressionEvaluationError' ?
                     error.toJSON() :
                     error.message,
-            };
+            });
         }
-    });
+    }
+    return outputs;
 }
 
 function getStylePropertyExpressionType(expression: StylePropertyExpression) {
--- a/test/integration/expression/expression.test.ts
+++ b/test/integration/expression/expression.test.ts
@@ -7,31 +7,71 @@ import {
     convertFunction,
     toString,
     ICanonicalTileID,
-    StylePropertyExpression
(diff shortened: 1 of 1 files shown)


test/integration/expression/expression.test.ts
```diff
@@ -7,180 +7,220 @@ import {
...
+    };
+}
 
-    const evaluationResult: any[] = [];
+function evaluateExpression(inputs: FixtureInput[], expression: StylePropertyExpression) {
+    const type = getStylePropertyExpressionType(expression);
 
-    const expressionValue = expression.value;
-    const type = (expressionValue as any)._styleExpression.expression.type; // :scream:
+    return inputs.map((input) => {
```
HarelM: I think it's better to have a for loop here instead as this is what actually runs the tests, and a `map` here is confusing (to me)...

test/integration/expression/expression.test.ts
```diff
@@ -7,180 +7,220 @@ import {
...
+    return {
+        result: 'success',
+        isFeatureConstant: kind === 'constant' || kind ==='camera',
+        isZoomConstant: kind === 'constant' || kind === 'source',
+        type: toString(type),
+    };
+}
 
-    const evaluationResult: any[] = [];
+function evaluateExpression(inputs: FixtureInput[], expression: StylePropertyExpression) {
```
HarelM: Better to define the return value of this method, as it's a bit long and hard to understand just by looking at it.

HarelM:
I've added some nit-picking, not something dramatic...

test/integration/expression/expression.test.ts
sruenwg: I think I disagree.
This function takes an array of inputs and performs a 1-to-1 transformation (a mapping if you will) to produce the output array, with no side effects. To me, a `map` communicates this idea well and `for` does not.

> Better to define the return value of this method, as it's a bit long and hard to understand just by looking at it.

I agree this could be helpful, so let's say this were added.
Then, when reading this part of the code in the future:
- with `map`, I could look at the type annotation and `map` on the second line to quickly determine with reasonable confidence that this function simply performs a pure 1-to-1 mapping, allowing me to skip over the rest of the function;
- with `for`, I tend to anticipate some irregular control flow (maybe the input-output arities are different, or the function throws in certain cases) or some other side effect (maybe some external variable is updated), so I would have to read through the entire `for` body to be confident I understand the function.

Please let me know how `map` is confusing here or how `for` is advantageous here because I might be missing something…

HarelM: I guess it's a matter of taste/what we are used to...
The confusion from my point of view is that I'm usually treating map method as simple conversion, this map method has try-catch and other complicated logic which is not what I would expect to see in simple conversion logic.
It is the core logic of the test harness, so placing it inside a map method seems odd, to me. 
Having said that, I don't have strong feelings either way...

sruenwg: Okay, I do see that `map` seems to currently only be used for relatively simpler transformations, so I'll switch to `for` to keep in line with that.

HarelM (approved):

PR #1184, opened 2025-06-21
Add fonts property to style spec

louwers:
Implement @alanchenboy's design proposal at https://github.com/maplibre/maplibre-style-spec/issues/868.

We need this property to be able to reliably load fonts on with MapLibre Native. Loading font files as opposed to glyphs is required for rendering complex scripts (with Harfbuzz). Relying on fonts being available on the platform is not a viable strategy for MapLibre Native, and I really want to give users the ability to use a style that depends on complex scripts without external platform-specific configuration and distribution. See earlier discussion over at: https://github.com/maplibre/maplibre-style-spec/issues/1045

image


src/index.test.ts
```diff
@@ -144,6 +145,7 @@ function validSchema(k, v, obj, ref, version, kind) {
             } else if (typeof obj.value === 'object') {
                 validSchema(`${k}.value`, v, obj.value, ref, undefined, undefined);
             } else {
+                console.log(types);
```
louwers:
```suggestion
```

louwers:
```suggestion
```

src/reference/v8.json
```diff
@@ -218,6 +218,50 @@
         }
       }
     },
+    "fonts": {
+      "type": "array",
+      "value": "object",
+      "doc": "The `fonts` property can be used to specify what fonts to use for particular unicode ranges. Traditionally MapLibre styles only supported [glyphs](./glyphs.md), which do not contain the information needed to render complex texts such as [Devanagari](https://en.wikipedia.org/wiki/Devanagari), [Khmer](https://en.wikipedia.org/wiki/Khmer_script) among many others.\n\nThe `type` specifies the type of the font corresponding to the url. What types are supported is implementation-defined. Types that are not supported on a platform are skipped. When no (supported) font is specified for a particular unicode code point or when a font cannot be loaded, fall back to using the `glyphs` property (if present).\n\nThe `unicode-ranges` property should an array of one or more arrays of size two, each indicating a start and end of a unicode range. Numeric values can be represented as number or a string containing a four digit hexidecimal number.",
```
HarelM:
```suggestion
      "doc": "The `fonts` property can be used to specify what fonts to use for particular unicode ranges. This is different than [glyphs](./glyphs.md) as it contains more information needed to render complex texts such as [Devanagari](https://en.wikipedia.org/wiki/Devanagari), [Khmer](https://en.wikipedia.org/wiki/Khmer_script) among many others.\n\nThe `type` specifies the type of the font corresponding to the url. What types are supported is implementation-defined. Types that are not supported on a platform are skipped. When no (supported) font is specified for a particular unicode code point or when a font cannot be loaded, it will fall back to using the `glyphs` property (if present).\n\nThe `unicode-ranges` property should be an array of one or more arrays of size two, each indicating a start and end of a unicode range. Numeric values can be represented as number or a string containing a four digit hexidecimal number.",
```

alanchenboy:
Because the shaping information was in the font file, so the engine needs an API to download the font file to local. Anyway, if anyone has a solution to support download a font file, I can edit the PR https://github.com/maplibre/maplibre-style-spec/issues/868 to adopt that.

alanchenboy:
This proposal is aligned with @wipfli. The original design is that there is a hard-coded Unicode point range to define which Unicode points should undergo the shaping process. So must define a language define like
`
"font-faces": [
      {
          "text-font": ["Noto Sans Regular"],
          "font-files": [
              {
                  "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
                  "language": "khmer"
              },
              {
                  "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
                  "language": "myanmar"
              },
              {
                  "url":"https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
                   "language": "devanagari"
              }
              
          ]
      }
  ],
`

But this design, the backend must also understand the language things.

I think the origin pipeline is ok for simple text, if no unicode-range or language, then it means all unicode points must undergo shaping, and it's not make sense.

louwers:
I like that idea. 

I will adjust the spec to be more in line with CSS.

louwers:
Based on feedback from today's MapLibre Eastern Call:

- We keep `unicode-range` but it is made optional. Grab uses a common style for many countries with different scripts. I presume users are usually zoomed in which means only one complex script is being rendered. Having to download all defined fonts up front  would be prohibitively expensive for their use case.
- I have also added some documentation on where to find common unicode ranges, the similar CSS property and a tool called FontDrop to inspect what unicode codepoints exist in a font.
- We will not require specifying the type of font. This is also not a requirement in CSS. If needed we can always add it as an option later.

In addition:

- I like the suggestion from @ianthetechie to use arrays for the unicode ranges, but I think we should use strings to have a design inspired by CSS.
- `unicode-range` kept singular similar to CSS.

mwilsnd (approved):
Replicating CSS, offering an optional range strikes a reasonable balance I feel. Following a more ubiquitous specification should make the feature a bit more approachable. While not supporting the segmentation capability would make the specification more simple, the cost of processing entire fonts isn't worth it.

birkskyum (approved):

src/reference/v8.json
```diff
@@ -218,6 +218,48 @@
         }
       }
     },
+    "font-faces": {
+      "type": "array",
+      "value": "object",
```
HarelM, line 223: Can't we better define this? Can this be in any given format?

HarelM: I would also argue that the this needs more information on each field in the "font face object" to explain what are the valid values, for example why the `text-font` files is an array of strings and not just a string, what type of files are supported in the url, and why `unicode-range` is an array.
The more info we provide, the useful it will be for our users.

louwers: > Can't we better define this? Can this be in any given format?

I think a precise `value` is only used by MapLibre GL JS. I consider it to be part of the TypeScript implementation.

> What are the valid values, for example why the `text-font` files is an array of strings and not just a string

This means that, whenever a symbol layer with a particular `text-font`, for example:

```
"text-font": ["Open Sans Regular","Arial Unicode MS Regular"]
```

It will look for font-faces where "Open Sans Regular" is in the `text-font` array, then for font-faces where "Arial Unicode MS Regular" is in the `text-font` array.

Allowing multiple fonts to be specified is for parity with CSS, where multiple `font-family` can be specified for a particular `@font-face`.

> what type of files are supported in the url

I can update the support matrix with font types. The initial MapLibre Native implementation will only support TrueType fonts.

> why unicode-range is an array.

This is because you might be interested in a non-sequential set of unicode code points in a particular font.

@alanchenboy Is my interpretation correct? Can you help me improve the description below?

> The `font-faces` property can be used to specify what fonts to use. These font faces take precedence over [glyphs](./glyphs.md). Font faces contain information needed to render complex texts such as [Devanagari](https://en.wikipedia.org/wiki/Devanagari), [Khmer](https://en.wikipedia.org/wiki/Khmer_script) among many others.
> 
> When an optional `unicode-range` is specified the font is only used for characters within that unicode range. When specified it should be an array of strings, each indicating a start and end of a unicode range, similar to the [CSS descriptor with the same name](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range). Refer to the [Unicode Character Code Charts](https://www.unicode.org/charts/) to see ranges for scripts supported by Unicode. To see what unicode code points are available in a font, use a tool like [FontDrop](https://fontdrop.info/).

HarelM: > I think a precise value is only used by MapLibre GL JS. I consider it to be part of the TypeScript implementation.

The v8 file defines types, which is later on translated to typescript, but other types are defined in this file which later on gets published as the spec docs, and I think the font object should be defined well and documented as such, like other types in this v8 file. 
Let me know if you need help with that.

louwers: Probably can use some help there, it looks pretty manual for custom objects like here.

HarelM: I'll see if I can take a look at this tomorrow or so...

louwers: @HarelM To answer your question why `text-font` is an array, this is to align with `text-font` in symbol shaders.

So in the current implementation/design to use a particular font face, the `text-font` array in a symbol layer needs to match the `text-font` array in a `font-faces` exactly.

louwers:
This proposal in its current form does **not** interpret `text-font` as a cascading fallback list. Instead, it interprets at as a single font, so it must match the `text-font` specified by `font-faces` exactly. This makes the implementation easier, because MapLibre Native uses a a [hash of the font stack](https://github.com/maplibre/maplibre-native/pull/1439/files#diff-de8805e694044f7ba5566af4b9db5c059a69a93b95ffa1b077b7d97c86ff9070R379) to identify fonts.

However it probably makes sense from a design standpoint to interpret `text-font` as a cascading fallback list. Especially with https://github.com/maplibre/maplibre-style-spec/pull/1068 in mind, where the `text-font` property of symbol layers is now interpreted as a cascading fallback list instead of a single font when `glyphs` is not present, 

So I want to discuss the feasibility and desirability of going from this:

```json
{
  "font-faces": [
    {
      "text-font": [
        "Noto Sans Regular"
      ],
      "font-files": [
        {
          "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
          "unicode-range": [
            "U+1780-17FF"
          ]
        },
        {
          "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
          "unicode-range": [
            "U+1000-109F",
            "U+A9E0-0xA9FF",
            "U+AA60-0xAA7F"
          ]
        },
        {
          "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
          "unicode-range": [
            "U+0900-097F",
            "U+A8E0-0xA8FF"
          ]
        }
      ]
    }
  ]
}
```

to this:

```json
{
  "font-faces": {
    "Noto Sans Regular": [
      {
        "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
        "unicode-range": [
          "U+1780-17FF"
        ]
      },
      {
        "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
        "unicode-range": [
          "U+1000-109F",
          "U+A9E0-0xA9FF",
          "U+AA60-0xAA7F"
        ]
      },
      {
        "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
        "unicode-range": [
          "U+0900-097F",
          "U+A8E0-0xA8FF"
        ]
      }
    ]
  }
}
```

@HarelM Thought the fallback approach is cleaner.

@1ec5 @wipfli Interested to hear your feedback on this.

And lastly, @alanchenboy we already discussed privately, you said you think the proposal as it stands is easier / better. Still, I wonder if we can do some restructuring to facilitate this change, maybe with some help of @sjg-wdw's team.

HarelM:
It would also be great if the spec could support the most simple approach too, but it's nice to have (similar approach to how the sprite is currenrtly defined to support both multi sprite files and single sprite file):
```json
{
  "font-faces": {
    "Noto Sans Regular": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf"
  }
}
```

louwers:
@HarelM Should that then use `U+0-10FFFF` as unicode range?

louwers:
@alanchenboy informed me that interpreting `text-font` as a cascading fallback list fundamentally alters font loading mechanism. It is not a change his is comfortable tackling. So if we want to go with that approach someone else needs to finish the [implementation PR](https://github.com/maplibre/maplibre-native/pull/1439).

I made another PR since it is sufficiently different. https://github.com/maplibre/maplibre-style-spec/pull/1204

PR #1204, opened 2025-06-30
Add `font-faces` with `text-font` as cascading fallback list

louwers:
We need this property to be able to reliably load fonts on with MapLibre Native. Loading font files as opposed to glyphs is required for rendering complex scripts (with Harfbuzz). Relying on fonts being available on the platform is not a viable strategy for MapLibre Native, and I really want to give users the ability to use a style that depends on complex scripts without external platform-specific configuration and distribution. See earlier discussion over at: https://github.com/maplibre/maplibre-style-spec/issues/1045

The initial proposals [here](https://github.com/maplibre/maplibre-style-spec/pull/1184) and [here](https://github.com/maplibre/maplibre-style-spec/issues/868) used the entire `text-font` array to identify a font. This proposal interprets `text-font`  as cascading fallback list when resolving font faces.

Before adopting this specification we need to make sure someone is willing and able to make the needed changes to the [implementation PR](https://github.com/maplibre/maplibre-native/pull/1439) for MapLibre Native to support this specification.

### DIFF
--- a/build/generate-style-spec.ts
+++ b/build/generate-style-spec.ts
@@ -389,10 +389,12 @@ export type SchemaSpecification = {
 // State
 export type StateSpecification = Record<string, SchemaSpecification>;
 
-export type FontFacesSpecification = Record<string, string | {
+export type FontFace = string | {
     url: string,
     "unicode-range"?: string[]
-}>;
+};
+
+export type FontFacesSpecification = Record<string, FontFace>;
 
 ${objectDeclaration('StyleSpecification', spec.$root)}
 


zbigniewmatysek-tomtom:
How do you envision referencing local TTF files? CSS has `src > local`, what would be the equivalent in this proposal?

louwers:
@stefankarschti Asked some more details what this design proposal means for the implementation side of things. I will write some notes here.

The implementation PR currently treats fonts as an array

```
"text-font": [
  "Open Sans Semibold",
  "Arial Unicode MS Bold"
]
```
        
to match up font files you would need to specify the exact same array in font-faces. This is consistent with [how font loading currently works](https://github.com/mapbox/mapbox-gl-native/wiki/Text-Rendering). Basically, MapLibre Native does not implement a fallback mechanism for fonts, in theory you could implement that on the server, because all the font names in the fontstack are included when making a request for glyphs.

But since we we will be loading fonts locally now, design wise there seems to be some agreement that it makes sense to implement some kind of fallback mechanism when looking up font files. 

If you want to dig into what changes are required exactly, look up`FontStack` and `FontStackHash`:

https://github.com/maplibre/maplibre-native/blob/f457c2e0193b2ebab0e07d1007d0963e70139107/include/mbgl/util/font_stack.hpp#L13-L14

```cpp
using FontStack = std::vector<std::string>;
using FontStackHash = std::size_t;
```

These two classes are used to identify a font.

`FontFaces` is defined as:

```cpp
struct FontFace {
    GlyphIDType type; // a unique glyph id

    FontStack fontStack;                               // font stack
    std::string url;                                   // font file url
    std::vector<std::pair<uint32_t, uint32_t>> ranges; // unicode ranges
};

using FontFaces = std::vector<FontFace>;
```

What I think we should do is change the lookup for a `FontFace`. Not look for the FontFace where the fontStack is identical but instead find the first `FontFace` where one of the fonts in the `fontStack` is identical to the fonts contained in the fontstack.

So instead of:

```
findFontFace(fontFaces, fontStack)
  for (fontFace in fontFaces) 
    if (fontFace.fontStack == fontStack)
      return fontFace
```

do this:

```
findFontFace(fontFaces, fontStack)
  for (font in fontStack)
    for (fontFace in fontFaces)
        if (font == fontFace.font) // each fontFace has a single font name instead of a fontStack in the proposal
          return fontFace
```

src/reference/v8.json
```diff
@@ -218,6 +218,48 @@
         }
       }
     },
+    "font-faces": {
+      "type": "array",
+      "value": "object",
+      "doc": "The `font-faces` property can be used to specify what fonts to use. These font faces take precedence over [glyphs](./glyphs.md). Font faces contain information needed to render complex texts such as [Devanagari](https://en.wikipedia.org/wiki/Devanagari), [Khmer](https://en.wikipedia.org/wiki/Khmer_script) among many others.\n\nWhen an optional `unicode-range` is specified the font is only used for characters within that unicode range. When specified it should be an array of strings, each indicating a start and end of a unicode range, similar to the [CSS descriptor with the same name](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range). Refer to the [Unicode Character Code Charts](https://www.unicode.org/charts/) to see ranges for scripts supported by Unicode. To see what unicode code points are available in a font, use a tool like [FontDrop](https://fontdrop.info/).",
+      "example": [
+        {
```
stefankarschti: The PR description uses this JSON as example:
```json
"font-faces": {
  "Noto Sans Regular": [
    {
      "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
      "unicode-range": [
        "U+1780-17FF"
      ]
    },
    {
      "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
      "unicode-range": [
        "U+1000-109F",
        "U+A9E0-0xA9FF",
        "U+AA60-0xAA7F"
      ]
    },
    {
      "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
      "unicode-range": [
        "U+0900-097F",
        "U+A8E0-0xA8FF"
      ]
    }
  ]
}
```

which is different from the one in the specification:

```json
"example": [
        {
          "text-font": [
            "Noto Sans Regular"
          ],
          "font-files": [
            {
              "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/khmer.ttf",
              "unicode-range": [
                "U+1780-17FF"
              ]
            },
            {
              "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/myanmar.ttf",
              "unicode-range": [
                "U+1000-109F",
                "U+A9E0-0xA9FF",
                "U+AA60-0xAA7F"
              ]
            },
            {
              "url": "https://alanchenboy.github.io/harfbuzzResource/fonts/Noto%20Sans%20Regular/devanagari.ttf",
              "unicode-range": [
                "U+0900-097F",
                "U+A8E0-0xA8FF"
              ]
            }
          ]
        }
      ]
```

louwers: Oh sorry, I must have forgot to push it?

1ec5:
Yes, that would definitely be an improvement. All three platforms should already have APIs that can use a `local` specification.

louwers:
Thanks a lot to @stefankarschti for bringing the implementation in line with this spec. https://github.com/maplibre/maplibre-native/pull/3611/files 


If it is OK with you I would like to move the discussion of loading local fonts to a separate discussion. I can make a proposal for allowing `{ local: "Noto Sans Regular" }` in `font-faces` later.

build/generate-docs.ts
```diff
@@ -29,14 +29,17 @@ type JsonSdkSupport = {
...
+
 type JsonObject = {
     required?: boolean;
     units?: string;
     default?: string | number | boolean | {};
     type: string;
     doc: string;
     requires?: any[];
-    example: string | object | number;
+    example?: Example;
```
HarelM: There are two fields? Not just one? Does it make sense?

build/generate-style-spec.ts
```diff
@@ -389,6 +389,11 @@ export type SchemaSpecification = {
 // State
 export type StateSpecification = Record<string, SchemaSpecification>;
 
+export type FontFacesSpecification = Record<string, string | {
```
HarelM: [optional] consider defining the inner font object as a separate type as well.

src/reference/v8.json
```diff
@@ -218,6 +218,45 @@
...
+      "examples": [[{
```
HarelM: I think it would be better to use the existing "example" field, even for multiple examples...

louwers: But how do you differentiate between an example that is an array value and multiple examples?

HarelM: How would it change from a docs perspective? 
I think there are other places that has multiple examples and uses an object or array...
Not sure though...

louwers: `examples` is `[example1, example2]` where both `example1` and `example2` are valid values for the property

You cannot specify `example: [example1, example2]` because that is not a valid value for the property

I can revert it though and collapse these two examples into one.

louwers: I was also confused, probably better to use only one example. Reverted!

PR #1216, opened 2025-07-09
Fix arg arity error message in `index-of` and `slice` expressions

sruenwg:
This fixes the expected number of args described in the error message given when passing an invalid number of args to `index-of` and `slice` expressions.

See the changed `*/unexpected-fourth-arg/test.json` files — The parser currently incorrectly returns "Expected 3 or 4 arguments, but found 4 instead.", which is clearly nonsensical; this PR changes the message to correctly say "Expected 2 or 3 arguments".

## Launch Checklist


HarelM (approved):

HarelM:
Thanks once again!

PR #1246, opened 2025-08-08
Make sure example keys are surrounded by quotes

louwers:
Makes it easier to copy.

Before

image

After

image


HarelM (approved):
Good catch!

PR #1251, opened 2025-08-14
Add encoding field to vector sources

### LINKED ISSUE #1252: Design Proposal: Add encoding field to vector source
# Design Proposal: Add encoding field to vector source to allow for alternative tile formats.

## Motivation

To support [MapLibre Tiles](https://github.com/maplibre/maplibre-tile-spec), or other formats, MapLibre needs to know that a source is in that format.

## Proposed Change

Add an `encoding` field to vector source, closely based on the same field from raster source.

The default is "mvt," indicating the existing behavior.

## Examples

These are equivalent, both producing the same results as any previous style:

```json
"source-vector-default": {
      "type": "vector",
      "url": "https://demotiles.maplibre.org/tiles/tiles.json"
    },
    "source-vector-mapbox": {
      "type": "vector",
      "url": "https://demotiles.maplibre.org/tiles/tiles.json",
      "encoding": "mvt"
    },
```

This would cause MapLibre to expect MLT-format tiles:

```json
    "source-vector-maplibre": {
      "type": "vector",
      "url": "https://demotiles.maplibre.org/mlt-tiles/tiles.json",
      "encoding": "mlt"
    },
```

## API Modifications

No API changes are necessary.

Proposed changes are fully backward-compatible, so no migration is needed.

## Rejected Alternatives

A new source `type`.  

## References

#1251

TimSylvester:
Adds an optional `encoding` field to vector sources, based on the one in raster sources, to allow the use of MapLibre Tiles.

Implementation issues:

- Native: [maplibre-native #3721](https://github.com/maplibre/maplibre-native/issues/3721)
- JS: [maplibre-gl-js #6258](https://github.com/maplibre/maplibre-gl-js/issues/6258)

Resolves #1252

I don't think this has any effect on APIs, can someone confirm?

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license)
 - [x] Link to related issues.
 - [ ] ~~Include before/after visuals or gifs if this PR includes visual changes.~~
 - [ ] ~~Post benchmark scores.~~


src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
...
+        },
+        "maplibre": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mapbox",
+      "doc": "The encoding used by this source. Mapbox Vector Tiles encoding is used by default.",
+      "sdk-support": {
+        "doc": "none yet",
+        "mapbox, terrarium": {
```
HarelM: Probably a copy-paste mistake...?

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
...
+          "doc": "Mapbox Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
+        },
+        "maplibre": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mapbox",
+      "doc": "The encoding used by this source. Mapbox Vector Tiles encoding is used by default.",
+      "sdk-support": {
+        "doc": "none yet",
```
HarelM: We link the relevant issues in this table to indicate anything is approved from a spec perspective and can be tracked in the relevant platform.

src/reference/v8.json
TimSylvester: Yes, good catch.

src/reference/v8.json
TimSylvester: I didn't find any relevant existing ones, do I need to create one? One per platform?

src/validate/validate_source.test.ts
```diff
@@ -0,0 +1,110 @@
+import {validateSource} from './validate_source';
```
HarelM: I was surprised this file didn't exist, but I believe most tests are defined here:
https://github.com/maplibre/maplibre-style-spec/tree/main/test/integration/style-spec/tests

I think adding it there makes sense...

src/reference/v8.json
HarelM: Yes, there's a need to open an issue (if there isn't one opened already) one in native and one in gl-js repo and link to them here.

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
         }
       }
     },
+    "encoding": {
+      "type": "enum",
+      "values": {
+        "mapbox": {
+          "doc": "Mapbox Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
+        },
```
nyurik: To my knowledge, MVT initially stood for Map Vector Tile

```suggestion
        "mvt": {
          "doc": "Map Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
        },
```

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
...
+        "maplibre": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mapbox",
```
nyurik:
```suggestion
      "default": "mvt",
```

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
         }
       }
     },
+    "encoding": {
+      "type": "enum",
+      "values": {
+        "mapbox": {
+          "doc": "Mapbox Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
+        },
+        "maplibre": {
```
nyurik:
```suggestion
        "mlt": {
```

nyurik (commented):
Encoding focuses on "format", not the organization that produced it.  `mlt`, `mvt`, ...

src/reference/v8.json
louwers: Initially, but not anymore. Let's just avoid this by using the abbriviation.

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
         }
       }
     },
+    "encoding": {
+      "type": "enum",
+      "values": {
+        "mapbox": {
```
louwers:
```suggestion
        "mvt": {
```

src/reference/v8.json
louwers: Good idea!

src/reference/v8.json
```diff
@@ -364,6 +364,24 @@
...
+        },
+        "maplibre": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mapbox",
+      "doc": "The encoding used by this source. Mapbox Vector Tiles encoding is used by default.",
+      "sdk-support": {
+        "doc": "none yet",
+        "mapbox, maplibre": {
```
louwers:
```suggestion
        "mvt, mlt": {
```

louwers (commented):
I prefer using `mvt` and `mlt` instead of `mapbox` and `maplibre`.

It is not unthinkable that either Mapbox or MapLibre releases another vector tile format in the future. Then these names would be very confusing. I already think `mapbox` is quite confusing because I personally don't associate MVTs with Mapbox even though they invented the format.

src/reference/v8.json
TimSylvester: I considered avoiding mentioning Mapbox, but "Mapbox Vector Tile" is used both here:

https://github.com/mapbox/vector-tile-spec

and here:

> `vector` | Mapbox Vector Tileset | Binary tile format following the Mapbox Vector Tile spec...

https://docs.mapbox.com/style-spec/reference/sources/

src/reference/v8.json
```diff
@@ -364,6 +364,26 @@
...
+          "doc": "Mapbox Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
+        },
+        "mlt": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mvt",
+      "doc": "The encoding used by this source. Mapbox Vector Tiles encoding is used by default.",
+      "sdk-support": {
+        "mvt, mlt": {
```
HarelM: These should be two "lines", as mvt is already supported.

TimSylvester: Oh yeah, I got distracted trying to figure out what versions to put there.

HarelM:
This looks good in general.
Let's have the design proposal open for a few more days to let more people comment, and if no one has new insights we will be able to merge this.
Thanks for taking the time to open this PR and implement it!

src/reference/v8.json
```diff
@@ -364,6 +364,31 @@
...
+      "values": {
+        "mvt": {
+          "doc": "Mapbox Vector Tiles. See http://github.com/mapbox/vector-tile-spec for more info."
+        },
+        "mlt": {
+          "doc": "MapLibre Vector Tiles. See https://github.com/maplibre/maplibre-tile-spec for more info."
+        }
+      },
+      "default": "mvt",
+      "doc": "The encoding used by this source. Mapbox Vector Tiles encoding is used by default.",
```
CommanderStorm, line 378: 🤔 
Currently, maplibre-gl-js does not set the `Accept`-header for MVT.
For a tileserver it is thus kind of hard to know if a client wants MVT or MLT.
Or tileservers need to be configured to do this statically.

=> could we specify that for MLT we set this header?

HarelM: Would a tile server serve different types of tiles in the same URL?
How is this different from raster DEM for mapbox and terranium encoding?

CommanderStorm: Some tileservers serve different content depending on `Accept` headers.
More specifically, those who are OGC compliant need to do this.

If a tileserver supports more than one output format (mlt/mvt/geojson) this is also a kind of reasonable way of not having to have this configured twice (in the URL and encoding field).

TimSylvester: That would seem to fit with the intent of `Accept`, like a browser indicating its ability to decode image formats.  But the implication is that there's a need is to use the exact same URL but have it produce MLT in some circumstances, and I'm skeptical that that's really necessary.

The mbtiles spec indicates IETF Media Types for the type field and there is a semi-official "application/vnd.mapbox-vector-tile" type for MVT, so there is an appropriate value for that field, and "application/vnd.maplibre-vector-tile" is almost as valid though not officially assigned.

But this isn't quite an aspect of the style spec, more an implicit expectation on the application.  If we're going to have any guidance, I would say it should be something like:

- If the source does not specify any encoding, the application may send one or both of "application/vnd.mapbox-vector-tile" and "application/vnd.maplibre-vector-tile" corresponding to what it supports, using the response type to determine what was returned, defaulting to MVT.
- If the source does specify an encoding explicitly, only the corresponding type should be sent.

However, this does presume that the tile format decision is being made per-tile rather than per-source, which is not currently the case in Native.  It does not seem necessary to me that we support different formats for different tiles in a tileset, and I would argue that this dynamic negotiation is overkill.

If some behavior difference is needed for the same URL, I would advocate a more general facility for sending arbitrary custom headers along with HTTP requests to the given source, allowing integrators to implement their own tracking, authorization, etc., in coordination with the host.

HarelM: There is a transformrequest to allow adding headers in gl-js if needed. Just a side note...

CommanderStorm: It would be very helpful to have a ***standard way for native/gl-js to indicate that they want MLT*** to tileservers.
This could be done via an `Accept` header, a custom `X-…` header, or a recommended conventional URL pattern.

Given that the standard for this seems to be `Accept`, this seems logical, but 🤷🏻‍♂️

Right now, the only option seems to be telling users (in Martin’s docs) that
- they must use `transformRequest`,
- a global config switch to get MLTs
- a tile source config

All of these are not a great user experience, it would be much simpler if the frontends were to communicate what they would want ^^
**The header/.. can also just be in the case when `encodign` is explicitly configured.**

To be clear, this negotiation doesn’t need to be dynamic on the client side — we just need a signal that the server should encode geometry as MLT rather than MVT.

Without such an indicator, there’s really no reliable way for us to know when to use MLT instead of MVT. Currently, we rely on PostGIS/downstream logic, but it seems reasonable to handle this directly.

CommanderStorm: Regarding the IANA Media type, we can just fill out this form to register this => https://www.iana.org/form/media-types

HarelM: While I don't disagree (although requesting {z}/{x}/{y}.mlt vs {z}/{x}/{y}.mvt seems like a good enough solution to me :-)), I still think the client need to know "how to parse" the tiles so I don't see a way around adding this encoding. Adding a header seems like more of implementation details rather part of the spec. Moreover, one can use different tile server that support different features and also offline mb/pmtiles files so header seems like a "nice to have" from this point of view...

pnorman: > semi-official "application/vnd.mapbox-vector-tile" type for MVT

It's not semi-official. It's [official and registered at IANA](https://www.iana.org/assignments/media-types/application/vnd.mapbox-vector-tile).

Content-Negotiation is a well established process for requesting data that may be in different formats. Maplibre SHOULD set these headers when requesting, rather than saying it can accept anything.

As a practical matter, I didn't give this much consideration when writing my tile server. It can only produce MVTs and the URL ends in .mvt.

louwers (approved):

louwers:
@HarelM Let's merge?

test/integration/style-spec/tests/sources.input.json
```diff
@@ -64,6 +64,42 @@
...
+      "type": "vector",
+      "url": "https://demotiles.maplibre.org/tiles/tiles.json",
+      "encoding": 3
+    },
+    "source-vector-array": {
+      "type": "vector",
+      "url": "https://demotiles.maplibre.org/tiles/tiles.json",
+      "encoding": [
+        "mvt"
+      ]
```
pnorman, line 107: The tests only check tilejson usage. Can you add usage with a direct xyz tile URL?

pnorman (changes requested):

test/integration/style-spec/tests/sources.input.json
TimSylvester: There's no interaction between `url` and `encoding` but sure, why not.

louwers:
No need, looks like it is live. https://maplibre.org/maplibre-style-spec/sources/#encoding

acalcutt:
I am probably a bit late here, but I was looking at the MLT maplibre-js PR to get an idea of how I would implement it in tileserver-gl and was wondering about this choice to use 'encloding' instead of 'format' for the type of tile in MLT. it seems like other tile types use 'format' metadata to distinguish between what is in the tile . "encoding" seems like we only use in terrain right now, and for example in tileserver-gl it is set to either mapbox or terrarium right now. This being a vector source it is probably not much of an issue, but it seems different from the convention we have been using.

TimSylvester:
It's just for consistency with the other source definition.

PR #1267, opened 2025-09-02
use `global-state` set on expression instead of the one passed as a parameter to `evaluate` function

melitele:
This PR changes the way `global-state` is passed when evaluating an expression. Instead of passing it to `evaluate` function inside `GlobalProperties` it is set as the expression's `globalState` property.

The expression's `globalState` property is only used when set - passing `global-state` directly to `evaluate` continues to be supported.

While not strictly necessary this change will improve code of the PR [#6321](https://github.com/maplibre/maplibre-gl-js/pull/6321). Specifically with this change the code that attaches `global-state` to expression won't need to override expression's `evaluate` function in `maplibre-gl-js` module as it does in the current version of the PR [#6321](https://github.com/maplibre/maplibre-gl-js/pull/6321):
```
setGlobalState(globalState: Record<string, any>) {
  //@ts-ignore
  this.expression.evaluate = (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState, canonical?: ICanonicalTileID, availableImages?: Array<string>, formattedSection?: FormattedSection) => {
    // force the use of global state object when evaluating any expression
    //@ts-ignore
    globals.globalState = globalState;
    return this._evaluate.call(this.expression, globals, feature, featureState, canonical, availableImages, formattedSection);
  };
}
```
[see code](https://github.com/maplibre/maplibre-gl-js/blob/f3784da125bb265a9f500c0146f1c4552ea5b6e9/src/style/properties.ts#L102)

With this change the code above will be much simpler:
```
setGlobalState(globalState: Record<string, any>) {
  this.expression.globalState = globalState;
}
``` 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Improve patch code coverage.


src/expression/index.ts
```diff
@@ -169,6 +169,7 @@ export class ZoomConstantExpression<Kind extends EvaluationKind> {
     isStateDependent: boolean;
     globalStateRefs: Set<string>;
     _styleExpression: StyleExpression;
+    globalState: Record<string, any> | null;
```
HarelM: Is this set only from the outside? I don't see how this is being used...
Should it be set in the constructor?

HarelM:
Looks like this is missing some tests according to the coverage report...

src/expression/index.ts
melitele: Yes, it is set only from outside in [maplibre-gl-js/src/style/properties.ts](https://github.com/melitele/maplibre-gl-js/blob/ad02f0c04eec49491b18a922c389b15067762ce5/src/style/properties.ts#L98)
If it is set in the constructor of `ZoomConstantExpression` (and `ZoomDependentExpression`) then it'll have to be pushed down through constructors of `StyleLayer` (and all derived layer classes), `Layout`, and `PropertyValue`.

HarelM: So now you can pass globalState from the evaluate method and also just update it from the outside, seems confusing in terms of public API if you'd ask me.
Wouldn't it be better to only have a single way to set global state?

melitele: No, there will only be one way of passing global state: by setting it on expression. It won't be passed as parameter to `evaluate` method.
However removing it from `evaluate` method now means a breaking change. My plan is to add it as an option here, change `maplibre-gl-js` to using set `globalState` on expression over a series of PR separately for layout, paint and filter properties and then remove the option of passing it as a parameter of `evaluate` here.

HarelM: I would prefer a more direct approach if this is acceptable by you, remove it here while creating a breaking change and fix it in maplibre-gl in a single PR.

melitele: In this case you could consider merging the PR [#6321](https://github.com/maplibre/maplibre-gl-js/pull/6321) as it fixes a known bug. I'll continue working on this more comprehensive fix.

src/expression/expression.test.ts
```diff
@@ -69,6 +69,38 @@ describe('evaluate expression', () => {
         expect(value.evaluate({globalState: {}, zoom: 10} as GlobalProperties)).toBe(42);
         expect(console.warn).not.toHaveBeenCalled();
     });
+
+    test('global state as expression property', () => {
```
HarelM, line 73: It's great that you added tests here, don't get me wrong, but I think we recently moved most of the expression tests to files, which you correctly added as well. Do these tests add more use cases?

melitele: coverage for `global-state` in [ZoomConstantExpression.evaluate](https://github.com/melitele/maplibre-style-spec/blob/941e1acee2313b985a6e0d1310addc4140c66c9e/src/expression/index.ts#L213) and [ZoomDependentExpression.evaluate](https://github.com/melitele/maplibre-style-spec/blob/941e1acee2313b985a6e0d1310addc4140c66c9e/src/expression/index.ts#L117)

HarelM: Great! Thanks! I'll release a new version shortly so you can integrate it to maplibre-gl-js.

PR #1279, opened 2025-09-14
Fix passing reference to `global-state` object when creating filter

melitele:
This PR adds `global-state` parameter to `featureFilter` function. Without this fix filters based on `global-state` won't work. It should have been done as part of PR #1267.
 
## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/feature_filter/index.ts
+++ b/src/feature_filter/index.ts
@@ -2,7 +2,7 @@ import {createExpression, findGlobalStateRefs} from '../expression';
 import type {GlobalProperties, Feature} from '../expression';
 import {ICanonicalTileID} from '../tiles_and_coordinates';
 import {StylePropertySpecification} from '..';
-import {ExpressionFilterSpecification} from '../types.g';
+import {ExpressionFilterSpecification, type FilterSpecification} from '../types.g';
 
 type FilterExpression = (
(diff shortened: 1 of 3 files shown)


src/feature_filter/index.ts
```diff
@@ -76,9 +76,10 @@ const filterSpec = {
  *
  * @private
  * @param {Array} filter MapLibre filter
+ * @param {Object} [globalState] - Global state object to be used for evaluating 'global-state' expressions
```
HarelM: I think that those annotations were removed from most of the code (the "{type}"), feel free to remove it from here as well.

src/feature_filter/index.ts
```diff
@@ -76,9 +76,10 @@ const filterSpec = {
...
  * @returns {Function} filter-evaluating function
  */
-export function featureFilter(filter: any): FeatureFilter {
+export function featureFilter(filter: any, globalState?: Record<string, any>): FeatureFilter {
```
HarelM: Can filter type be improved here to avoid the use of `any`?

HarelM (approved):

HarelM:
I've updated the changelog and will release a new version shortly.

PR #1316, opened 2025-10-10
chore(ci): apply a few easy hardenings suggested by zizmore

CommanderStorm:
This PR makes a few of our actions a bit more resistant against supply chain security

In terms of actual risk, the biggest issue (how we bump versions) remains

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -50,6 +50,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v5
         with: { persist-credentials: false }


.github/workflows/test-all.yml
```diff
@@ -6,17 +6,15 @@ on:
   pull_request:
   workflow_dispatch:
 
-permissions:
```
HarelM, line 9: Are you sure about this?

.github/workflows/test-all.yml
```diff
@@ -51,6 +52,7 @@ jobs:
     runs-on: ${{ matrix.os }}
```
CommanderStorm, line 52: we can also remove them here

```suggestion
    runs-on: ${{ matrix.os }}
    permissions:
      contents: read
```

.github/workflows/test-all.yml
CommanderStorm: CI does not seem to need the permissions, not sure why they are here 🤷🏻‍♂️

PR #1335, opened 2025-10-20
Add "unit" option to "number-format"

ginnyTheCat:
This fixes https://github.com/maplibre/maplibre-gl-js/issues/7024 by adding an option to use units in the "number-format" expression.

I encountered this when trying to format heights of mountains as "1,200 m". This works well when just using latin languages, but cyrillic uses a м and other languages change the unit based on number (pluralization rules).

Therefore I think it's worth adding this since units are quite common to find on maps.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/expression/definitions/number_format.ts
+++ b/src/expression/definitions/number_format.ts
@@ -8,11 +8,26 @@ import type {Type} from '../types';
 export class NumberFormat implements Expression {
     type: Type;
     number: Expression;
-    locale: Expression | null; // BCP 47 language tag
-    currency: Expression | null; // ISO 4217 currency code, required if style=currency
-    unit: Expression | null; // CLDR or ECMA-402 unit specifier, required if style=unit
-    minFractionDigits: Expression | null; // Default 0
-    maxFractionDigits: Expression | null; // Default 3
+    /**
+     * BCP 47 language tag
+     */
+    locale: Expression | null;
+    /**
+     * ISO 4217 currency code, required if style=currency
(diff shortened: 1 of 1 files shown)


src/reference/v8.json
```diff
@@ -4051,6 +4056,11 @@
             "js": "0.54.0",
             "android": "8.4.0",
             "ios": "supported"
+          },
+          "`unit` property": {
+            "js": "24.5.0",
```
ginnyTheCat: Not sure what version I should specify here

HarelM: This should link to an open issue in maplibre-gl-js, once the issue is closed and a version is released you should come back and update this field with the released version of maplibre-gl-js (or native if you implement this on native).

src/expression/definitions/number_format.ts
```diff
@@ -10,20 +10,23 @@ export class NumberFormat implements Expression {
     number: Expression;
     locale: Expression | null; // BCP 47 language tag
     currency: Expression | null; // ISO 4217 currency code, required if style=currency
+    unit: Expression | null; // CLDR or ECMA-402 unit specifier, required if style=unit
```
HarelM: Can you change the comments here to be tsdoc comments? (Use `@default` for default values.

HarelM:
You're probably right... 😕

HarelM (approved):

PR #1351, opened 2025-10-31
Add validation for font-faces property

### LINKED ISSUE #1350: `font-faces` should be an object but validator requires an array
maplibre/maplibre-style-spec#1204 added a `font-faces` property to the root of the style specification, along with an example of how to specify downloadable fonts using that property. Unfortunately, that example fails validation:

```
Error: font-faces: array expected, object found
    emitValidationErrors validate_style.ts:48
    emitValidationErrors style.ts:41
    _load style.ts:434
    loadJSON style.ts:423
evented.ts:153:21
```

This is an object:

https://github.com/maplibre/maplibre-style-spec/blob/45ac05d10065021730f66ff0506e8a6f6b4e6c9d/src/reference/v8.json#L225-L251

but this requires an array:

https://github.com/maplibre/maplibre-style-spec/blob/45ac05d10065021730f66ff0506e8a6f6b4e6c9d/src/reference/v8.json#L222

I’m not sure how maplibre/maplibre-native#3611 accepts this property. Does it skip style validation, or is there an override for this specific property?

louwers:
Closes #1350


src/reference/v8.json
```diff
@@ -219,7 +219,7 @@
       }
     },
     "font-faces": {
-      "type": "array",
+      "type": "fontFaces",
```
HarelM, line 222: Is "value" below still relevant?

src/validate/validate_font_faces.test.ts
HarelM, line 1: I think most tests here are duplicate of json files, right? In that case I suggest to remove them. Only keep those that add coverage and can't be done using the json files.

louwers: Which type of tests do you prefer?

HarelM: Json is preferred

HarelM (approved):

PR #1359, opened 2025-11-03
Allow global state expressions in layer visibility

melitele:
- #1364

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/expression/visibility.test.ts
+++ b/src/expression/visibility.test.ts
@@ -3,7 +3,9 @@ import {describe, test, expect, vi} from 'vitest';
 
 describe('create visibility expression', () => {
     test('throws Error for invalid function', () => {
-        expect(() => createVisibilityExpression(['bla'] as any, {})).toThrowError('Unknown expression "bla". If you wanted a literal array, use ["literal", [...]].');
+        expect(() => createVisibilityExpression(['bla'] as any, {})).toThrowError(
+            'Unknown expression "bla". If you wanted a literal array, use ["literal", [...]].'
+        );
     });
 });
 
@@ -16,7 +18,7 @@ describe('evaluate visibility expression', () => {
         expect(value.getGlobalStateRefs().size).toBe(0);
         expect(console.warn).not.toHaveBeenCalled();
     });
-    
+
     test('literal value visible', () => {
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
@@ -52,7 +54,10 @@ describe('evaluate visibility expression', () => {
 
     test('global state flag set to false', () => {
         const globalState: Record<string, any> = {};
-        const value = createVisibilityExpression(['case', ['global-state', 'x'], 'visible', 'none'], globalState);
+        const value = createVisibilityExpression(
+            ['case', ['global-state', 'x'], 'visible', 'none'],
+            globalState
+        );
 
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
@@ -64,7 +69,10 @@ describe('evaluate visibility expression', () => {
 
     test('global state flag set to true', () => {
         const globalState: Record<string, any> = {};
-        const value = createVisibilityExpression(['case', ['global-state', 'x'], 'visible', 'none'], globalState);
+        const value = createVisibilityExpression(
+            ['case', ['global-state', 'x'], 'visible', 'none'],
+            globalState
+        );
 
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
@@ -75,7 +83,10 @@ describe('evaluate visibility expression', () => {
     });
 
     test('falls back to default for invalid expression with zoom', () => {
-        const value = createVisibilityExpression(['case', ['==', ['zoom'], 5], 'none', 'visible'], {});
+        const value = createVisibilityExpression(
+            ['case', ['==', ['zoom'], 5], 'none', 'visible'],
+            {}
+        );
 
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
@@ -89,7 +100,9 @@ describe('evaluate visibility expression', () => {
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
         expect(value.evaluate()).toBe('visible');
-        expect(console.warn).toHaveBeenCalledWith('Expected value to be of type string, but found null instead.');
+        expect(console.warn).toHaveBeenCalledWith(
+            'Expected value to be of type string, but found null instead.'
+        );
     });
 
     test('warns and falls back to default for invalid expression with feature state', () => {
@@ -98,7 +111,9 @@ describe('evaluate visibility expression', () => {
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
         expect(value.evaluate()).toBe('visible');
-        expect(console.warn).toHaveBeenCalledWith('Expected value to be of type string, but found null instead.');
+        expect(console.warn).toHaveBeenCalledWith(
+            'Expected value to be of type string, but found null instead.'
+        );
     });
 
     test('warns and falls back to default for missing global property', () => {
@@ -107,7 +122,9 @@ describe('evaluate visibility expression', () => {
         vi.spyOn(console, 'warn').mockImplementation(() => {});
 
         expect(value.evaluate()).toBe('visible');
-        expect(console.warn).toHaveBeenCalledWith('Expected value to be of type string, but found null instead.');
+        expect(console.warn).toHaveBeenCalledWith(
+            'Expected value to be of type string, but found null instead.'
+        );
     });
 
     test('warns and falls back to default for invalid global property', () => {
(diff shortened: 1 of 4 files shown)


HarelM:
Is this considered a style spec change?
Cc: @louwers

src/expression/visibility.test.ts
HarelM, line 1: I think most exression tests are in json file, aren't they?

melitele: They test different functions. The json tests are for `createPropertyExpression` while `visibility.test.ts` is for `createVisibilityExpression`.

melitele:
> ... Also check out the coverage report, it's a bit "off".

wrong file extension - fixed now

src/expression/visibility.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        globalState.x = 'none';
+        expect(value.evaluate()).toBe('none');
+        expect(console.warn).not.toHaveBeenCalled();
+
+        globalState.x = 'visible';
+        expect(value.evaluate()).toBe('visible');
+        expect(console.warn).not.toHaveBeenCalled();
+    });
+
+    test('global state flag as visibility expression', () => {
```
HarelM: Can you split these tests to have a single expect in them.

src/expression/visibility.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default function createVisibility(visibility: VisibilitySpecification, globalState: Record<string, any>) {
+    const expression = {
+        setValue,
+        evaluate: null
+    };
+    setValue(visibility);
+    return expression;
+
+    function setValue(visibility: VisibilitySpecification) {
```
HarelM: Is there a reason this method is inside the other one?

src/expression/visibility.ts
```diff
@@ -0,0 +1,43 @@
...
+        const compiled = createExpression(visibility, visibilitySpec, globalState);
+        if (compiled.result === 'error') {
+            throw new Error(compiled.value.map(err => `${err.key}: ${err.message}`).join(', '));
+        }
+        expression.evaluate = () => compiled.value.evaluate({} as GlobalProperties);
+        addGlobalStateRefs(expression, () => findGlobalStateRefs(compiled.value.expression));
+    }
+}
+
+function addGlobalStateRefs(visibility, getGlobalStateRefs: () => Set<string> = () => new Set<string>()) {
```
HarelM: types are missing for visibility, also I'm not sure I understand why this is getting a callback as a parameter...?

src/expression/visibility.ts
```diff
@@ -0,0 +1,43 @@
+import {createExpression, findGlobalStateRefs} from '.';
```
HarelM: Should this be imported from a more specific location? It's weird to see import from "."...

src/expression/visibility.ts
```diff
@@ -0,0 +1,43 @@
...
+    expression: {
+        interpolated: false,
+        parameters: ['global-state']
+    },
+    values: {visible: {}, none: {}},
+    transition: false,
+    default: 'visible'
+};
+
+export default function createVisibility(visibility: VisibilitySpecification, globalState: Record<string, any>) {
```
HarelM: What is the return type of this method? What is it doing? Where are the tsdoc comments on it?

src/expression/visibility.ts
melitele: `setValue` needs access to `globalState` object that was passed when creating the visibility expression

src/expression/visibility.ts
melitele: > also I'm not sure I understand why this is getting a callback as a parameter...

how would you do it?

src/expression/visibility.ts
HarelM: If this is the case then `VisibilityExpression` should be a `class` and not a `type`.

src/expression/visibility.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type VisibilityExpression = {
+    evaluate: () => 'visible' | 'none';
+    setValue: (visibility: VisibilitySpecification) => void;
+    getGlobalStateRefs: () => Set<string>;
+};
+
+/**
+ * Creates a visibility expression from a visibility specification.
+ * The visibility expression exposes following functions:
```
HarelM: Another indication that this should be a class and not a type, as this explains the methods in the class, instead of having these methods with tsdocs describing them.

src/expression/visibility.ts
melitele: [a8db65a](https://github.com/maplibre/maplibre-style-spec/pull/1359/commits/a8db65a5411aff5c6f940097f8b8bc2515e850c2)

src/expression/visibility.ts
melitele: [a8db65a](https://github.com/maplibre/maplibre-style-spec/pull/1359/commits/a8db65a5411aff5c6f940097f8b8bc2515e850c2)

src/expression/visibility.ts
```diff
@@ -0,0 +1,64 @@
...
+    expression: {
+        interpolated: false,
+        parameters: ['global-state']
+    },
+    values: {visible: {}, none: {}},
+    transition: false,
+    default: 'visible'
+};
+
+export interface VisibilityExpression {
```
HarelM, line 17: Is this expected to be implemented in another place besides this file?
I'm not sure I understand why the interface is exported and not the class itself.

src/expression/visibility.ts
```diff
@@ -0,0 +1,64 @@
...
+class VisibilityExpressionClass implements VisibilityExpression {
+    private _globalState: Record<string, any>;
+    evaluate: () => 'visible' | 'none';
+    getGlobalStateRefs: () => Set<string>;
+
+    constructor(globalState: Record<string, any>) {
+        this._globalState = globalState;
+    }
+
+    setValue(visibility: VisibilitySpecification) {
```
HarelM, line 47: Can't we simply store the `visibility` as a member and keep the evaluate method instead of override it?
Something like:
```ts
evaluate() {
    if (visibility === null || visibility === undefined || visibility === 'visible' || visibility === 'none') {
        return visibility === 'none' ? 'none' : 'visible';
    }
    return this._compiled.value.evaluate(this._globalState)
}
```
Where _compiled is only calculate when calling `setValue` with an expression.

HarelM: Same for `getGlobalStateRefs` - keep a member that is updated when calling `setValue` and return it, no need to override the method.

src/expression/visibility.ts
```diff
@@ -0,0 +1,64 @@
...
+        this.getGlobalStateRefs = () => findGlobalStateRefs(compiled.value.expression);
+    }
+}
+
+/**
+ * Creates a visibility expression from a visibility specification.
+ */
+export default function createVisibility(visibility: VisibilitySpecification, globalState: Record<string, any>): VisibilityExpression {
+    const expression = new VisibilityExpressionClass(globalState);
+    expression.setValue(visibility);
```
HarelM: Is it possible to create visibility without setting the visibility? If not, I would move the `visibility` parameter to the constructor.

src/expression/visibility.ts
melitele: The reason to override `evaluate` is to avoid calculating the `if` condition every time the value of visibility is checked.

src/expression/visibility.ts
melitele: The class is not used outside this file, the interface is.

src/expression/visibility.ts
melitele: [26f61c2](https://github.com/maplibre/maplibre-style-spec/pull/1359/commits/26f61c206e74965edcdb9026afd1c6e893473209)

src/expression/visibility.ts
melitele: https://github.com/maplibre/maplibre-style-spec/pull/1359/commits/3d3bb0593ee282576446cb91c69d2c0df2021620

src/expression/visibility.ts
HarelM: Why? What are we trying to protect?

src/expression/visibility.ts
HarelM: I don't see a good reason to do this optimization. I prefer readability if this doesn't create any major performance issues, which I don't see how it would.

src/expression/visibility.ts
melitele: `VisibilityExpression` establishes a contract for the return value of `createVisiblity` function: it returns an object with 3 functions: `evaluate`, `setValue`, `getGlobalStateRefs` and nothing more. Replacing it with `VisiblityExpressionClass` would change this contract to returning an object that, in addition to having those 3 functions has a property `_globalState`. The latter is an implementation detail which has no reason to be exposed as part of this contract. In fact, as per previous versions of the implementation, it doesn't need to be a class at all.

src/expression/visibility.ts
melitele: The proposed alternative version of `evaluate` is not only less optimal, however minor, it is also misleading. It suggests that even when `visibility` is specified as literal, something can change it between subsequent calls of the `evaluate` function so that it has to be re-calculated. However the only operation that can change it is `setValue` function and current version reflect that.

HarelM: This can still happen even if evaluate is a method that don't get overwritten when calling `setValue` but instead, `setValue` updates the internal state of the class (like updating the compiled expression or nulling it.
Currently `evaluate` is hard to understand and the state of the class is split between private members and how `setValue` works which updates the `evaluate` method, which is weird...

melitele: Sure thing.

HarelM (approved):

hiddewie:
Thanks both for the effort implementing this 🎉!

PR #1173, opened 2025-06-11
Move Vitest expression tests to JSON

sruenwg:
Converts Vitest expression tests to JSON test fixtures so there is a unified way of testing expressions.
These changes are a reduced/reorganized version of the changes originally implemented in https://github.com/maplibre/maplibre-style-spec/pull/1158. Original motivation [here](https://github.com/maplibre/maplibre-style-spec/pull/1128#issuecomment-2915370037).

As recommended in https://github.com/maplibre/maplibre-style-spec/pull/1158#issuecomment-2953737315, this PR is deliberately kept simple: each commit simply removes all tests for an expression from `expression.test.ts` and creates JSON test fixture files under `test/integration/expression` for any test which did not already exist.
The test folders may thus be slightly disorganized for now — as suggested, I plan to do some refactoring of the tests in a separate PR after this one is merged.

Please refer to this if it is ever unclear where the JSON equivalent for some removed unit test is:
```ts
const testMappings = {
  'distance': {
    // 'Invalid expression'
    'missing geometry': 'missing-geometry',
    'invalid geometry': 'invalid-geometry',
    'expression as geometry': 'expression-geometry',
    
    // 'valid expression'
    'multi point geometry': 'multi-point',
    'multi line geometry': 'multi-line',
    'multi polygon geometry': 'multi-polygon',
    
    'Distance from point': 'from-point',
    'point to point in the same location': 'to-point-same-location',
    'point to point in different location': 'to-point-different-location',
    'point to line': undefined, // not copied; was not actually point-to-line test
    'point to line that passes through the point': 'to-line-through-point-horizontal',
    'point to line that pass through the vertically': 'to-line-through-point-vertical',
    'point to containing polygon': 'to-polygon-contained',
    'point to near by polygon': 'to-polygon-nearby',
    'point to polygon with hole around the point': 'to-polygon-hole',

    'Distance from line': 'from-line',
    'line to point that is on the line': 'to-point-on-line',
    'line to point that is on the middle of the line': 'to-point-on-line-middle',
    'line to point that is close by horizontally': 'to-point-nearby-horizontal',
    'line to point that is close by vertically': 'to-point-nearby-vertical',
    'line to line that are parallel': 'to-line-parallel',
    'line to line that are perpendicular': 'to-line-perpendicular',
    'line to polygon that contains the line': 'to-polygon-contained',
    'line to polygon that is close by': 'to-polygon-nearby',
    'line to polygon with hole around the line': 'to-polygon-hole',

[...]

PR #1372, opened 2025-11-10
Add length=2 for `center`

CommanderStorm:
## Launch Checklist

The `center` property is missing the `length` which means that this does not get properly validated.

### DIFF
--- a/src/validate/validate_array.test.ts
+++ b/src/validate/validate_array.test.ts
@@ -68,54 +68,12 @@ describe('Validate Array', () => {
         expect(errors[0].message).toBe('testArray: array length 2 expected, length 3 found');
     });
 
-    test('Should return error if array length is less than min-length', () => {
-        let errors = validateArray({
-            validateSpec: validate,
-            key: 'testArray',
-            value: [1, 2],
-            valueSpec: {type: 'array', value: 'number', 'min-length': 3},
-            styleSpec: v8
-        });
-        expect(errors).toHaveLength(1);
-        expect(errors[0].message).toBe('testArray: array length at least 3 expected, length 2 found');
-
-        errors = validateArray({
-            validateSpec: validate,
-            key: 'testArray',
-            value: [],
-            valueSpec: {type: 'array', value: 'number', 'min-length': 1},
-            styleSpec: v8
-        });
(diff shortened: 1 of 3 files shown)


src/validate/validate_array.test.ts
HarelM, line 1: I believe most of the validations are in json files, I think the only stuff that are in unit tests are typechecks.
I'm not sure, so please double check me.

CommanderStorm: If you are refering to `min-length`, fair. Good catch.
That does not appear in the spec, so can be removed.

Not sure where the line between types is.
I oriented the kinds of tests to add for this functionality after `src/validate/validate_color_array.test.ts`

Not sure what tests you are objecting to 😉
The `array` expression tests are different, this is not about expressions.

HarelM: These tests I believe:
https://github.com/maplibre/maplibre-style-spec/tree/main/test/integration/style-spec/tests
There was a clean up not long ago to move most of the tests to use json instead of a unit test framework.
Somewhat discussed here:
https://github.com/maplibre/maplibre-style-spec/pull/1173
It might be that it's mostly expression, but I do see the above folder, so let me know what you think.
Don't get me wrong, tests are important.

CommanderStorm: Expressions and type=.. tests are slightly different.

Exprssions just expressions, but it makes sense to test the underlying type system throughly as well.
`validate_array.test.ts` is a test for the code validating the type system code.

Could these tests be json instead? Likely.
Different issue though as the underlying infra does not exist for type-level tests in json.

HarelM: Fair enough. 
 Do you think other tests for center are needed? String, boolean, empty array, object?

CommanderStorm: No, that is what `validate_array.test.ts` asserts in the following test:
> Should return error if array contains invalid element types

I can add a test for this. It just would not add any coverage/..

CommanderStorm: Would also be kind of wasteful to have such a test for every array type in the style spec

PR #1376, opened 2025-11-23
chore: add more example values for expressions

CommanderStorm:


src/reference/v8.json
```diff
@@ -2297,6 +2298,7 @@
           "feature"
         ]
       },
+      "example": ["top", [0, 4], "left", [3,0], "bottom", [1, 1]],
```
HarelM:
```suggestion
      "example": ["top", [0, 4], "left", [3, 0], "bottom", [1, 1]],
```

HarelM (approved):
Thanks!

PR #1388, opened 2025-11-29
fix: filter type system in v8.json being inconsistent to how the rest of the type system works

CommanderStorm:
This PR fixes how the part of the v8.json type system works when it comes to filters


CommanderStorm:
tested the gl-js tests as working. maputnik does not seem to use the typos here

PR #1389, opened 2025-11-29
chore: document expression support better in v8.json

CommanderStorm:
This PR fixes a part of the expression documentation to clarify how expressions work and what expression_name has to do with expression

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -2916,12 +2916,12 @@
   },
   "expression": {
     "type": "array",
-    "value": "*",
+    "value": "expression_name",
     "minimum": 1,
-    "doc": "An expression defines a function that can be used for data-driven style properties or feature filters."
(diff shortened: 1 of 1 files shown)


HarelM:
Might be worth checking with maputnik too as it parses the v8 to generate the relevant UI.

CommanderStorm:
> Does this create a correct typescript types?

How do I check this?
I tried overriding the version of the dependency used but this does not seem to be possible.

or at least not via 

```diff
diff --git a/package.json b/package.json
index 63b89bc15..aeb0d43f4 100644
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@mapbox/unitbezier": "^0.0.1",
     "@mapbox/vector-tile": "^2.0.4",
     "@mapbox/whoots-js": "^3.1.0",
-    "@maplibre/maplibre-gl-style-spec": "^24.3.1",
+    "@maplibre/maplibre-gl-style-spec": "github:maplibre/maplibre-style-spec.git#expressions-reworking",
     "@maplibre/mlt": "^1.1.0",
     "@maplibre/vt-pbf": "^4.1.0",
     "@types/geojson": "^7946.0.16",
```

HarelM:
You can run 
npm pack
In the spec repo which will generate tgz file.
npm i <tgz file>
In the gl js repo.

Or you can use npm link, but I found it complicated...

PR #1391, opened 2025-11-30
chore: reworked how the underlying expressions-type system works

CommanderStorm:
currently our expressions type system is typescript-ish types (not quite) and mostly documented in MD.

I think documenting this in a structured form has advantages, such as moving the docs to the appropriate place:

image

Breakage report:
- [x] gl-js -> ✅ render tests, codegen
- [x] maputnik ->  ✅ unit-tests, dev mode

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -3746,7 +3746,7 @@
         }
       },
       "format": {
-        "doc": "Returns a `formatted` string for displaying mixed-format text in the `text-field` property. The input may contain a string literal or expression, including an [`'image'`](#image) expression. Strings may be followed by a style override object. [Change the case of labels](https://maplibre.org/maplibre-gl-js/docs/examples/change-case-of-labels/)\n\n - [Display and style rich text labels](https://maplibre.org/maplibre-gl-js/docs/examples/display-and-style-rich-text-labels/)",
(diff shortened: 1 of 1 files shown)


src/reference/v8.json
```diff
@@ -3441,13 +3441,13 @@
...
               "parameters": ["interpolation_type", "input", "stop_1_input", "stop_1_output", "...", "stop_n_input", "stop_n_output"],
-              "output-type": "color | array<color>"
+              "output-type": ["color", "array<color>"]
             }
           ],
           "parameters": [
             {
               "name": "interpolation_type",
-              "type": "[\"linear\"] | [\"exponential\", base] | [\"cubic-bezier\", x1, y1, x2, y2]"
+              "type": ["[\"linear\"]", "[\"exponential\", base]", "[\"cubic-bezier\", x1, y1, x2, y2]"]
```
CommanderStorm, line 3450: this should likely be documented as a array with one syntax enum in it.
Still fairly messy as-is.

Different PR though.

CommanderStorm:
I am having a bit of trouble getting the maputnik integration tests to run at all.
They don't seem to work on my PC at all. Not sure how to setup this environment.

Sice the dev mode works and the unit tests run, i am not super concearned.
Also, maputnik does not really support expressions.

CommanderStorm:
> not sure where this is now though, and if it moves us in this direction or not.

I don’t think fully correct and complete codegen is realistically possible right now without manual hacks.
What this PR does achieve is moving us closer to a state where codegen could become viable.

> Might be outside the scope of this PR although I'm not sure what this PR adds then...?

This PR adds value by making the type system for expressions more consistent and easier to work with.

Documenting things in Markdown works for now, but if we want to support expression codegen in the future - and some of our tools already rely on general codegen - then having a more coherent type system becomes increasingly important.

As noted above, this PR doesn’t represent the final state needed for actual codegen, but it is an important stepping stone toward that goal.

HarelM (approved):

PR #1394, opened 2025-11-30
docs: zensical based docs

CommanderStorm:
This PR migrates us from markdown for Material to its successor project.
As discussed in slack a while a go, material has been discontinued and this is now zensical.

image

I have not experienced any issues and have looked at every page we offer here

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,6 @@ The [documentation](https://maplibre.org/maplibre-style-spec) of the style speci
 We use [Zensical](https://www.zensical.org/).
 
 To work on the documentation locally, you need to have Docker installed and running.
-Start MkDocs with
 Start Zensical with
 
 ```bash


louwers:
I think it is easier to install Python than Docker. Also Docker is not very performant on macOS. So I would prefer this approach.

I noticed that the interface shifts slightly when a page with a sidebar is opened. Will report that as a bug.

louwers:
@HarelM You don't need Java, Rust, Go or Rust, right? I think you just need [`uv`](https://docs.astral.sh/uv/), it even installs the right Python version if it is not available on the system.

Please try it out, I know managing Python dependencies used to be a nightmare, but it has gotten downright pleasant since uv came to the scene.

HarelM:
I maintain a lot of projects on my machine, and every time I need to install something globally I fear that I will break another project I'm working on.
Docker solved this problem for most parts, especially for command line tools like this one, I don't see a reason to rush into migrating to zensical and installing anything else besides Docker (and unfortunately nvm as this is mainly a typescript project). We can wait for a docker image for this "side" feature, and probably other fixes that should be coming as other people migrate to it.
There's no security risk here and no real pressure to move.
Don't get me wrong, I do think we should move, but not right now. This can wait a bit.

birkskyum:
can we add a Dockerfile, so it's possible to run in container if preferred?

louwers:
uv doesn't install anything globally. Well, except that it has its own global cache (which makes it really fast). For what it is worth, it got an '[Adopt](https://www.thoughtworks.com/en-de/radar/tools/uv)' rating from the Thoughtworks Technology Radar and is probably one of the [most loved tools](https://blog.jetbrains.com/pycharm/2025/08/the-state-of-python-2025/) in the Python ecosystem.

@HarelM Could you try it out and think about it a little, then if you still don't like it, we can wait and go for a Docker-based approach. But I think you will like it. It's much faster and a nicer experience to use zensical (a native Rust based CLI tool) directly without Docker.

CommanderStorm:
Plus the sidebar looks more modern 😋

HarelM:
I'm having a hard time merging something with the following statement in the docs:
https://zensical.org/docs/community/get-involved/

image

package.json
```diff
@@ -22,8 +22,8 @@
     "generate-style-spec": "node --no-warnings --loader ts-node/esm build/generate-style-spec.ts",
     "generate-typings": "dts-bundle-generator -o ./dist/index.d.ts ./src/index.ts",
     "generate-docs": "node ${WATCH+--watch} --no-warnings --loader ts-node/esm build/generate-docs.ts",
-    "mkdocs": "docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material",
-    "mkdocs-build": "npm run generate-docs && docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build --strict",
+    "docs": "uvx zensical serve --open",
+    "docs-build": "npm run generate-docs && uvx zensical build",
```
CommanderStorm: This is what was requested above.
I think we would be better off with my version, but if this is what it takes to get this shipped, I am fine with it

```suggestion
    "start-docs": "docker run --rm -v ${PWD}:/docs zensical/zensical serve --open",
    "docs": "npm run generate-docs && docker run --rm -v ${PWD}:/docs zensical/zensical build",
```

.github/workflows/build-docs.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       with:
         node-version-file: '.nvmrc'
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
```
CommanderStorm:
```suggestion
```

.github/workflows/build-docs.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       with:
         node-version-file: '.nvmrc'
+    - name: Install uv
```
CommanderStorm:
```suggestion
```

README.md
```diff
@@ -21,10 +21,10 @@ We aim to avoid breaking changes in the MapLibre style specification because it
 ## Documentation
 
 The [documentation](https://maplibre.org/maplibre-style-spec) of the style specification also lives in this repository.
-We use [MkDocs](https://www.mkdocs.org/) with the [Material for MkDocs](https://squidfunk.github.io/mkdocs-material) theme.
+We use [Zensical](https://www.zensical.org/).
 
-To work on the documentation locally, you need to have Docker installed and running.
-Start MkDocs with
+To work on the documentation locally, you need to have [UV](https://docs.astral.sh/uv/getting-started/installation/) and [python](https://www.python.org/downloads/) installed.
```
HarelM: Is this part still needed?

package.json
```diff
@@ -22,8 +22,8 @@
     "generate-style-spec": "node --no-warnings --loader ts-node/esm build/generate-style-spec.ts",
     "generate-typings": "dts-bundle-generator -o ./dist/index.d.ts ./src/index.ts",
     "generate-docs": "node ${WATCH+--watch} --no-warnings --loader ts-node/esm build/generate-docs.ts",
-    "start-docs": "docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material",
-    "docs": "npm run generate-docs && docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build --strict",
+    "start-docs": "docker run --rm -v ${PWD}:/docs zensical/zensical serve --open",
+    "docs": "npm run generate-docs && docker run --rm -v ${PWD}:/docs zensical/zensical build",
```
HarelM, line 26: Is "strict" the default now?

README.md
CommanderStorm:
```suggestion
To work on the documentation locally, you need to have Docker installed and running.
Start MkDocs with
```

package.json
CommanderStorm: `--strict` is not currently supported.

PR #1400, opened 2025-12-04
chore: improve `interpolation` documentation by making it an syntax-enum

CommanderStorm:
This PR documents how interpolations work in a more structured way.

image

I am not sure if my docs are too verbose, I tried to be fairly precisce since cutting is mostly simpler.

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -195,9 +195,7 @@ function expressionSyntaxToMarkdown(key: string, syntax: JsonExpressionSyntax) {
     });
     markdown += `${codeBlockMarkdown(codeBlockLines.join('\n'), 'js')}\n`;
     for (const parameter of syntax.parameters ?? []) {
-        const type = parameterTypeToType(parameter.type)
-            .replaceAll('<', '&lt;')
-            .replaceAll('>', '&gt;');
+        const type = parameterTypeToType(parameter.type);
         markdown += `- \`${parameter.name}\`: \`${type}\``;
         if (parameter.doc) {
             markdown += `- ${parameter.doc}`;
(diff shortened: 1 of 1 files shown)


build/generate-docs.ts
```diff
@@ -217,6 +217,26 @@ function expressionSyntaxToMarkdown(key: string, syntax: JsonExpressionSyntax) {
                 }
             });
         }
+        if (parameter.type === 'interpolation') {
```
HarelM, line 211: Is there a different way to achieve this? I prefer to be as generic as possible here...

CommanderStorm: The other "length1-array with syntax-enum as value" is expressions.
Expressions are manually codegen-ed, so there is no code to reuse.

Also, the indentation and newline requirement (those two spaces before \n) would be a bit tricky to do this way.

Unless you want a major change to how this system works (which might make sense, but would be a larger change)

src/reference/v8.json
```diff
@@ -7992,5 +7994,150 @@
       "type": "string",
       "doc": "A name of a feature property to use as ID for feature state."
     }
+  },
+  "interpolation": {
+    "type": "array",
+    "value": "interpolation_name",
+    "minimum": 1,
+    "doc": "An interpolation defines a function that can be used for transitioning between style properties or feature filters. The first element of an interpolation array is a string naming the interpolation operator, e.g. `\\\"linear\\\"` or `\\\"exponential\\\"`. Elements that follow (if any) are the _arguments_ to the interpolation."
```
CommanderStorm: maybe a bit simpler to read

```suggestion
    "doc": "An interpolation defines how to transition between items. The first element of an interpolation array is a string naming the interpolation operator, e.g. `\\\"linear\\\"` or `\\\"exponential\\\"`. Elements that follow (if any) are the _arguments_ to the interpolation."
```

build/generate-docs.ts
HarelM: In saying the same thing in both comments basically: is there a more generalized way to achieve the same thing? Or maybe reuse some code?
At the very least, since this is specific, it should be extracted to a method.

src/reference/v8.json
```diff
@@ -7992,5 +7994,150 @@
...
+                "base"
+              ],
+              "output-type": "interpolation"
+            }
+          ],
+          "parameters": [
+            {
+              "name": "base",
+              "type": "number literal",
+              "doc": "controls the rate at which the output increases: higher values make the output increase more towards the high end of the range. With values close to 1 the output increases linearly"
```
CommanderStorm: maybe this way?

```suggestion
              "doc": "rate at which the output increases in `f(x) = x^r`. Values higher than 1 increase, close to one behave linearly and below one decrease."
```

src/reference/v8.json
```diff
@@ -7992,5 +7994,150 @@
...
+            },
+            {
+              "name": "x2",
+              "type": "number literal",
+              "doc": "X-coordinate of the second control point (P2). Must be between 0 and 1 for a valid monotonic easing curve. Controls when the curve begins to decelerate toward the end."
+            },
+            {
+              "name": "y2",
+              "type": "number literal",
+              "doc": "Y-coordinate of the second control point (P2). Typically between 0 and 1, but may exceed this range for overshoot effects. Influences the ending slope of the curve."
```
CommanderStorm:
```suggestion
              "doc": "X-coordinate of the first control point. Must be between 0 and 1 for a valid monotonic easing curve. Controls how quickly the curve accelerates at the beginning."
            },
            {
              "name": "y1",
              "type": "number literal",
              "doc": "Y-coordinate of the first control point. Typically between 0 and 1, but may exceed this range for overshoot effects. Influences the starting slope of the curve."
            },
            {
              "name": "x2",
              "type": "number literal",
              "doc": "X-coordinate of the second control point. Must be between 0 and 1 for a valid monotonic easing curve. Controls when the curve begins to decelerate toward the end."
            },
            {
              "name": "y2",
              "type": "number literal",
              "doc": "Y-coordinate of the second control point. Typically between 0 and 1, but may exceed this range for overshoot effects. Influences the ending slope of the curve."
```

build/generate-docs.ts
```diff
@@ -195,33 +195,77 @@ function expressionSyntaxToMarkdown(key: string, syntax: JsonExpressionSyntax) {
     });
     markdown += `${codeBlockMarkdown(codeBlockLines.join('\n'), 'js')}\n`;
     for (const parameter of syntax.parameters ?? []) {
-        const type = parameterTypeToType(parameter.type)
-            .replaceAll('<', '&lt;')
-            .replaceAll('>', '&gt;');
+        const type = parameterTypeToType(parameter.type);
```
CommanderStorm, line 198: unrelated, but was a bad idea to include.

image

build/generate-docs.ts
CommanderStorm: Generalised? Not without basically rewriting the entire thing.
This is one of the core parts and the impact of this is likely larger.
I don't think that reusing code is possible.

I moved the two things you commented on to a function each.

HarelM (approved):
Looks good, thanks! 
Changing the variable type doesn't affect style json validation, right?

PR #1506, opened 2026-02-18
Instructions to setup a local style-spec in downstream projects

eddy-geek:
## Launch Checklist

This is pretty standard node stuff but it was not obvious to me (nor to the AI :P)

Steps were tested with maplibre-gl-js, I am not so sure if all steps are needed for maplibre-native, as it vendors v8.json

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,3 +47,35 @@ In order to publish the package to NPM:
 2. Run the create bump version PR action to create a PR with the version bump
 3. Approve and merge this PR after you have reviewed the Changelog file
 4. Squash and merge the PR, an automatic action will pick up the change and publish the package to npm
+
+# Using a local style-spec downstream.
+
+When developing style-spec changes that aren't yet published, you can point the downstream project (e.g. `maplibre-gl-js` or `maplibre-native`) at a local copy:
+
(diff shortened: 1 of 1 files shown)


HarelM:
I personally usually do the following:
1. Build
2. `npm pack`
3. in other project I use `npm install ../tar.gz`

This has the drawback of the fact that you need to reinstall after every change.

This has the benefit of making sure that a change to the package will be OK after all the commit etc as you are using an installed package and not a symlink..
But I don't have strong feelings either way...

Thanks for looking into this.

HarelM:
I think recent AI can suggest how to do it, and there's too many assumptions here.
Closing. Thanks for this PR!

CommanderStorm:
> I think recent AI can suggest how to do it, and there's too many assumptions here.

The issue is mosty that if you don't know how to do this, it is not possible to know that this is possible.

I think for me this would have helped to know that this is possible..
Any assumptions can be fixed "by AI" too, so that is just a bad argument to make.

I cannot fix something if I don't know that this is the recomented approach.
I was under the assumption that the only way of working with these repos is:
- get shit merged into this repo
- implement the change in gl-js

That it is actually possible to implement them at the same time was not something that I thought was possible.

PR #1545, opened 2026-03-11
fix: mixed legacy+expression filter handling in `maplibre-style-spec` (#1544)

eddy-geek:
Filters like this one were misclassified as legacy filters and then sent through legacy conversion, silently mangling expression-only operators like 'case' and 'global-state', and causing filters to fail without any error.

```ts
['all', ['==', '$type', 'Point'], ['case', ...]]
```

The fix aligns runtime behavior with the deprecation docs: legacy filter syntax and expression syntax cannot be mixed in one filter definition. Instead of attempting broad recursive normalization, mixed filters now fail explicitly with clearer diagnostics and rewrite suggestions where the legacy subfilter has an obvious expression equivalent.

- Updated [isExpressionFilter()] so mixed `all` / `any` / `none` trees are classified as expressions when they contain expression children, preventing accidental fallback into legacy conversion.
- Preserved pure legacy behavior, including legacy `none`, by keeping legacy-only filters on the [convertFilter()] path.
- Added shared mixed-filter diagnostics in [src/feature_filter/index.ts]:
  - detect legacy subfilters inside expression contexts
  - produce explicit error messages
  - suggest expression rewrites for cases like:
    - `["==", "$type", "Point"]`
    - `["in", "$type", "Point", "LineString"]`
    - `["==", "class", "primary"]`
- Reused the same mixed-filter detection/message path in [src/validate/validate_filter.ts] so validation and runtime report consistent errors.
- Expanded integration coverage in [filters.input.json] for mixed cases, including:
  - plain property filters mixed with expressions
  - `$type` equality mixed with expressions
  - mixed `none`
  - `$type in`
  - nested mixed syntax inside `case` conditions
- Rebalanced [feature_filter.test.ts] so it keeps runtime-specific coverage only:
  - the original classification regression shape
  - mixed `none`
  - unsupported legacy special-key operator errors

We considered broader recursive reuse of [validateFilter()] / [validateNonExpressionFilter()] or permissive recursive conversion of mixed trees, but dropped that approach.

Mixed syntax is context-sensitive: only boolean/filter positions are sensible places to detect or reason about legacy-looking subtrees. General recursive conversion would risk implying support in positions where conversion is unsafe or misleading, such as non-boolean expression arguments or result branches.

So this change prefers:

- broad coverage in validation fixtures
- narrow, explicit runtime behavior
- clear failures instead of magical fixups

Mixed legacy + expression filters now fail cleanly with actionable messages, for example:

- `Mixing deprecated filter syntax with expression syntax is not supported. Replace ["==","$type","Point"] with ["==",["geometry-type"],"Point"].`

Unsupported legacy special-key combinations still fail explicitly, e.g.:

- `"$type" cannot be use with operator ">"`

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
@eddy-geek version 6 has started the pre-release process, so we can consider merging this now and releasing a new version.

src/validate/validate_filter.ts
```diff
@@ -4,10 +4,34 @@ import {validateEnum} from './validate_enum';
...
+    for (const index of path) {
+        current = current[index];
+    }
+    return current;
+}
 
 export function validateFilter(options) {
-    if (isExpressionFilter(deepUnbundle(options.value))) {
+    const value = deepUnbundle(options.value);
+    if (isExpressionFilter(value)) {
```
HarelM, line 23: Can this if be negated and reduce indentation in this method?

src/validate/validate_filter.ts
```diff
@@ -4,10 +4,34 @@ import {validateEnum} from './validate_enum';
...
+        const mixedLegacyDiagnostic = findMixedLegacyFilter(value);
+        if (mixedLegacyDiagnostic) {
```
HarelM, line 25: Same as above.

src/validate/validate_filter.ts
```diff
@@ -4,10 +4,34 @@ import {validateEnum} from './validate_enum';
...
 import {unbundle, deepUnbundle} from '../util/unbundle_jsonlint';
 import {extendBy as extend} from '../util/extend';
-import {isExpressionFilter} from '../feature_filter';
+import {
+    findMixedLegacyFilter,
+    getMixedFilterErrorMessage,
+    isExpressionFilter
+} from '../feature_filter';
+
+function getValueAtPath(value, path) {
```
HarelM, line 13: Add types.

src/validate/validate_filter.ts
HarelM: Also add types to `options` object.

src/feature_filter/index.ts
```diff
@@ -46,20 +50,160 @@ export function isExpressionFilter(filter: any): filter is ExpressionFilterSpeci
...
+
+export function findMixedLegacyFilter(
+    filter: unknown,
+    path: Array<number> = []
+): MixedFilterDiagnostic | null {
+    if (!Array.isArray(filter) || filter.length < 1) {
+        return null;
+    }
+
+    const checkChild = (index: number) => {
```
HarelM, line 162: Move this out to a separate method.

src/feature_filter/index.ts
```diff
@@ -46,20 +50,160 @@ export function isExpressionFilter(filter: any): filter is ExpressionFilterSpeci
...
+                const diagnostic = checkChild(i);
+                if (diagnostic) return diagnostic;
+            }
+            break;
+    }
+
+    return null;
+}
+
+export function validateNoMixedExpressionFilter(filter: unknown) {
```
HarelM, line 200: Isn't filter should be of type `FilterSpecification` or something like that?

src/feature_filter/index.ts
```diff
@@ -89,8 +233,13 @@ export function featureFilter(
         return {filter: () => true, needGeometry: false, getGlobalStateRefs: () => new Set()};
     }
 
-    if (!isExpressionFilter(filter)) {
+    if (Array.isArray(filter) && filter[0] === 'none' && isExpressionFilter(filter)) {
+        validateNoMixedExpressionFilter(filter);
```
HarelM, line 237: I would expect the validate to be before the initail if and maybe move the logic of `Array.isArray(filter) && filter[0] === 'none'` inside this method.

HarelM:
What's the performance impact of this change? I would assume filtering is probably a hot path in maplibre...

CommanderStorm:
> What's the performance impact of this change? I would assume filtering is probably a hot path in maplibre...

Filter VALIDATION is unlikely to be in the hot path.
It is also disablable.

Expression eval is also unlikely to be in our hot path according to the evals that I have done, unless you are doing something severly cursed.

HarelM:
Fair point.

HarelM:
I'll merge this to a branch and continue working on it on that branch.

eddy-geek:
Sorry for not following up, thanks for taking it up!

HarelM:
@eddy-geek, there's a comment from @dschep about this PR in the maplibre-gl-js slack channel.
Can you check his findings about a specific expression in stadia maps' style?
https://osmus.slack.com/archives/C03TFH5NE83/p1781568929642539?thread_ts=1781556507.994529&cid=C03TFH5NE83
Thanks!

Issue #1374, opened 2025-11-15
Design Proposal: Color terrain by normal-derived attributes: slope, aspect, curvature, matcap, etc


jo-chemla:
# Design Proposal: Color terrain by normal-derived attributes: slope, aspect, curvature, matcap, etc

## Motivation

https://github.com/maplibre/maplibre-gl-js/issues/6711

Recent maplibre features introduced additional hillshade options (discussion https://github.com/maplibre/maplibre-gl-js/issues/5665 and pr https://github.com/maplibre/maplibre-gl-js/pull/5768 inspired by gdal: combined, multi-dir, igor), hypsometric tint (discussion https://github.com/maplibre/maplibre-style-spec/issues/1067 and pr https://github.com/maplibre/maplibre-gl-js/issues/5666) or contours. In various mapping contexts it is useful to add other visualization modes - based off of the surface normal - for the micro-scale (similar to hillshade) to highlight micro-relief, topographical indicators of on-the-ground features, or for the macro-scale (similar to hypso) to grasp the environment at a glance (aspect, slope etc). 

## Proposed Change

Some of these attributes can be live-computed, shader-side, directly from the surface normal vector. They could then be colormapped via `paint color-relief-color interpolate`:
 - slope is the z component of the normal vector, 
 - aspect is the angle between up/north vector and the normal
 - slightly more convoluted, matcap is using a sphere texture that describes the color of the light ray coming from eye to surface reflected by surface normal (easy equation, but requires eye position being passed to the shader as uniform)
 - others like curvature would be a 2nd order derivative of the surface, so 1st degree derivative of the normal vector field.

The advanced morphological descriptors are probably not needed, but some simple like slope, aspect, and evnentually normals/matcap would be useful. Happy to hear everyone's thoughts about adding (some of) these rendering modes!

```json
{
"layers": [
    {
        "id": "terrain-Visualization",
        "source": "terrain-source",
        "type": "normal-slope", // could be one of normal-slope, normal-aspect, normal-curvature etc
        "paint": {
            "color": ["interpolate",
                ["linear"],
                ["slope"],
                0,"black",
                8000,"#FFFFFF"
            ]
        }
    }
]
}
```

## API Modifications

No change to the API.

## Migration Plan and Compatibility

No change to existing functionality. To use the new functionality, create a `normal-attribute` layer.

## Rejected Alternatives

[...]

PR #1546, opened 2026-03-11
Terrain analysis. related : #1374

eddy-geek:
This is an implementation related to #1374


mizmay:
Hey @eddy-geek, nice work here, take a look at my proposal in #1111 . I didn't explicitly argue for adding `["dem", "slope"]` and `["dem", "aspect"] but instead argued for making space for them. Your input would be useful, particularly if this fits your implementation and you'd still make use of it, or if there's something I missed.

HarelM:
This is currently solved using plugins. If we decide to implement it in the future this PR can be reopened.
Closing it for now...

PR #1603, opened 2026-04-15
Mention uv as an option to serve docs locally

louwers:
I really like not running Docker if I can avoid it. Even on my Linux box, I don't use Docker but podman.

And uv is really nice. But! To each their own. I would like to just mention it as an option in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,13 +25,13 @@ We use [Zensical](https://www.zensical.org/).
 
 For local development of the documentation, start Zensical. When you have Docker installed and running, you can use:
 
-```
+```bash
 npm run start-docs
 ```
 
 Alternatively, if you have [uv](https://docs.astral.sh/uv/) installed, you can use:
 
-```
+```bash
 uvx zensical serve --open
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,17 +23,23 @@ We aim to avoid breaking changes in the MapLibre style specification because it
...
+
+```bash
+uvx zensical serve --open
+```
+
 Most of the documentation is generated (from e.g. `v8.json`).
 In another terminal, run:
 
 ```bash
+npm install
```
louwers, line 42: This is probably second nature to JavaScript developers, but it is not obvious to the rest of the world.

README.md
```diff
@@ -66,28 +72,28 @@ This repo contains scripts for migrating GL styles of any version to the latest
 You can migrate a style like this:
 
 ```bash
-$ gl-style-migrate bright-v7.json > bright-v8.json
+gl-style-migrate bright-v7.json > bright-v8.json
```
louwers, line 75: These dollar signs prevent copying easily, so I'm piggy backing it in this PR.

CommanderStorm (approved):

PR #1604, opened 2026-04-15
add missing color-relief support for MapLibre Native

louwers:
Forgot a few in #1595.


CommanderStorm (approved):
I wonder if we can utilise the render tests to see if this is implemented or the issue is closed and raise an issue/PR

louwers:
@CommanderStorm 
image

I'd say that would save 5 minutes, monthly. Definitelly will need more than 5 hours over a 5 year period to make that reliable. Not worth it!

PR #1132, opened 2025-05-14
use rolldown and tsgo

birkskyum:
Similar to maplibre-gl-js, move from rollup to rolldown for bundling, and ts6 to ts7 for type checking.
## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [ ] Briefly describe the changes in this PR.
 - [ ] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #1627, opened 2026-05-02
Restore Specification type exports dropped in v24.8.2

### LINKED ISSUE #1626: Breaking change in v24.8.2: `StyleSpecification`, `LayerSpecification`, and other types are no longer exported
**Describe the bug**

Since `v24.8.2`, TypeScript types such as `StyleSpecification`, `LayerSpecification`, `FillLayerSpecification`, `LineLayerSpecification`, etc., are no longer exported from the package.

Applications that build on top of the style spec and rely on these typings are affected with errors like:

```
Module '"@maplibre/maplibre-gl-style-spec"' declares 'StyleSpecification' locally, but it is not exported.ts(2459)
```

**To Reproduce**

Use any of the following imports in a TypeScript project with `v24.8.2`:

```ts
import { StyleSpecification, format } from '@maplibre/maplibre-gl-style-spec';
```

```ts
import {
  LayerSpecification,
  FillLayerSpecification,
  LineLayerSpecification,
} from '@maplibre/maplibre-gl-style-spec';
```

**Expected behavior**

Applications continue working after updating from `v24.8.1` to `v24.8.2` (semver).

birkskyum:
Closes #1626 

The v24.8.2 build-tooling switch (dts-bundle-generator → rolldown-plugin-dts)  ( #1132 ) inadvertently dropped the public type exports of `StyleSpecification`, `LayerSpecification`, `FillLayerSpecification`, and ~50 other `*Specification` types. The new emitter only includes what `src/index.ts` explicitly exports; the old one inlined everything reachable from the entry.

The fix is to add in index.ts

`export type * from './types.g';`

To avoid this from happening another time, i used vitest's ability to do type testing, so there is a regression test now.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/index.ts
```diff
@@ -74,6 +74,7 @@ import type {
     AllPaintProperties,
     AllLayoutProperties
 } from './types.g';
+export type * from './types.g';
```
HarelM: This should be placed at the lower part, next to the other exports.
Note that the line above is importing only part of the generated types.
If we are exporting everything from there we might want to avoid import only part of the generated types.
i.e. there's some redundancy here.

test/build/types.test-d.ts
HarelM, line 1: To be honest, I'm not sure this extra complexity is worth this test.
We are checking what we know is missing, and I'm guessing we will forget to update this in the future, so I don't find it very useful...

HarelM: Bottom line, let's just fix the issue and release a new version.
If we find out that it's worth investing time in the future as this will happen again, let's do that then...

src/index.ts
birkskyum: I've moved it to the bottom. Also, the AllPaintProperties/AllLayoutProperties are handled by this wildcard export now.

HarelM (approved):

PR #1650, opened 2026-05-15
feat: add `{line,fill}-layer-opacity`

### LINKED ISSUE #1651: `{line,fill,circle,..}-layer-opacity` for feature-uniform opacity
# Design Proposal: `{line,fill,circle,..}-layer-opacity` for feature-uniform opacity

## Motivation

Map styles **today** expose opacity primarily as a **feature-level** rendering property, where blending is applied during rasterization and compositing of individual geometries within a layer.
This ties visual intensity to feature-by-feature accumulation rather than treating a layer as a single visual unit.

In practice, style authors often **need** to control the **opacity of entire layers as cohesive entities**, for example to:

- fade thematic layers in or out based on zoom
- smoothly transition visibility during storytelling or temporal animation
- adjust emphasis without changing intra-layer feature contrast
- hiding data-overlaps on transparent fill layers

These patterns are currently approximated using workarounds such as layer duplication or feature-level opacity tuning, which do not cleanly separate feature styling from layer-level presentation control.

This proposal introduces a missing abstraction: uniform post-compositing control over a layer’s rendered output.

Currently, having opacity on lines cannot be done since every feature is rendered by itself.
It looks buggy.

## Proposed Change

This proposal introduces a clear separation between:

- feature-level styling (how individual geometries are rendered and blended within a layer), and
- layer-level compositing (how the fully rendered layer is integrated into the map)

`{filll,circle,heatmap,line,symbol,hillshade,color-relief}-layer-opacity` applies a uniform alpha multiplier to the fully composited output of a layer, after feature rendering and before final integration into the framebuffer.

This enables treating a layer as a single visual unit, while preserving existing feature-level opacity semantics. The property is orthogonal to existing *-opacity values and supports transitions and Zoom-based expressions.

So as a precise spec:

> This property applies a **uniform alpha multiplier** to the **fully composited raster result of a single layer**, **after all feature-level opacity blending** and **before integration into the framebuffer**.
It does **not affect in- or inter feature oppactity accumulation**.

<details><summary>Or visually (click to expand)</summary>
<p>

Or in more concrete usecases

| using `line-opacity` vs `line-layer-opacity` |
|--|
| image |

| `line-opacity` | `line-layer-opacity` |
|--------|--------|
| image | image |

See https://github.com/maplibre/maplibre/discussions/503 for more diagrams

</p>
</details> 

## API Modifications

Add `{line,fill,circle,..}-layer-opacity`.

These properties:
- accept values in the range 0..=1
- default to 1
- are transitionable
- support zoom expressions
- are data-constant (because they interact with an entire layer, data-driven or feature-state would not make sense)

This is backwards compatible.

## Rejected Alternatives

Some alternatives and workarounds discussed include:
- continuing to use existing `*-opacity` properties,
- rendering layers into intermediate frame buffers manually,
- duplicating layers with adjusted negative styling,
- or introducing special blending/.. behaviour into existing opacity properties.

However, these approaches are either:
- visually inconsistent,
- inefficient,
- difficult to author,
- or unable to reproduce true feature-uniform opacity behavior cleanly.

Additional discussion and alternative approaches:
https://github.com/maplibre/maplibre/discussions/503.

Furthermore, we rejected a "bare" `layer-opacity` because while consistent with `resampling` and `visibility`, it is inconsistent with the rest of the `line-*` properties, so adding the prefix is maybe better.

## Implementation roadmap

I have PRs are open for:

- [x] line https://github.com/maplibre/maplibre-style-spec/pull/1650 https://github.com/maplibre/maplibre-gl-js/pull/7570
- [x] fill https://github.com/maplibre/maplibre-style-spec/pull/1650 https://github.com/maplibre/maplibre-gl-js/pull/7570
- [ ] symbol
- [ ] circle
- [ ] heatmap
- [ ] fill-extrusion
- [ ] hillshade
- [ ] color-relief
- raster - wontfix because semantically equivalent existing properties exist
- background - wontfix because semantically equivalent existing properties exist

The implementation plan for all layer types is the same as noted above

> This property applies a **uniform alpha multiplier** to the **fully composited raster result of a single layer**, **after all feature-level opacity blending** and **before integration into the framebuffer**.
It does **not affect in- or inter feature opacity accumulation**.

CommanderStorm:
As discussed in https://github.com/maplibre/maplibre/discussions/503 this is the way to go.
Nobody is really going all out for the alternative proposal, so we might as well go forward with this.

- Resolves https://github.com/maplibre/maplibre-style-spec/issues/1651


CommanderStorm:
Not sure if I should add it for the properties where I don't currently have an open PR for.

.gitignore
```diff
@@ -8,8 +8,14 @@ dist
 coverage/
 site/
 docs/*
+!docs/assets
+docs/assets/*
 !docs/assets/extra.css
 !docs/assets/logo.svg
+!docs/assets/hillshade_methods.png
+!docs/assets/resampling.png
```
HarelM: Maybe *.png and *.svg instead of per image file?

.gitignore
```diff
@@ -8,8 +8,14 @@ dist
...
+!docs/assets/layer-opacity-fill.svg
+!docs/assets/layer-opacity-line.svg
```
CommanderStorm:
```suggestion
!docs/assets/*.png
!docs/assets/*.svg
```

.gitignore
CommanderStorm: I was not quite sure why this was done this way tbh.

src/reference/v8.json
```diff
@@ -6055,6 +6055,28 @@
...
     },
+    "layer-opacity": {
+      "type": "number",
+      "default": 1,
+      "minimum": 0,
+      "maximum": 1,
+      "doc": "Opacity applied to the whole fill layer as a single composite. Unlike `fill-opacity` or `fill-color`, both of which apply per feature and make overlapping features look more opaque where they cross, `layer-opacity` is applied to the layer as a whole.\n\n![fill-opacity vs. layer-opacity](assets/layer-opacity-fill.svg)\n\nIf both are set, the effective opacity is `fill-opacity * layer-opacity`. Prefer `fill-opacity` when possible - `layer-opacity` is significantly slower and uses more memory.",
+      "transition": true,
+      "sdk-support": {
+        "basic functionality": {
```
HarelM, line 6066: This will need an update before this can be merged.

CommanderStorm: Ok, created the fanout to the following as requested
- https://github.com/maplibre/maplibre-native/issues/4298

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main
 ### ✨ Features and improvements
+- Added `layer-opacity` paint property to `fill` and `line` layers - applies opacity to the whole layer as a single composite so overlapping features stay uniform. ([#1650](https://github.com/maplibre/maplibre-style-spec/pull/1650)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```
CommanderStorm:
```suggestion
- Added `{line,fill}-layer-opacity` paint property. It applies opacity to the whole layer as a whole layer so overlapping features change in an uniform way. ([#1650](https://github.com/maplibre/maplibre-style-spec/pull/1650)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```

src/reference/v8.json
```diff
@@ -6055,6 +6055,28 @@
       },
       "property-type": "data-driven"
     },
+    "fill-layer-opacity": {
+      "type": "number",
+      "default": 1,
+      "minimum": 0,
+      "maximum": 1,
+      "doc": "Opacity applied to the entire fill layer after compositing. Unlike `fill-opacity` and alpha values in `fill-color`, which are evaluated per feature and therefore accumulate where fills overlap, `fill-layer-opacity` is applied once to the final rendered output of the layer.\n\n![fill-opacity vs. fill-layer-opacity](assets/fill-layer-opacity.svg)\n\nIf both are set, the effective opacity is `fill-opacity * fill-layer-opacity`. Prefer `fill-opacity` when possible - `fill-layer-opacity` is significantly slower and uses more memory.",
```
mizmay:
```suggestion
      "doc": "Opacity applied to the layer as a whole. Unlike `fill-opacity` and alpha values in `fill-color`, which are applied per feature and accumulate where fills overlap, `fill-layer-opacity` is applied once so overlapping fills appear as a single surface at the given opacity. When both `fill-opacity` and `fill-layer-opacity` are set, the per-feature, overlap accumulation is preserved, then layer-opacity composites the result. This property affects only the layer it is set on and cannot be used to composite multiple layers together or control how this layer blends with layers above or below it.
```

mizmay: I'd recommend against including the SVG, there are not images in the spec typically and it is confusing whether it is depicting features within the same layer or different layers. Also GitHub is indicating the file is broken. Better to put this in a guide.

src/reference/v8.json
```diff
@@ -6451,6 +6473,28 @@
       },
       "property-type": "data-driven"
     },
+    "line-layer-opacity": {
+      "type": "number",
+      "default": 1,
+      "minimum": 0,
+      "maximum": 1,
+      "doc": "Opacity applied to the entire line layer after compositing. Unlike `line-opacity` and alpha values in `line-color`, which are evaluated per feature and therefore accumulate where lines overlap, `line-layer-opacity` is applied once to the final rendered output of the layer.\n\n![line-opacity vs. line-layer-opacity](assets/line-layer-opacity-line.svg)\n\nIf both are set, the effective opacity is `line-opacity * line-layer-opacity`. Prefer `line-opacity` when possible - `line-layer-opacity` is significantly slower and uses more memory.",
```
mizmay:
```suggestion
      "doc": "Opacity applied to the layer as a whole. Unlike `line-opacity` and alpha values in `line-color`, which are applied per feature and accumulate where lines overlap, `line-layer-opacity` is applied once so overlapping lines appear as a single surface at the given opacity. When both `line-opacity` and `line-layer-opacity` are set, the per-feature, overlap accumulation is preserved, then layer-opacity composites the result. This property affects only the layer it is set on and cannot be used to composite multiple layers together or control how this layer blends with layers above or below it.",
```

src/reference/v8.json
mizmay: Also it isn't a good idea to comment on performance in the spec.

mizmay (commented):
Suggested more spec-friendly language.

src/reference/v8.json
CommanderStorm: > I'd recommend against including the SVG, there are not images in the spec typically and it is confusing whether it is depicting features within the same layer or different layers. 
> Better to put this in a guide.

Which guide? We don't have a guide.. we have a spec.
Maybe we should have a "how to style" guide, but that is a differnt story.

Same as for hill shade options, I think an image clears this up immediately WHY this exists and what it does in comparison.

-> I would like to keep an image. What it displays I am flexible on.

> Also GitHub is indicating the file is broken. 

Cannot reproduce, the files tab is showing it correctly. Can you share a screenshot what you mean by "broken"?

CommanderStorm: > it isn't a good idea to comment on performance in the spec.

I don't see a path using the current algorithm that we can achive the same speed.
If we find a better way, we can remove the disclaimer, but it is a significant difference in what this property does in comparison to the other property doing similar things.

As a developer, this is the information that I would like to see.

CommanderStorm:
```suggestion
      "doc": "Opacity applied to the layer as a whole. Unlike `fill-opacity` and alpha values in `fill-color`, which are applied per feature and accumulate where fills overlap, `fill-layer-opacity` is applied once so overlapping fills appear as a single surface at the given opacity. When both `fill-opacity` and `fill-layer-opacity` are set, the per-feature, overlap accumulation is preserved, then `fill-layer-opacity` composites the result. This property affects only the layer it is set on and cannot be used to composite multiple layers together or control how this layer blends with layers above or below it. Prefer `fill-opacity` when possible - `fill-layer-opacity` is slower and uses more memory.\n\n![fill-opacity vs. fill-layer-opacity](assets/fill-layer-opacity.svg)",
```

mizmay: I didn't try to build the docs so if it looked right maybe just weirdness in the Github UI. 

A future guide that explains the aspects that don't quite belong in the spec: https://maplibre.org/maplibre-gl-js/docs/guides/

The hillshade image is great, I think it would be more visible in a guide than buried in a very long scroll. 

I am a purist here, the style spec should be words. But if you were to make a picture, I'd expect it to make clear the difference between `*-opacity`, `*-layer-opacity` *and both used together*. I'd recommend using the same color, not different ones to make it clearer these are features in the same layer.

src/reference/v8.json
CommanderStorm:
```suggestion
      "doc": "Opacity applied to the layer as a whole. Unlike `line-opacity` and alpha values in `line-color`, which are applied per feature and accumulate where lines overlap, `line-layer-opacity` is applied once so overlapping lines appear as a single surface at the given opacity. When both `line-opacity` and `line-layer-opacity` are set, the per-feature, overlap accumulation is preserved, then layer-opacity composites the result. This property affects only the layer it is set on and cannot be used to composite multiple layers together or control how this layer blends with layers above or below it. Prefer `line-opacity` when possible - `line-layer-opacity` is slower and uses more memory.\n\n![line-opacity vs. line-layer-opacity](assets/line-layer-opacity-line.svg)",
```

src/reference/v8.json
CommanderStorm: I am a developer and the image immediately tells me what this property does.
Same as the resampling one.

The text needs conscious effort and is specifically tuned to give understanding of what this does and how it interacts with other properties.

image

mizmay: > I am a developer and the image immediately tells me what this property does.


This is a good argument for creating a guide, so that you can explain, visually to other developers what this property does. The spec has many other audiences, notably: non-developers, and applications that consume and display it in their own context downstream. You can't control that context. In some markets, like the United States, your image with no alternative description, will be a problem for those who must be WC3 compliant under federal law.

CommanderStorm: > your image with no alternative description

The image has a description describing what the image is showing, no change nessary.

Not sure what "WC3 compliant under federal law" requires.
Since we never audited for this and nobody added tooling to our CI 🤷🏻‍♂️

mizmay (approved):
Approving, with reservations. I don't want to be a blocker on this.

The headers on the diagram run together, please fix.

CommanderStorm:
> The headers on the diagram run together, please fix.

I am not sure if I follow…
Looks good to me

CommanderStorm:
I can fix all docs related feedback in a followup

mizmay:
Because the header row on that diagram runs together, it is difficult to understand what the titles mean.

A guide that fully explains how to use this new property would make more sense.

PR #1684, opened 2026-06-06
fix: resolve global-state to null for unset key when other keys are set

### LINKED ISSUE #1531: Global state value becomes `undefined` when not set but other keys are set
**Describe the bug**
When using the `["global-state", key]` expression with a key that has not been set yet, and if any other keys _have_ been set, the value is resolved as `undefined`.

**To Reproduce**
I made a tiny reproduction at https://github.com/Alduino/maplibre-global-state-reprod, you can run it with `pnpm run dev`.
It just sets the text colour to red if `["==", ["global-state", "test"], null]` or blue otherwise.
1. Text initially is red; global state object is empty
2. Click `Set "unrelated" to true` which sets a different global state key to a value; text becomes blue due to the resolved value for `test` becoming `undefined`
3. Click `Set "test" to null` to explicitly set the test global state value to null; text becomes red again.

Or if it's easier to read pseudocode:
```js
evaluate(["==", ["global-state", "test"], null]); // null
map.setGlobalState("unrelated", true);
evaluate(["==", ["global-state", "test"], null]); // undefined(!)
map.setGlobalState("test", null);
evaluate(["==", ["global-state", "test"], null]); // back to null
```

**Expected behavior**
As [described in the docs](https://maplibre.org/maplibre-style-spec/expressions/#global-state) the global state expression should return `null` if the value isn't set.

**Screenshots**
| | | |
|---|---|---|
| Image | Image | Image |

**Desktop (please complete the following information):**
 - OS: Fedora 43 KDE Plasma Edition
 - Browser: Firefox 148
 - Version: maplibre-gl v5.19.0, @maplibre/maplibre-gl-style-spec 24.6.0

**Smartphone (please complete the following information):**
 - N/A

**Additional context**
Not sure if this should be posted here or in the maplibre-gl-js repo, issue seems to be from the code in this repo though.

As far as I can tell it's coming from here:
- https://github.com/maplibre/maplibre-style-spec/blob/8d85feb97dbe27112f4504ee6317d3e0b55bfed7/src/expression/definitions/global_state.ts#L38-L44
- https://github.com/maplibre/maplibre-style-spec/blob/8d85feb97dbe27112f4504ee6317d3e0b55bfed7/src/util/get_own.ts#L13-L15

If the global state object isn't defined or has no keys then `null` is returned, otherwise `getOwn` returns `undefined` instead of `null`.
Maybe `evaluate` could return `getOwn() ?? null`? That would also return `null` even if the value is explicitly set to `undefined`, although seeing as afaict in the style spec `undefined` basically doesn't exist maybe that would make sense too?

SAY-5:
Fixes #1531.

`["global-state", key]` returned `undefined` instead of `null` when the requested key was not set but some other global state key was, because `getOwn` returns `undefined` for a missing key. This made comparisons like `["==", ["global-state", key], null]` evaluate to `false` once an unrelated key existed. Coalescing the lookup with `?? null` matches the documented behaviour, as suggested in the issue.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 - _...Add new stuff here..._
+- Resolve `["global-state", key]` to `null` instead of `undefined` when the key is unset but other global state keys are set ([#1531](https://github.com/maplibre/maplibre-style-spec/issues/1531))
 
 ## 24.8.5
 
--- a/src/expression/definitions/global_state.ts
+++ b/src/expression/definitions/global_state.ts
@@ -40,7 +40,7 @@ export class GlobalState implements Expression {
 
         if (!globalState || Object.keys(globalState).length === 0) return null;
 
-        return getOwn(globalState, this.key);
+        return getOwn(globalState, this.key) ?? null;
     }
 
     eachChild() {}
--- a/test/integration/expression/tests/global-state/unset-with-other-keys/test.json
+++ b/test/integration/expression/tests/global-state/unset-with-other-keys/test.json
@@ -1,0 +1,30 @@
+{
+  "expression": [
+    "==",
+    [
+      "global-state",
+      "x"
+    ],
+    null
+  ],
+  "inputs": [
+    [
+      {},
+      {}
+    ]
+  ],
+  "globalState": {
+    "y": true
+  },
+  "expected": {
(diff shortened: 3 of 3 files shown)


HarelM (approved):
Nice work!

PR #1709, opened 2026-06-15
fix: mixed legacy+expression filter handling

HarelM:
## Launch Checklist

A follow up PR on this one:
- #1545

It changes filter expression validation to raise errors instead of creating invalid expressions that are silently failing.

 - [x] Link to related issues.


src/feature_filter/index.ts
```diff
@@ -46,20 +50,182 @@ export function isExpressionFilter(filter: any): filter is ExpressionFilterSpeci
...
     }
 }
 
+function getFilterPropertyExpression(property: string): unknown {
+    switch (property) {
+        case '$id':
+            return ['id'];
+        default:
+            return ['get', property];
+    }
```
CommanderStorm: nit: I would not use a switch here..


```suggestion
    if (property === '$id') return ['id'];
    return ['get', property];
```

CommanderStorm (approved):

src/feature_filter/index.ts
HarelM: Fair point.

HarelM: Done in 09dc028

HarelM:
I'm not 100% sure, but the remaining missing coverage seems like dead code, but I'm a bit hesitance to remove it. one of them is a default switch case, which I think makes sense, the other one is a `has` expression, which I'm not sure how to make it legacy, since it is also a "modern" expression as well.
Bottom line, merging...

PR #1712, opened 2026-06-16
has filter returns false for undefined property values

xavierjs:
A property set to `undefined` is semantically equivalent to a missing property. The `get` expression already returns `null` for `undefined` values, so `has` should be consistent and return `false` in that case. This is the behavious of filtering in previous Maplibre (like the 5.6.10)

`has` and `filter-has` expressions now return `false` when a property exists but is set to `undefined`, treating it the same as a missing property
Previously, `has` only checked `key in obj`, which returned `true` for properties explicitly set to `undefined`                                                                                                                                   

 - [X] Link to related issues.
 - [X] Include before/after visuals or gifs if this PR includes visual changes.
 - [X] Post benchmark scores.

### DIFF
--- a/src/feature_filter/feature_filter.test.ts
+++ b/src/feature_filter/feature_filter.test.ts
@@ -646,7 +646,7 @@ describe('legacy filter tests', () => {
             expect(f({zoom: 0}, {properties: {foo: true}} as any as Feature)).toBe(true);
             expect(f({zoom: 0}, {properties: {foo: false}} as any as Feature)).toBe(true);
             expect(f({zoom: 0}, {properties: {foo: null}} as any as Feature)).toBe(true);
-            expect(f({zoom: 0}, {properties: {foo: undefined}} as any as Feature)).toBe(true);
+            expect(f({zoom: 0}, {properties: {foo: undefined}} as any as Feature)).toBe(false);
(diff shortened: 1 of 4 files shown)


src/feature_filter/feature_filter.test.ts
```diff
@@ -646,7 +646,7 @@ describe('legacy filter tests', () => {
             expect(f({zoom: 0}, {properties: {foo: true}} as any as Feature)).toBe(true);
             expect(f({zoom: 0}, {properties: {foo: false}} as any as Feature)).toBe(true);
             expect(f({zoom: 0}, {properties: {foo: null}} as any as Feature)).toBe(true);
-            expect(f({zoom: 0}, {properties: {foo: undefined}} as any as Feature)).toBe(true);
+            expect(f({zoom: 0}, {properties: {foo: undefined}} as any as Feature)).toBe(false);
```
CommanderStorm, line 651: Could you add a code comment here and below why `null` and `undefined` SHOULD behave differently?
That otherwise will confuse the next reader...

HarelM:
LGTM, @CommanderStorm I'll let you approve if there's something else you might want to be changed.

CommanderStorm (approved):

PR #1713, opened 2026-06-19
test: enable restoreMocks and isolate console.warn spies in expression tests

CommanderStorm:
Currently, we don't restore our mocks, which leads to really odd artefacts in the tests that are stable only because they are exectued in a determinisitc order


vitest.config.build.ts
```diff
@@ -4,6 +4,9 @@ export default defineConfig({
     test: {
         name: 'build',
         environment: 'node',
+        // Restore spies to their originals and clear call history before each test,
```
HarelM:
```suggestion
```

vitest.config.build.ts
```diff
@@ -4,6 +4,9 @@ export default defineConfig({
...
+        // so mock state never leaks across tests in a file.
```
HarelM:
```suggestion
```

HarelM (approved):
THANKS!

PR #1714, opened 2026-06-19
feat: improve runtime error messages

### LINKED ISSUE #1049: Improve evaluate warning messages
# Design Proposal: Improve evaluate warning messages

## Motivation

Some errors in style evaluation throw errors, that are caught and written as a warning in console, but the message is not helpful enough to find the real cause of the issue.

## Proposed Change

Style expression evaluation process like can result in error like [this](https://github.com/maplibre/maplibre-style-spec/blob/7229c705f2e19c8c65ab8275eed796d8cc2b38f4/src/expression/definitions/assertion.ts#L92). It would be helpful, if the message contained more info about the evaluation context. At least the `feature` data, better also info about the expression definition and style layer name, where it is defined.

CommanderStorm:
Currently, when a gl-js runtime error occurs, we just error out with an

image

This is hard for me and for other users to debug and does really not tell the user what is happening.

We should tell the user where this happened.
So, this PR adds a mandatory, `readonly key` to the signature that then gets passed out on encoutering an runtime error.

Lots of the same change allover, and a few unrelated chagnes which will go away assuming we merge my other PRs.
So for now, this is a draft due to the unrelated changes

Resolves #1049


src/expression/types/variable_anchor_offset_collection.test.ts
```diff
@@ -38,12 +38,17 @@ describe('VariableAnchorOffsetCollection', () => {
 
         test('should throw with mismatched endpoints', () => {
             expect(() =>
-                i11nFn(parseFn(['top', [0, 0]]), parseFn(['bottom', [1, 1]]), 0.5)
+                i11nFn(parseFn(['top', [0, 0]]), parseFn(['bottom', [1, 1]]), 0.5, '')
```
HarelM, line 41: I think it would be better to pass a key, assuming the parameter that was added here is the key.

src/feature_filter/index.ts
```diff
@@ -231,11 +231,14 @@ const filterSpec = {
...
  * @private
  * @param filter MapLibre filter
+ * @param rootKey Location of the filter in the style JSON (e.g. `layers[3].filter`),
+ * used to prefix runtime warnings
  * @param [globalState] Global state object to be used for evaluating 'global-state' expressions
  * @returns filter-evaluating function
  */
 export function featureFilter(
     filter: FilterSpecification | void,
+    rootKey: string,
```
HarelM, line 222: Is this a breaking change?

CommanderStorm: Yes, but we only break ourselves and I will plumb this up in the next pr of the series

HarelM: Do we want to introduce this as a non breaking change first (and later make it mandatory) as we just released a breaking change version recently...

HarelM:
The `expression[0]` in the json integration tests is not what I would've expected to see in these files. I'm guessing this is defined in the test execution. I wonder if this can be improved to showcase the added value this PR brings.

As a site note, you might consider moving the classes' field to be constructor parameters to avoid the duplication in defining the fields and then adding them to the constructor signature (this is a "nice-to-have", but since I saw it I felt obligated to raise the idea).

src/feature_filter/index.ts
CommanderStorm: Are there users for the expression eval part?
What you are asking for is a lot mor churn in terms of code. Let's not churn unless there is a reason.

All "simple fixes" (like `= ""`) result in bad error messages or need churn where I fix it here, fix it GL js, fix it here, fix it GL js compared to fix it here, fix it gl-js.

HarelM: We don't really know how people are using this library, so it's hard to say. 
I'm not sure I understand the last comment about the churn.
The bottom line from my point of view, is considering the issue linked and the screenshot if the problem I'm not sure I understand how would this look after the fix. Can you provide the same screenshot of the warning after the fix here?

CommanderStorm: I don't think I can provide that screenshot because I cannot use this code in gl-js without a merge and release.
Or is there a trick that I don't know?

HarelM: You can build it locally, pack and then use it in maplibre locally using npm link or simply installing the tar.gz file.

CommanderStorm:
will give it another attempt where I open both PRs at the same time

src/feature_filter/index.ts
CommanderStorm: It would be really handy if you could document how to do this, because for me as not a JS dev, I don't quite know how to do all of this linking or tar.gz s.

Our docs for this repo or gl-js don't contain a word how to develop them at the same time..

HarelM: There's was a different PR on this, but it's not specific to this repo, it's a general JavaScript process, and AI can explain it better I believe:
https://github.com/maplibre/maplibre-style-spec/pull/1506

I felt it shouldn't be here though, but in maplibre-gl repo. 

In any case, feel free to add it where you think you might be looking for it...

PR #1727, opened 2026-06-23
fix: legacy+expression filter error suggests a broken expression

ciscorn:
Follow-up to #1709 

I found the mixed legacy+expression filter error (#1709) can suggest invalid expressions for "$type" filters. It emits a one-argument `["==", ["in", ...]]`.

The fix suggests the bare `["in", ...]` for `"=="` and `["!", ["in", ...]]` for `"!="`.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [ ] ~Include before/after visuals or gifs if this PR includes visual changes.~
 - [x] Write tests for all new functionality.
 - [ ] ~Document any changes to public APIs.~
 - [ ] ~Post benchmark scores.~
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

Issue #965, opened 2025-01-09
Adjust `geometry-type` expression to return `Point`, `LineString` or `Polygon`?


louwers:
Historically, in both MapLibre GL JS up to v4 and MapLibre Native, [contrary to what the style specification specifies](https://maplibre.org/maplibre-style-spec/expressions/#geometry-type), the `geometry-type` expression returned `Point`, `LineString` or `Polygon`. This was reported in https://github.com/maplibre/maplibre-gl-js/issues/3516 as a bug, fixed in https://github.com/maplibre/maplibre-style-spec/pull/519 and released as a breaking change in MapLibre GL JS 5.0.0.

[^1]: both MapLibre GL JS up to v4 and MapLibre Native

We had a discussion about this topic [on Slack](https://osmus.slack.com/archives/C01G3D28DAB/p1736226120557009) and during today's TSC Meeting. I am creating this issue, because I feel we haven't reached a consensus on this topic yet. To add some more context to the discussion I tried the styles of some MapLibre sponsors. Most have issues.

**Jawg Streets**

<img src="https://github.com/user-attachments/assets/f5464c0c-9044-478a-a74e-6949389ff73c" width="45%" />
<img src="https://github.com/user-attachments/assets/86f3441e-3138-497b-9c7a-7223cab8a123" width="45%" />

**AWS Open**

<img width="45%" alt="image" src="https://github.com/user-attachments/assets/7b7350f0-d6d6-41c1-add5-948ddf5dae58" />
<img width="45%"  alt="image" src="https://github.com/user-attachments/assets/02f15d01-eb75-42d9-9799-2500442efb71" />

**Radar Maps**

<img width="45%" alt="image" src="https://github.com/user-attachments/assets/d9697033-2064-46c5-800e-94c7e47a0cb2" />

<img width="45%" alt="image" src="https://github.com/user-attachments/assets/891ac171-ef7f-415e-a905-db0973efbf49" />

**Facebook Light**

<img width="45%" alt="Image" src="https://github.com/user-attachments/assets/48b4ee4b-08ac-44db-8df7-7d90de010a31" />

<img width="45%" alt="Image" src="https://github.com/user-attachments/assets/073bf637-08e7-4c9f-804a-d3b0aa1717ac" />

---

Obligatory XKCD

<img width="200px" alt="image" src="https://github.com/user-attachments/assets/1ed35bce-4c22-4adf-aebf-9bcf84298f2f" />

louwers:
[...]

PR #1728, opened 2026-06-25
fix: simplify legacy+expression $type filter suggestions

ciscorn:
More follow-up to #1727, #1709.

@HarelM's comment https://github.com/maplibre/maplibre-style-spec/pull/1727#issuecomment-4790855282:
> I think this was reported as a wrong suggestion as $type and geometry-type are acting the same, so there's no need for the literal and "in" complexity...

## Launch Checklist

 - [x] Link to related issues.
 - ~[ ] Include before/after visuals or gifs if this PR includes visual changes.~
 - ~[ ] Document any changes to public APIs.~
 - ~[ ] Post benchmark scores.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🐞 Bug fixes
 - `has` filter now returns `false` when a property exists but is set to `undefined` ([#1712](https://github.com/maplibre/maplibre-style-spec/pull/1712)) (by [@xavierjs](https://github.com/xavierjs))
-- Fix incorrect replacement suggestion in error messages for mixed legacy+expression `$type` filters ([#1727](https://github.com/maplibre/maplibre-style-spec/pull/1727)) (by [@ciscorn](https://github.com/ciscorn))
+- Fix incorrect replacement suggestion in error messages for mixed legacy+expression `$type` filters ([#1727](https://github.com/maplibre/maplibre-style-spec/pull/1727), [#XXXX](https://github.com/maplibre/maplibre-style-spec/pull/XXXX)) (by [@ciscorn](https://github.com/ciscorn))
 - _...Add new stuff here..._
 
 ## 25.0.0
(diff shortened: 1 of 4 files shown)


CommanderStorm:
No, @HarelM must have meant something else.
I think the suggestion is correct as-merged.

`$type` for `Polygon` will match `Polygon` and `MultiPolygon` while `geometry-type` will match the exact geometry type specified.
Therefore, `in` is nesseary.

Context why there is both `$type` and `geometry-type` and why they behave this way in case you are curious:
https://github.com/maplibre/maplibre-style-spec/issues/965

HarelM:
I believe geometry-type exact match was reverted in 5.0.1 and geometry-type and $type behave the same in latest versions: 
https://github.com/maplibre/maplibre-style-spec/issues/965#issuecomment-2585862428

Feel free to fact check me on this though.

ciscorn:
@CommanderStorm @HarelM  Thanks both. I'm a bit confused, but looking at the latest code, `geometry-type` seems to behave the same as `$type`. No `Multi*` in [`geometryTypes`](https://github.com/maplibre/maplibre-style-spec/blob/v25.0.0/src/expression/evaluation_context.ts#L6).

So maybe the `in` / `literal` isn't needed here? Not fully sure though.

HarelM:
I think the in/litwral is not needed, I mistakenly added it as I remembered something "wrongly".

HarelM:
The following render test confirms it, `gemotry-type` has the same output as `$type`:
https://github.com/maplibre/maplibre-gl-js/blob/main/test/integration/render/tests/geometry-type/all-geometries/expected.png

What I've added in the following commit is probably wrong and need to be revered:
https://github.com/maplibre/maplibre-style-spec/pull/1709/changes/bc6260e9a3bcd006f24fb67a0fc8039d4b27172c

HarelM (approved):

PR #1740, opened 2026-07-05
Allow debugging runtime warnings

### LINKED ISSUE #1049: Improve evaluate warning messages

CommanderStorm:
Fixes 

| before | after |
|--------|--------|
|  |  |

Resolves https://github.com/maplibre/maplibre-style-spec/issues/1049

Mostly meachanical change with the change requests from the other PR.


HarelM:
Can you share a before and after screenshot after the integration into maplibre-gl-js?

CommanderStorm:
bit hard to give exact before after since the warnings depend on what was in the viewport, but this should sufficiently show what the improvement is.

| before | after |
|--------|--------|
|  |  |

vitest.config.unit.ts
```diff
@@ -4,6 +4,8 @@ export default defineConfig({
     test: {
         name: 'unit',
         environment: 'node',
+        // Restore spies to their originals and clear call history before each test,
```
HarelM: Can you remove these comments in all 3 files?

test/integration/expression/expression.test.ts
```diff
@@ -16,12 +16,21 @@ import {
...
 } from '../../../src';
 import {ExpressionParsingError} from '../../../src/expression/parsing_error';
+import {RuntimeError} from '../../../src/expression/runtime_error';
 import {getGeometry} from '../../lib/geometry';
 import {deepEqual, stripPrecision} from '../../lib/json-diff';
 import {describe, expect, test} from 'vitest';
 
 const DECIMAL_SIGNIFICANT_FIGURES = 6;
 
+// The location of the expression within the style JSON, exactly as a real
```
HarelM: There's a lot of these redundant comments (probably added by AI), can you remove those that are not interesting or adding value?

test/integration/expression/tests/number-format/currency-and-unit/test.json
```diff
@@ -3,9 +3,18 @@
     "number-format",
     1234.567,
     {
-      "locale": ["get", "locale"],
-      "currency": ["get", "currency"],
-      "unit": ["get", "unit"]
+      "locale": [
```
HarelM: Consider reverting this change as I don't think there's any change here besides white spaces so I would like to avoid a git history for this change, if possible.

test/integration/expression/tests/legacy/exponential/composite/test.json
```diff
@@ -123,15 +123,15 @@
     },
     "outputs": [
       {
-        "error": "Expected value to be of type number, but found null instead."
+        "error": "layers[0].paint.some-property[4][2]: Expected value to be of type number, but found null instead."
```
HarelM, line 126: What does `some-property[4][2]` means?

test/integration/expression/tests/collator/variable-not-equals/test.json
HarelM, line 1: Same here for white space changes.

test/integration/expression/tests/collator/variable-lteq/test.json
HarelM, line 1: Same here for white space changes.

test/integration/expression/tests/collator/variable-gt/test.json
HarelM, line 1: Same here for white space changes.

test/integration/expression/tests/collator/variable-equals/test.json
HarelM, line 1: Same here for white space changes.

src/expression/visibility.ts
```diff
@@ -30,12 +30,18 @@ export interface VisibilityExpression {
 }
 
 class VisibilityExpressionClass implements VisibilityExpression {
+    private _rootKey: string;
```
HarelM, line 33: Would it make more sense to pass down only the layer key and have the `layout.visibility` added to it automatically? Otherwise one could pass here weird stuff and get weird stuff back...?

src/expression/index.ts
```diff
@@ -182,22 +185,63 @@ export class StyleExpression {
                 throw new RuntimeError(
                     `Expected value to be one of ${Object.keys(this._enumValues)
                         .map((v) => JSON.stringify(v))
-                        .join(', ')}, but found ${JSON.stringify(val)} instead.`
+                        .join(', ')}, but found ${JSON.stringify(val)} instead.`,
+                    ''
```
HarelM, line 188: What's this ''?

HarelM: Oh, an empty key.
This looks like a hack...

src/expression/index.ts
```diff
@@ -538,48 +585,91 @@ export function createPropertyExpression(
 export class StylePropertyFunction<T> {
     _parameters: PropertyValueSpecification<T>;
     _specification: StylePropertySpecification;
+    readonly _rootKey: string;
```
HarelM, line 587: Why were all these fields added?

HarelM:
Are you sure about the key/rootkey usage and its implementation? I find it a bit off for some reason...

vitest.config.unit.ts
```diff
@@ -4,6 +4,8 @@ export default defineConfig({
...
+        // so mock state never leaks across tests in a file.
```
CommanderStorm:
```suggestion
```

vitest.config.integration.ts
```diff
@@ -4,6 +4,8 @@ export default defineConfig({
     test: {
         name: 'integration',
         environment: 'node',
+        // Restore spies to their originals and clear call history before each test,
+        // so mock state never leaks across tests in a file.
```
CommanderStorm:
```suggestion
```

vitest.config.build.ts
```diff
@@ -4,6 +4,8 @@ export default defineConfig({
     test: {
         name: 'build',
         environment: 'node',
+        // Restore spies to their originals and clear call history before each test,
+        // so mock state never leaks across tests in a file.
```
CommanderStorm:
```suggestion
```

test/integration/expression/expression.test.ts
```diff
@@ -16,12 +16,21 @@ import {
...
+// caller (e.g. gl-js) would pass it. Every runtime error a fixture asserts is
+// prefixed with this `rootKey` plus the throwing sub-expression's index path
+// (e.g. `layers[0].paint.some-property[1]: ...`), mirroring the location
+// reporting gl-js shows at render time. `some-property` stands in for whatever
+// property the expression happens to drive.
+const ROOT_KEY = 'layers[0].paint.some-property';
```
CommanderStorm:
```suggestion
const ROOT_KEY = 'layers[0].paint.some-property';
```

test/integration/expression/tests/number-format/currency-and-unit/test.json
```diff
@@ -3,9 +3,18 @@
...
+        "locale"
+      ],
+      "currency": [
+        "get",
+        "currency"
+      ],
+      "unit": [
+        "get",
+        "unit"
+      ]
```
CommanderStorm:
```suggestion
      "locale": ["get", "locale"],
      "currency": ["get", "currency"],
      "unit": ["get", "unit"]
```

src/expression/index.ts
CommanderStorm: No, the path here should be "", since we are at the root.

src/expression/index.ts
CommanderStorm: The whole PR gives legacy stop-functions the same runtime-error handling that StyleExpression already has:
On an evaluation throw, warn once with the property's location and fall back to the spec default, instead of either throwing uncaught or console.warn-ing a bare e.message with no context.

Previously StylePropertyFunction had no evaluate method, so `createFunction()` produced a raw evaluate closure that `extendBy` copied directly onto the instance.
To wrap that raw closure in try/catch/warn/fallback logic, the class needs a real eval method plus the state that method reads.
Hence the four new fields

src/expression/visibility.ts
CommanderStorm: Would be inconsistent, lets not:

rootKey follows the same convention I set up in `featureFilter` and `createExpression`:
caller passes full location of the value in the style JSON (`layers[3].layout.visibility`), and it's only used to prefix runtime warnings, so a "weird" value just produces a weird log prefix, nothing more.
See `featureFilter`: a filter is always at `.filter`, but we still pass the whole path rather than magically/inconsistently appending it inside.

CommanderStorm:
> Consider reverting this change as I don't think there's any change here besides white spaces so I would like to avoid a git history for this change, if possible.

I just ran `npm run fmt` on those files I think.

CommanderStorm:
> Are you sure about the key/rootkey usage and its implementation? I find it a bit off for some reason...

There are not that many options to implement this really I think..
And I think the tests show it working quite well.
Yes, a lot of plumbing but I don't see how that can be avoided.

test/integration/expression/tests/legacy/exponential/composite/test.json
CommanderStorm: Lets explain this step by step:
- `layers[0].paint.some-property` is the rootKey, i.e. where the thing occurred.
- [4][2] is the index path into the compiled expression AST.
  This test is for legacy composite (zoom + property) functions.
  `convertFunction` desugars them into a nested interpolate, which means that our stlye looks like this:
  ```json
    [
      "interpolate",
      ["linear"], ["zoom"],
      0,
      [ // <- arg index [4]
       "interpolate",
       ["linear"],
       ["number", ["get","p"]], // <- arg index [2]
       0,0, 1,1],   
      1,
      ["interpolate", ["linear"], ["number", ["get","p"]], 0,2, 1,3]]

CommanderStorm: An alternative would be to not include the index in legacy functions, but that is then less context to debug on..

src/expression/expression.test.ts
```diff
@@ -127,8 +136,128 @@ describe('evaluate expression', () => {
...
+        vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+        expect(value.evaluate({} as GlobalProperties, {properties: {}} as any as Feature)).toBe(42);
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].paint.line-width: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+    });
+
+    test('an empty rootKey throws at construction', () => {
+        // There is no opt-out: a warning without a location is not actionable,
```
HarelM: This comment is probably redundant... See if you can spot other redundant AI comments and remove them...

src/expression/expression.test.ts
```diff
@@ -127,8 +136,128 @@ describe('evaluate expression', () => {
...
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].paint.line-width[2]: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].paint.line-width[3]: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+    });
+
+    test('an auto-wrapped assertion carries its nested index path', () => {
+        // Bare `["get"]` arms get implicitly wrapped in an assertion (the production
```
HarelM: Add the relevant information to the test name and remove this comment.

src/expression/expression.test.ts
CommanderStorm: I think it is valuable context WHY this is done, but sure.

src/expression/expression.test.ts
```diff
@@ -127,8 +136,128 @@ describe('evaluate expression', () => {
...
+
+        expect(value.evaluate({} as GlobalProperties, {properties: {}} as any as Feature)).toBe(42);
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].paint.line-width: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+    });
+
+    test('an empty rootKey throws at construction', () => {
+        // There is no opt-out: a warning without a location is not actionable,
+        // so a missing rootKey is a programmer error surfaced at style load.
```
CommanderStorm:
```suggestion
    test('an empty rootKey throws at construction, because not actionable', () => {
```

src/expression/expression.test.ts
```diff
@@ -127,8 +136,126 @@ describe('evaluate expression', () => {
...
+            'layers[0].paint.line-width[2]: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].paint.line-width[3]: Expected value to be of type number, but found null instead. Falling back to 42.'
+        );
+    });
+
+    test('an auto-wrapped assertion carries its nested index path', () => {
+        // Bare `["get"]` arms get implicitly wrapped in an assertion (the production
+        // case that emitted the original warning); the wrapper must carry its path.
```
CommanderStorm:
```suggestion
    test('an auto-wrapped assertion carries its nested index path, because bare `["get"]` arms get implicitly wrapped in an assertion', () => {
```

src/expression/index.ts
```diff
@@ -538,48 +585,91 @@ export function createPropertyExpression(
...
+        this._rootKey = rootKey;
+        this._defaultValue = getDefaultValue(specification);
+        this._warningHistory = {};
+        // The raw evaluate must not land on the instance, where it would shadow
+        // the error-handling `evaluate` method on the prototype.
+        const {evaluate, ...functionProperties} = createFunction(
+            this._parameters,
+            this._specification
+        );
+        extendBy(this, functionProperties);
```
HarelM: Is there a better way to solve this instead of using `extendBy`? I find this code hard to understand...
The comment above it is basically a "smell"...

src/expression/expression.test.ts
CommanderStorm: Don't like it, more complicated to read, but I can make that longer.

src/expression/index.ts
```diff
@@ -538,48 +585,91 @@ export function createPropertyExpression(
...
+        this._innerEvaluate = evaluate;
+    }
+
+    /**
```
HarelM, line 616: Make sure to solve the jsdocs warnings...

test/integration/expression/expression.test.ts
```diff
@@ -245,8 +249,15 @@ function evaluateExpression(
             }
             outputs.push(value);
         } catch (error) {
+            // Prefix with the `rootKey + index path` location the runtime warning
+            // mechanism reports, so the fixtures pin down where in the expression a
+            // throw originates (e.g. `layers[0].paint.some-property[3]: ...`). A
+            // non-RuntimeError throw has no index path, so it anchors at the root ('').
+            const path = error instanceof RuntimeError ? error.path : '';
+            const message =
```
HarelM: Can we use the same `instanceof` technique here too? It would be better to use the same mechanism if possible.

test/integration/expression/expression.test.ts
```diff
@@ -245,8 +249,15 @@ function evaluateExpression(
             }
             outputs.push(value);
         } catch (error) {
+            // Prefix with the `rootKey + index path` location the runtime warning
```
HarelM: Probably a redundant comment...

src/expression/index.ts
CommanderStorm: that change will cascade a bunch, but sure.
Your request.

src/expression/expression.test.ts
```diff
@@ -188,9 +317,200 @@ describe('evaluate expression', () => {
     });
 });
 
+describe('actionable warnings: runtime throw sites carry their index path', () => {
+    // Each throwing expression is nested at index [1] of `typeof`, which takes any
```
HarelM: Is this the comment for the method below it or the "test suite"?
If this is related to the method below please use tsdoc comment style.

To be honest, I could decipher the comment to something meaningful, so I would suggest to simply remove it.

src/expression/expression.test.ts
```diff
@@ -188,9 +317,200 @@ describe('evaluate expression', () => {
...
+    // value type (no coercion) and propagates throws. No propertySpec → null default
+    // → no fallback suffix, isolating the prefix. `get` keeps the inner expression
+    // non-constant so it isn't constant-folded into a parse error.
+    function warnFor(expression: unknown, feature: any): string {
```
HarelM: Can this be better typed somehow?

src/expression/expression.test.ts
```diff
@@ -188,9 +317,200 @@ describe('evaluate expression', () => {
...
+        assert(expr.result === 'success');
+        const warnMock = vi.spyOn(console, 'warn').mockImplementation(() => {});
+        expr.value.evaluate({} as GlobalProperties, {properties: feature} as any as Feature);
+        expect(warnMock).toHaveBeenCalledTimes(1);
+        const message = warnMock.mock.calls[0][0] as string;
+        warnMock.mockRestore();
+        return message;
+    }
+
+    test('["at"] with an out-of-bounds index', () => {
```
HarelM, line 332: Is it worth converting these tests to be like the expression validation test where there are json files with all the expressions and expected warnings?
Giving it a different file and an easy way to extend these tests...?
Just a though, I don't have strong feelings either way.

HarelM:
First review pass is complete. Try and make the rest of the changes without a "force-push" so I could review the relevant commits instead of trying to figure out what part of the file changed between the time I finished the review and the time I review it again.

src/expression/expression.test.ts
CommanderStorm:
```suggestion
describe('actionable warnings: runtime throw sites carry their index path', () => {
    function warnFor(expression: unknown, feature: any): string {
```

CommanderStorm: not really.. that is basically the type of our api..

test/integration/expression/expression.test.ts
```diff
@@ -245,8 +249,14 @@ function evaluateExpression(
...
+            // mechanism reports, so the fixtures pin down where in the expression a
+            // throw originates (e.g. `layers[0].paint.some-property[3]: ...`). A
+            // non-RuntimeError throw has no index path, so it anchors at the root ('').
```
CommanderStorm:
```suggestion
```

test/integration/expression/expression.test.ts
CommanderStorm: I think that explaining the WHY here is imprtant, but your call. Removed

src/expression/expression.test.ts
CommanderStorm: No, that would mean we route through `evaluateWithoutErrorHandling`. We specifically want the actual path used in prod here.

HarelM (approved):

Issue #1682, opened 2026-06-01
Design Proposal: Fill Extrusion Rounded Corner Fields


sjg-wdw:
# Design Proposal: Fill Extrusion Rounded Corner Fields

## Motivation

These are specific changes to support rounded corners for extruded polygons (i.e. buildings).  This is part of the [Polygon Corner Radius](https://github.com/maplibre/maplibre-style-spec/issues/1047) proposal.

We're breaking this out because the fields for 2D and 3D rounded corners are different.  We have implementations for both of the major mobile platforms and users can have access to those as soon as we've agreed upon the style change.

## Proposed Change

Add the field `fill-extrusion-rounded-corner-distance` field to the [Fill Extrusion](https://maplibre.org/maplibre-style-spec/layers/#fill-extrusion) layer type.

The units are in meters and the value defaults to 0, meaning no rounded corners.

This value is not a radius, instead it's a distance from the corner itself along each of the segments attached to that vertex.  Here is the acute case.
<img width="612" height="196" alt="Image" src="https://github.com/user-attachments/assets/1deed8fb-8719-4e89-94a4-19d984179e55" />

And here is an obtuse case.
<img width="903" height="229" alt="Image" src="https://github.com/user-attachments/assets/3022d6f9-e944-46bd-a500-3c9aa0c609e8" />

When there isn't enough room for 2x that distance along a segment, the intersection point will move closer to the corner until it collapses completely.  This handles degenerate cases reasonably well.

<img width="2048" height="1170" alt="Image" src="https://github.com/user-attachments/assets/c7021dc9-d701-4b6a-88eb-9b0b48c2709b" />

## API Modifications
The only changes are to the style spec.  The APIs will not change.

## Migration Plan and Compatibility

This is a new field and new functionality that will not be activated if the field is not present.  Corners for extruded polygons will not be rounded if the field is not there.

HarelM:
Thanks for this write up! 
A few question: 
What about the "roof" rounding? 
What about the bottom part rounding (for "mid air extrusions")?
What will happen for cases with a lot of points that the distance between them is smaller than the rounded corner distance? I've seen issues with line offset which I believe have similarities to the proposed algorithm, and it's not pretty there...
Regadless, it would be good to also bring this up in the next monthly meeting if this wasn't brought up before.

sjg-wdw:
> Thanks for this write up! A few question: What about the "roof" rounding? 

[...]

PR #1748, opened 2026-07-06
Add fill-extrusion-edge-radius layout property

Kotelberg:
## What this adds

An experimental, non-interpolated constant layout property `fill-extrusion-edge-radius` (number, meters, default `0`, minimum `0`) on the `fill-extrusion` layer. A non-zero radius rounds a building footprint's corners at bucket build time, producing a smoother silhouette and smoother facade normals. The default 0 leaves geometry byte-identical to today's output - existing styles and deployments see no change.

This PR also introduces the `experimental` property attribute to the reference schema (this is the first property to use it) and teaches the schema test to accept it. Happy to split that into a separate PR if preferred.

## Relationship to Mapbox GL JS v3

The property name and concept follow Mapbox GL JS v3's public API for ecosystem familiarity, but the semantics were designed independently: the value is expressed in meters rather than a 0–1 value, and Mapbox's `maximum: 1` cap is deliberately omitted so real-world corner radii above one meter are expressible (the renderer guards degenerate inputs - sharp turns, short edges, triangles - at the geometry level). The spec text and the implementation are original; no Mapbox code or spec text was referenced or copied.

## Implementation

A complete renderer implementation ships alongside this spec change in maplibre-native (maplibre/maplibre-native#4379), including render tests proving byte-identical output at radius 0. A maplibre-gl-js implementation of the same algorithm exists and will follow as its own PR, so the property lands with native + web parity.

Will be raising the same for maplibre-js in next few days.

## Launch Checklist

 - [x] Confirm no Mapbox backports
 - [x] Describe changes
 - [x] Related issues - (maplibre/maplibre-native#4379)
 - [x] Post benchmark scores.

### DIFF
--- a/src/reference/v8.json
+++ b/src/reference/v8.json
@@ -1145,6 +1145,23 @@
         ]
       },
       "property-type": "data-constant"
+    },
+    "fill-extrusion-edge-radius": {
+      "type": "number",
+      "experimental": true,
+      "default": 0,
+      "minimum": 0,
+      "doc": "Radius of a fill extrusion edge in meters. If not zero, rounds extrusion edges for a smoother appearance.",
+      "sdk-support": {
+        "basic functionality": {}
+      },
+      "expression": {
+        "interpolated": false,
+        "parameters": [
+          "zoom"
+        ]
+      },
+      "property-type": "constant"
     }
   },
   "layout_line": {
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -91,6 +91,7 @@ function validSchema(k, v, obj, ref, version, kind) {
         'default',
         'doc',
         'example',
+        'experimental',
         'function',
         'zoom-function',
         'property-function',
--- a/src/validate/validate_fill_extrusion_edge_radius.test.ts
+++ b/src/validate/validate_fill_extrusion_edge_radius.test.ts
@@ -36,9 +36,7 @@ describe('fill-extrusion-edge-radius validation', () => {
     });
 
     test('accepts a non-interpolated zoom camera expression', () => {
-        const errors = validateStyleMin(
-            styleWithEdgeRadius(['step', ['zoom'], 0, 14, 4])
-        );
+        const errors = validateStyleMin(styleWithEdgeRadius(['step', ['zoom'], 0, 14, 4]));
         expect(errors).toHaveLength(0);
     });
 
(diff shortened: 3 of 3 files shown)


src/reference/v8.json
```diff
@@ -1145,6 +1145,23 @@
...
       "property-type": "data-constant"
+    },
+    "fill-extrusion-edge-radius": {
+      "type": "number",
+      "experimental": true,
+      "default": 0,
+      "minimum": 0,
+      "doc": "Radius of a fill extrusion edge in meters. If not zero, rounds extrusion edges for a smoother appearance.",
+      "sdk-support": {
+        "basic functionality": {}
```
HarelM, line 1156: This should be filled with relevant issues in native and gl-js repo.

src/reference/v8.json
```diff
@@ -1145,6 +1145,23 @@
         ]
       },
       "property-type": "data-constant"
+    },
+    "fill-extrusion-edge-radius": {
+      "type": "number",
+      "experimental": true,
```
HarelM, line 1151: I'm not sure this exists...?

CommanderStorm:
@sjg-wdw I think this conflicts with your work on https://github.com/maplibre/maplibre-style-spec/issues/1682 or? Which way should we name it

sjg-wdw:
> @sjg-wdw I think this conflicts with your work on #1682 or? Which way should we name it

Bart was going to engage and discuss how they implemented their PR.  But this is a good question.

It's not just naming, this implies a slightly different implementation.

I think MapLibre would prefer our extruded building corner implementation because it's entirely clean.  We can point to how it was developed, who paid for it, and guarantee that no Mapbox code was involved.  Theirs looks to be LLM coded, so it's... murkier.  How many extruded building implementations are out there for the LLM to.... borrow?

Oh and we have Vulkan and instancing implementations.  So ours is also probably further along.

As for the style spec change, that's even trickier.  The style sheet spec is not a true open standard, so does MapLibre put itself in any danger adopting the exact same syntax as Mapbox?  If there were a standards body involved (ala Kronos or OGC) and they'd submitted a proposed changed to the spec, then it's simple.  But there isn't and they didn't.

Adopting their syntax would give me pause.

HarelM:
Ha, I missed the fact that this is a different implementation/property name.
The discussion around it is here:
- https://github.com/maplibre/maplibre-style-spec/issues/1682

It is important to start of with a design proposal to change the spec before diving into the implementation, which is what @sjg-wdw did with the linked proposal.
There's a monthly meeting today, I suggest bringing this up there and making sure we align on the name, implementation etc.

HarelM:
Closing as we don't plan to do a different style spec property for the same concept.

PR #1752, opened 2026-07-12
fix: stop mixed-filter diagnostics from blanking the whole map (#1751)

### LINKED ISSUE #1751: Unexpected errors validating styles in v25 that worked fine for a long time in maplibre gl js v5
**Describe the bug**
Some styles that have been in use for a long time are failing to validate with spec v25 with errors about mixing legacy filters and expressions. It's unclear if these expressions worked as intended in v5 or were silently misbehaving but weren't noticed by users.

**To Reproduce**
Validate a style containing a filter like this in v25 (this existed in a @stadiamaps style before they patched it):
```json
[
  "all",
  [
    "==",
    "$type",
    "LineString"
  ],
  [
    "all",
    [
      "==",
      "class",
      "rail"
    ],
    [
      "has",
      "service"
    ]
  ]
]
```

Concrete code example (for now, ohm might fix their style as stadia maps did)
```js
import { validateStyleMin, v8 } from "@maplibre/maplibre-gl-style-spec";

const resp = await fetch("https://www.openhistoricalmap.org/map-styles/main/main.json");
const style = await resp.json();

const errors = validateStyleMin(style, v8);
console.log(errors)
```

```
% node test.mjs
[
  ValidationError {
    message: 'layers[39].filter[2][1]: Mixing deprecated filter syntax with expression syntax is not supported. Replace ["in","name",""] with ["in",["get","name"],["literal",[""]]].'
  },
  ValidationError {
    message: 'layers[40].filter[2][1]: Mixing deprecated filter syntax with expression syntax is not supported. Replace ["in","surface","water"] with ["in",["get","surface"],["literal",["water"]]].'
  },
  ValidationError {
    message: 'layers[41].filter[2][1]: Mixing deprecated filter syntax with expression syntax is not supported. Replace ["in","surface","water"] with ["in",["get","surface"],["literal",["water"]]].'
  }
]
```

It can also easily be replicated using ultra, the dev version of which uses ml v6

* ml v6 - validation error - https://dev.overpass-ultra.us/#run&style=https://www.openhistoricalmap.org/map-styles/main/main.json
* ml v5 - no validation error - https://overpass-ultra.us/#run&style=https://www.openhistoricalmap.org/map-styles/main/main.json

**Expected behavior**
Styles that have worked for years shouldn't suddenly raise validation errors if they are correctly rendered.

**Additional context**

Here's the slack thread where I first reported concerns: https://osmus.slack.com/archives/C03TFH5NE83/p1781556507994529

HarelM:
## Launch Checklist

Fixes #1751. Follow-up to #1545 / #1709.

Issue 1: False positive — a pure legacy filter reported as "mixed"

```json
["all", ["==", "$type", "LineString"], ["all", ["==", "class", "rail"], ["has", "service"]]]
```

This is **100% legacy syntax — nothing mixed about it**, and it evaluated correctly before #1545. It now fails validation.

Root cause: `["has", key]` is valid *and semantically identical* in both syntaxes.

Solution: tri-state — `expression` / `legacy` / `ambiguous`. `["has", key]` and bare booleans are `ambiguous`: they are no evidence either way, so they never decide the syntax of the tree they sit in. Only an unambiguously expression-only child promotes a tree.

---

Issue 2: Severity — one bad filter takes down the entire map

Genuinely mixed filters were raised as **validation errors**. In maplibre-gl-js:

```js
// Style._load()
if (options.validate && emitValidationErrors(this, validateStyle(nextState))) {
    return;   // aborts the ENTIRE style load
}
```

`emitValidationErrors` returns `true` on *any* error and `options.validate` defaults to `true`, so `_load` returns before sources or layers are created. The result is a **blank map**, not a broken layer — for styles that have rendered for years.

Solution: warnings. The filter compiles and evaluates exactly as it did before mixing was diagnosed, and the author gets a located, actionable message instead of a dead map.

---

 - [x] Link to related issues.

Assisted by Claude Code - opus 4.8


src/error/validation_error.ts
```diff
@@ -1,20 +1,36 @@
...
 
     constructor(
         key: string,
         value: any & {
             __line__: number;
         },
         message: string,
-        identifier?: string | null
+        identifier?: string | null,
+        severity?: ValidationSeverity
```
CommanderStorm: Can the type be tighter?

```suggestion
        severity: ValidationSeverity
```

src/feature_filter/feature_filter.test.ts
```diff
@@ -815,19 +815,22 @@ describe('global-state in filter', () => {
...
-            '"$type" cannot be use with operator ">"'
+        expect(() => featureFilter(filter, 'layers[0].filter', {active: true})).not.toThrow();
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].filter[1]: "$type" cannot be use with operator ">"'
         );
     });
 
-    test('featureFilter throws explicit replacement guidance for mixed none filter', () => {
+    test('none is never an expression, so it is always converted from legacy syntax', () => {
+        vi.spyOn(console, 'warn').mockImplementation(() => {});
```
CommanderStorm, line 853: Please go over the change.
Every time you spy, please assert the number of calls and the warning message too, since just spying like here is not a good assertion ^^

src/feature_filter/feature_filter.test.ts
```diff
@@ -840,20 +843,85 @@ describe('global-state in filter', () => {
...
+        expect(console.warn).toHaveBeenCalledWith(
+            'layers[0].filter[1]: Mixing deprecated filter syntax with expression syntax is not supported. Replace ["!=","$type","LineString"] with ["!=",["geometry-type"],"LineString"].'
         );
     });
 });
+
+// Which filters are classified as mixed, and what they report, is covered by the validation
+// fixture in test/integration/style-spec/tests/filters.input.json. These tests cover what the
+// fixture cannot: what a filter actually evaluates to for a given feature.
+describe('legacy filters containing a neutral operator (#1751)', () => {
```
CommanderStorm: You would have surely objected to this comment and maybe the location of the test. 
I don't care.

CommanderStorm (approved):

src/feature_filter/feature_filter.test.ts
HarelM: Sorry, I didn't finish the review of the tests yet. My bad...

src/error/validation_error.ts
HarelM: The problem is that adding this will change a lot of tests, so I decided against it at this point to make it shorter. But I can add error for non-warning errors.

src/feature_filter/feature_filter.test.ts
HarelM: Done. reduced a lot of tests.

src/feature_filter/feature_filter.test.ts
HarelM: You are right.

src/error/validation_error.ts
HarelM: Something like the following is the expected change from making it mandatory, I don't mind either way, let me know what you think is better:
image

CommanderStorm: I am fine with a bigger diff, the severity is an interesting property it seems

HarelM: Done in 3488b01

HarelM:
@CommanderStorm let me know if you want to review the recent changes or simply release a new version.
This is the last outstanding v6 issue/PR/feature I'm aware of.
If you know of something else let me know.

CommanderStorm (approved):
If I had wanted to re-review, I would have pressed the request changes button ^^

PR #1766, opened 2026-07-14
fix(build): strip bare .json side-effect imports from generated .d.ts

CommanderStorm:
fixes CI


HarelM:
I want to include the json import with the appropriate typings, I'll see if I can manage to do that and open a different PR.
Other packages such as Maputnik, and now maplibre-gl-js can benefit from a typed v8.json file so I'll see if this can be solved. There's an exported `latest` variable which is basically v8.json imported into a variable, if this can be properly typed a lot of problems will be solved down the line.

HarelM:
- Closing in favor of #1767

PR #1768, opened 2026-07-14
fix: missing interpolation type in style function

HarelM:
## Launch Checklist

- Fixes an issue with a missed field that was copied by extendBy in #1740

This is why extendBy should not be used...

 - [x] Link to related issues.

### DIFF
--- a/src/expression/index.ts
+++ b/src/expression/index.ts
@@ -1,7 +1,6 @@
 import {ExpressionParsingError} from './parsing_error';
 import {ParsingContext} from './parsing_context';
 import {EvaluationContext} from './evaluation_context';
-
 import {
     CompoundExpression,
     isFeatureConstant,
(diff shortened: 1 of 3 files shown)


src/expression/index.ts
```diff
@@ -1,29 +1,32 @@
...
     isExpressionConstant
 } from './compound_expression';
-
 import {Step} from './definitions/step';
 import {Interpolate} from './definitions/interpolate';
 import {Coalesce} from './definitions/coalesce';
 import {Let} from './definitions/let';
 import {expressions} from './definitions';
-
+import {Color} from './types/color';
```
louwers, line 16: Shouldn't `import type` be used for types?

louwers (approved):
LGTM.

src/expression/index.ts
HarelM: It's a style spec type, but an actual class, so it needs a "full" import.

PR #1770, opened 2026-07-19
docs: document global state support for paint and layout

CommanderStorm:
partially mechanical edit, but should be fine. I reviewed that I have no bad edits in here


HarelM (approved):

louwers (changes requested):
Should be `global-state` or "global state". I would go with

```
[`global-state`](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression
```

src/reference/v8.json
```diff
@@ -948,7 +948,7 @@
           "android": "2.0.1",
           "ios": "2.0.0"
         },
-        "`global state` expression": {
+        "[`global-state`](https://maplibre.org/maplibre-style-spec/expressions/#global-state) expression expression": {
```
louwers: expression is repeated now

louwers (approved):

PR #1772, opened 2026-07-19
fix: `global-state` in `projection.type`

CommanderStorm:
Part of the series of PRs adding global state in a bunch of places.
This property was a bit under-tested it seems

## Launch Checklist

 - [x] Link to related issues.


src/expression/definitions/coercion.ts
```diff
@@ -173,8 +174,16 @@ export class Coercion implements Expression {
                 return Formatted.fromString(valueToString(this.args[0].evaluate(ctx)));
             case 'resolvedImage':
                 return ResolvedImage.fromString(valueToString(this.args[0].evaluate(ctx)));
-            case 'projectionDefinition':
-                return this.args[0].evaluate(ctx);
+            case 'projectionDefinition': {
+                const input = this.args[0].evaluate(ctx);
+                if (input === null || ProjectionDefinition.parse(input) === undefined) {
+                    throw new RuntimeError(
```
HarelM: Where are other runtime error thrown? inside the other `fromString` methods? I suggest to align with them if this is the case...

HarelM:
Added some minor comments, looks good otherwise.

src/expression/parsing_context.ts
```diff
@@ -140,7 +140,7 @@ export class ParsingContext {
                         parsed = annotate(parsed, expected, options.typeAnnotation || 'assert');
                     } else if (
                         ('projectionDefinition' === expected.kind &&
-                            ['string', 'array'].includes(actual.kind)) ||
+                            ['string', 'array', 'value'].includes(actual.kind)) ||
```
HarelM, line 143: What does `value` stand for in this case?

CommanderStorm: the type of global state is "value", since it is not statically deternimable.
Thus the runtime error.

HarelM: Yeah, I thought this might be it. Thanks for the info.

src/expression/definitions/coercion.ts
CommanderStorm: No, because those accept ANY valid string.
We accept an enum, projection type array or expression, so `ProjectionDefinition.fromString(valueToString(..))` would be plainly inaccurate and wrong.

This is the only case where we have an runtime poly-type.

I can change the if-else to not be returning the error, which goes aginst your preference but is closer to how `numberArray` handles errors.
Really not much that I can do

HarelM (approved):

PR #1794, opened 2026-07-29
Allow negative fill-extrusion-base and fill-extrusion-height

clement-igonet:
Ref maplibre/maplibre-gl-js#8051.

Drops `minimum: 0` from `fill-extrusion-base` and `fill-extrusion-height` so floors below ground level can be expressed (e.g. underground station levels; Gare Saint-Lazare goes down to level -6).

Both docs now state that negative values extrude below ground; `base <= height` still applies.
Height needs negatives too so a floor that is entirely underground (base -18 m, top -15 m) can be expressed.

Precedents:
- terrain already pushes geometry below the surface, so renderers handle sub-ground geometry;
- `symbol-elevation` is being added without a minimum, and its doc says negative values lower the symbol, so labels can go underground while the floors they belong to currently cannot.

Working renderer proof (maplibre-gl-js with only its shader ground-clamp removed): https://maplibre.confinia.io/debug/underground-8051.html - a Saint-Lazare-like stack with levels +2..-3. I will open the matching maplibre-gl-js PR (shader change + render test) once this is agreed.

Assisted-By: Claude Fable 5 (claude-fable-5)


clement-igonet:
Parking this for now to keep the review queue light; will reopen together with the maplibre-gl-js rendering side once that is ready to review as a pair.

clement-igonet:
The rendering side is ready: maplibre/maplibre-gl-js#8069. Un-parking this so the two can be reviewed as a pair.

HarelM:
Thanks for taking the time to open this PR.
CC: @louwers @sjg-wdw

@clement-igonet I think it would be wise to share this proposal change in the monthly meeting.

louwers (changes requested):
Before we can merge this, the changed minimum should be documented in the `sdk-support` table so that users know what versions of MapLibre support this behavior. This will also require linking issues (or I guess pull requests) for MapLibre GL JS and MapLibre Native.

From my side a discussion during the monthly meeting is not really needed, since no properties are added, and the use cases are very clear.

clement-igonet:
Added the sdk-support entries for negative values on both properties, linking the GL JS tracking issue (maplibre/maplibre-gl-js#8051, implementation in maplibre/maplibre-gl-js#8069) and a new tracking issue on Native (maplibre/maplibre-native#4455). The validator only accepts issue links, so the issue is linked rather than the PR. Thanks for the quick review and for clearing the meeting question.

clement-igonet:
@louwers the `sdk-support` entries landed in be7b2a03, with links for GL JS (maplibre/maplibre-gl-js#8051) and Native (maplibre/maplibre-native#4455).

The rendering side is green and waiting on this one: maplibre/maplibre-gl-js#8069.
Without the change here a literal negative value fails style validation, so the two only work as a pair.

louwers (approved):
Perfect. Thank you.

HarelM (approved):

HarelM:
I'm not sure what should be the correct implementation, but from a spec perspective, I don't see why these value shouldn't be allowed.

PR #1798, opened 2026-08-01
feat: intial `{fill,line}-layer-blend` support

### LINKED ISSUE #1661: `{line,fill,circle,..}-layer-blend` for layer compositing modes
> [!IMPORTANT]
> I am **NOT** currently **championing to implement this proposal** myself.
> This issue exists to document and scope the design space.
> It needs a dedicated champion/implementor interested in driving the proposal forward to make progress.

# Design Proposal: `{line,fill,circle,..}-layer-blend` for layer compositing modes

## Motivation

MapLibre currently composites layers using a fixed blending model `normal`.

While existing paint properties control color and opacity, there is currently no way to control how a layer blends with previously rendered content.

Modern graphics systems commonly expose compositing or blending controls such as `normal`, `multiply`, `screen`, `overlay`, `darken`, `lighten`, `plus`, `erase`, ...

These are useful for:

* cartographic emphasis
* hillshade and terrain rendering
* heatmap and glow effects
* annotation overlays
* artistic/stylized maps
* improved visual integration between raster/vector layers

Currently, these effects require:

* preprocessing assets externally
* custom rendering pipelines
* additional render passes
* engine-specific extensions

A standardized layer-level blending mechanism would allow these effects to be expressed directly in the style specification.

| `layer-blend` | without blending | with blending |
|--------|--------|--------|
| multiply | Image | Image |
| colour burn | Image | Image |
| (multiple layers, see post) | Image | Image |

| | |
|--------|--------|
|  |  |

All images from https://www.helenmakesmaps.com/post/how-to-use-blending-modes-in-mapping

## Requirements

The style specification should support controlling how a fully composited layer blends with existing framebuffer contents.

`*-layer-blend` is evaluated after all features in the layer have been composited, and therefore applies to the layer output as a whole rather than to individual features.

Therefore, the compositing model is:

- features are rendered normally into the layer output
existing *-opacity properties continue to apply during feature compositing
- `*-layer-opacity`, if present, is applied to the composited layer output before blending
- `*-layer-blend` defines how the final layer output is composited into the destination framebuffer

## Proposed Change

Add new paint properties `{line,fill,circle}-layer-blend`

These properties control how the composited output of the layer blends with existing framebuffer contents.

Initial supported values could include `normal`, `multiply`, `screen`, `overlay`, `darken`, `lighten`, `plus`, `erase`, ...

> [!NOTE]
> The **exact supported set** would be what the **champion for this issue would need to clarify** where each of the values is useful

The properties would:

* be data-constant
* support transitions
* support zoom expressions

Unlike existing paint properties, these properties affect framebuffer compositing behaviour rather than feature styling.

This proposal is fully backwards compatible.

## Rejected Alternatives

Alternative approaches discussed include:

* render-pass systems
* custom shader hooks
* engine-specific extensions

These approaches are intentionally out of scope for this proposal.

CommanderStorm:

Resolves #1661


docs/assets/fill-layer-blend.svg
CommanderStorm, line 1: @justb4 FYI

justb4: Good to see progress! Will need to experiment. `hsv-value` in [GDAL Blend](https://gdal.org/en/stable/programs/gdal_raster_blend.html) looks nice, but don't know how difficult that is for MapLibre. 

(For me and @louwers : things may come together soon with: 
https://github.com/mapterhorn/mapterhorn/issues/247 !)

CommanderStorm: Also CC @jo-chemla @abelvm @wipfli if something is missing that SHOULD be there.

wipfli:
Exciting. Multiply blending would be particularly well suited for contour lines on top of hillshading. Can you maybe share how this example would look like with multiply line blending?

https://mapterhorn.com/examples/contour/

wipfli:
Also forests and glaciers on top of hillshading will probably look much better with multiply blending. I wrote a bit about it here: https://oliverwipfli.ch/canvas-multiply-blending-2025-11-24/

docs/assets/fill-layer-blend.svg
CommanderStorm: hsv-value (the way you want it) requires `hillshade-layer-blend`, which is not hard to add, but a different property a bit 😉
The reason is that hillshade in gdal is applied on top, not below

CommanderStorm: actually no, w3c has this as luminosity and color (+- different color space), so lets just add these two.

justb4: Think GDAL is general indeed, maybe the examples above show "on top". 

For what it is worth/inspiration: in [map5topo](https://map5topo.nl) raster (pre-rendered-)hillshade blending, Landcover layers are both below and on top the hillshade layer (0.5 and 5m reso)  with different [Mapnik blending modes](https://mapnik.org/mapnik-reference/#4.2.0/comp-op) (called `comp-op`) after experimentation: 

* `grain-merge` Hillshade over Landcover
* `soft-light` Landcover over hillshade

Diving too far here. Blending is also far wider than hillshade fill layers.  I suggest to start with a small set of modes like you specified, and expand from user-feedback and hopefully PRs..

gravitystorm:
As I discussed at State of the Map, I'd love to see this implemented, and I'm happy to do whatever I can to help!

As some background, I've used Mapnik for many years which has a [rich list of comp-ops](https://mapnik.org/mapnik-reference/#4.2.0/comp-op). These can be applied both per-layer and also per-feature, but I appreciate we're only talking about per-layer composition in this PR. I use a subset of these operators in various styles for various reasons (including `hard-light`, `multiply`, `darken`, `soft-light`, `dst-out` and `src-atop`).

My initial thoughts are that the list of blend modes suggested here could be aligned with the W3C "[Compositing and Blending Level 1](https://www.w3.org/TR/compositing-1/)" specification. This would mean that the available operators - and their names - would then align with CSS (`mix-blend-mode`), canvas (`globalCompositeOperation`), SVG, and various other graphics pipelines. When it comes to implementation, it would make sense for MapLibre to also use the maths given in the W3C spec.

Two additional things:
* "plus" is the name used in porter-duff, and also Mapnik, but it's "lighter" in the W3C level 1 spec and "plus-lighter" in the SVG used above (a clamped version of lighter used in compositing level 2) - and also not to be confused with "lighten". I think "plus" could be renamed to "lighter" to reduce confusion.
* "erase" isn't used as a term by any of the above, probably better as "destination-out". It also needs some indication as to what the resultant backdrop would be (e.g. style background-colour, or underlying canvas colour, or something else).

PR #1807, opened 2026-08-03
Revert "chore: update js sdk support for fill-extrusion-rounded-corners"

louwers:
Approved the PR, didn't see it was set to auto-merge.


CommanderStorm (approved):
Don't think it is nessary to do this, is fine as is imo

HarelM:
Let's avoid this, I'll release a new version in a few days anyway...

louwers:
OK.

PR #1817, opened 2026-08-14
feat: make url optional for image sources

mondsichtung:
### Make `url` optional for image sources
`ImageSource.updateImage` in maplibre-gl-js accepts an already-decoded image (`{image: ImageBitmap | ...}`) with no network involved (maplibre/maplibre-gl-js#7944). But an image source cannot currently be *created* without a `url`: the spec requires it, so apps that only ever feed the source through `updateImage` must supply a placeholder 1x1 data-URI url just to pass validation.  

This makes `url` optional on `source_image`. A source created without one starts empty and renders nothing until content arrives via `updateImage({image})` or `updateImage({url})`. `coordinates` remains required, video sources keep `urls` required.  

Companion gl-js PR: [maplibre/maplibre-gl-js#8167](https://github.com/maplibre/maplibre-gl-js/pull/8167) (skips the fetch for a url-less source and omits `url` from `serialize()`); it should land after this and a dependency bump.

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Thanks for taking the time to open this PR! Can you clarify the use case of this change. From a style validation perspective it makes sense to make sure the url is not empty. The fact that you can update an image through the API doesn't mean that from a style authoring perspective this makes sense.

mondsichtung:
Thank you for looking into this!  

The main use case is an overlay whose image doesn't exist as a fetchable file at all.  
In my case the app generates it at runtime through WebAssembly calculations of astronomical data.

Currently the only way to do this is to give the source a fake 1x1 data-URI url just to pass validation.  
Letting the url be omitted feels more honest than that placeholder.

```js
// current workaround: a minimal placeholder that gets fetched, decoded and briefly rendered
map.addSource('overlay', {
    type: 'image',
    url: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=',
    coordinates
});

// with this change
map.addSource('overlay', {type: 'image', coordinates});

// either way, the real content arrives here
map.getSource('overlay').updateImage({image: bitmap});
```

This wouldn't be the first source that starts empty and gets its data from the API: a geojson source can begin as an empty `FeatureCollection` and be filled later with [`setData`](https://maplibre.org/maplibre-gl-js/docs/API/classes/GeoJSONSource/#setdata), which is exactly how the [Measure distances example](https://maplibre.org/maplibre-gl-js/docs/examples/measure-distances/) works.

MapLibre Native already works this way:
- On iOS [`MLNImageSource.URL` is `nullable`](https://github.com/maplibre/maplibre-native/blob/550f64be2232e09934ffc660a9acdbacff8b164a/platform/darwin/src/MLNImageSource.h#L88) and according to its docs a source initialized from or updated with an image has a `nil` URL
- On Android [`ImageSource.setUri` accepts a nullable `String?`](https://github.com/maplibre/maplibre-native/blob/550f64be2232e09934ffc660a9acdbacff8b164a/platform/android/MapLibreAndroid/src/main/java/org/maplibre/android/style/sources/ImageSource.kt#L176), so a source can have no url there too

So the spec would just change the JSON with the shape the native runtimes already use.

HarelM:
> In my case the app generates it at runtime through WebAssembly calculations of astronomical data

If this is the case, what's preventing you from adding the relevant source and layer programmatically too?

It's not that I object too much to this spec change, don't get me wrong, but I'm trying to better understand the motivation around it.
Aligning to the native implementation is a fair point. @louwers what do you think?

mondsichtung:
Not sure if I understood you correctly, please correct me if I got you wrong.  

> If this is the case, what's preventing you from adding the relevant source and layer programmatically too?

That's exactly what I'm doing. The source and layer are already added through the API, the snippet above is that path: `map.addSource('overlay', {...})` followed by `map.addLayer` at runtime.

But gl-js validates added sources against the spec. [`Map.addSource` forwards to `Style.addSource` with no options](https://github.com/maplibre/maplibre-gl-js/blob/1dc7f02a098b9a1d6999082815e1613e5498a6d4/src/ui/map.ts#L2870-L2872), and [`Style.addSource` runs every source type the spec has a schema for through `validateStyle.source`](https://github.com/maplibre/maplibre-gl-js/blob/1dc7f02a098b9a1d6999082815e1613e5498a6d4/src/style/style.ts#L1013-L1014), image sources included. Without a url that fails the spec's required-field check and the source is simply not added:

```js
map.addSource('overlay', {type: 'image', coordinates});
// Error: sources.overlay: missing required property "url"
```

Therefore my app can't avoid the placeholder workaround.  
The requirement lives in the spec, which is why the fix has to start here rather than in gl-js.

louwers (dismissed):
Sounds good to make it optional. But the spec should not include references to platform specific APIs.

Please create an issue in MapLibre Native when this is merged "Support image sources with empty URL" because right now it's not supported.

https://github.com/maplibre/maplibre-native/issues/new

mondsichtung:
> Sounds good to make it optional. But the spec should not include references to platform specific APIs.
> 
> Please create an issue in MapLibre Native when this is merged "Support image sources with empty URL" because right now it's not supported.
> 
> https://github.com/maplibre/maplibre-native/issues/new

Changed in https://github.com/maplibre/maplibre-style-spec/pull/1817/changes/0b0e6b68ccacd3fa97a76946383adc84d234903c

I would open the issue after this is merged.

HarelM:
I'm not sure if the optional part should be documented in a sdk like table, but let's merge this as is and see if this needs to be fixed in the future...

HarelM (approved):

PR #1829, opened 2026-08-18
Publish a module-per-file ESM build so bundlers can tree-shake

ahocevar:
`dist/index.mjs` is a single module. `"sideEffects": false` in `package.json` therefore has not really anything to act on for bundlers that eliminate dead code at *module* granularity.

For example, an app that only needs one small helper:

```js
import {derefLayers} from '@maplibre/maplibre-gl-style-spec';
```

got the whole package back, minified (26.2.1):

- rollup: 65,021 bytes
- webpack 5: 129,379 bytes

rollup does better because it also does statement-level elimination inside a module, whereas webpack only can remove dead code at the module level. But even rollup cannot drop everything, because the module has top-level side effects (`CompoundExpression.register`).

This pull request suggests adding an ESM output (built with `preserveModules`) to `dist/esm/`, and point `package.json`'s `module` property at it.

With this change, with the same import, bundlers can do a much better job:

```js
import {derefLayers} from '@maplibre/maplibre-gl-style-spec';
```

now resolves through `dist/esm` instead of `dist/index.mjs`, minified:

- rollup: 65,021 → **477 bytes**
- webpack 5: 129,379 → **407 bytes**

webpack now matches rollup — the fix isn't webpack-specific, it helps any bundler once an app uses less than the full surface.

To benefit from treeshaking even more, 2b41151a47f8f4ab0919554ee9963ed52c51d61b replaces an `instanceof` chain in the constant analysis with two optional `Expression` flags, so `compound_expression.ts` no longer needs to hard-references `Within`, `Distance`, `CollatorExpression` and `GlobalState` just for the `instanceof` check. This adjustment is worth ~14.7 kB minified.

## AI disclosure

The build configuration change, the refactor and the test additions were drafted with AI assistance (Claude). 

## Launch Checklist

 - [x] Post benchmark scores.

### DIFF
--- a/test/build/style-spec.test.ts
+++ b/test/build/style-spec.test.ts
@@ -6,7 +6,7 @@ import {spawnSync} from 'node:child_process';
 import {tmpdir} from 'node:os';
 import {join} from 'node:path';
 import {describe, test, expect} from 'vitest';
-const minBundle = fs.readFileSync('dist/index.mjs', 'utf8');
+const refBundle = fs.readFileSync('dist/reference/v8.mjs', 'utf8');
 const validStyle = JSON.stringify({version: 8, sources: {}, layers: []});
 
 describe('@maplibre/maplibre-gl-style-spec npm package', () => {
@@ -30,7 +30,26 @@ describe('@maplibre/maplibre-gl-style-spec npm package', () => {
         expect(dirContents).toContain('index.mjs');
         expect(dirContents).toContain('index.mjs.map');
         expect(dirContents).toContain('latest.json');
-        expect(dirContents).toHaveLength(18);
+        expect(dirContents).toHaveLength(43);
+    });
+
+    test('index.mjs is a module-per-file build, not a single bundle', async () => {
+        // The point of building with `preserveModules` is that consumers can
+        // drop what they do not import, which only works if the modules that
+        // `dist/index.mjs` imports stay separate files alongside it.
+        const dirContents = await readdir('dist');
+        expect(dirContents).toContain('deref.mjs');
+        expect(dirContents).toContain('expression');
+        expect(fs.statSync('dist/index.mjs').size).toBeLessThan(5000);
+    });
+
+    test('cjs build exports the same names as the esm build', async () => {
+        const cjs = await import('../../dist/index.cjs');
+        const esm = await import('../../dist/index.mjs');
+        // The UMD build adds a synthetic `default` export pointing at itself,
+        // which the esm build has no equivalent for.
+        const cjsNames = Object.keys(cjs).filter((name) => name !== 'default');
+        expect(Object.keys(esm).sort()).toEqual(cjsNames.sort());
     });
 
     test('exports components directly, not behind `default` - https://github.com/mapbox/mapbox-gl-js/issues/6601', async () => {
@@ -39,7 +58,7 @@ describe('@maplibre/maplibre-gl-style-spec npm package', () => {
 
     test('trims reference.json fields', () => {
         expect(latest.$root.version.doc).toBeTruthy();
-        expect(minBundle.includes(latest.$root.version.doc)).toBeFalsy();
+        expect(refBundle.includes(latest.$root.version.doc)).toBeFalsy();
     });
 
     test('"latest" entry point should be defined', async () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - _...Add new stuff here..._
+- Publish the ESM build (`dist/index.mjs`, referenced by `package.json#module`) as one module per source file instead of a single bundle, so that bundlers which eliminate dead code at module granularity can drop unused parts of the package ([#1829](https://github.com/maplibre/maplibre-style-spec/pull/1829)) (by [@ahocevar](https://github.com/ahocevar))
 
 ### 🐞 Bug fixes
 - _...Add new stuff here..._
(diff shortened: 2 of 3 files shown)


HarelM:
Why is another exported target needed? Why not improve the existing one?

ahocevar:
Thanks for the quick feedback! If you're open to such a change, I'll happily do that. It means dropping `./dist/index.mjs` in favor of `./dist/esm/*`. Are you ok with that?

HarelM:
Why not keep it? What's the point in moving it under a folder?

ahocevar:
@HarelM Changed as you suggested. The one-module-per-file `dist/index.mjs` now replaces the previous `dist/index.mjs`.

test/build/style-spec.test.ts
```diff
@@ -30,7 +30,26 @@ describe('@maplibre/maplibre-gl-style-spec npm package', () => {
         expect(dirContents).toContain('index.mjs');
         expect(dirContents).toContain('index.mjs.map');
         expect(dirContents).toContain('latest.json');
-        expect(dirContents).toHaveLength(18);
+        expect(dirContents).toHaveLength(43);
```
CommanderStorm, line 33: does this mean that if an user downloads from the (f.ex.) jsdeliver CDN that we add another roundtrip?

src/expression/expression.ts
```diff
@@ -14,6 +14,16 @@ export interface Expression {
...
+    /**
+     * Set on expressions that may produce a different result on every evaluation, even
+     * with constant arguments, and so must never be folded into a constant.
+     */
+    readonly neverConstant?: boolean;
+    /**
+     * Set on expressions that read from the feature being evaluated, and so are never
+     * feature-constant regardless of their arguments.
+     */
+    readonly featureDependent?: boolean;
```
CommanderStorm: does this actually have an effect? If yes, could you explain a bit more clearly (in your words) how this change produces more tree-shakable code? Seems strange to me..

ahocevar: The direct effect is not here, but in the `compound_expression` module. The `Distance`, `Within` and `GlobalState` classes were imported in `compound_expression` only for an `instanceof` check. After the chain of changes that includes the new `neverConstant` and `featureDependent` properties, which are now used for entering different code paths in `compoud_expressions` instead of the previous `instanceof` checks, `compound_expression` no longer depends on `Distance`, `Within` and `GlobalState`. Since these classes have many imports, this does have significant impact on what can be tree-shaked when not all of this library is used by an application.

HarelM: Would be nice to remove all the `instanceof` from this method and use some class/interface properties instead...

test/build/style-spec.test.ts
ahocevar: Not for jselivr.net (neither old nor new `dist/index.mjs` would work in a browser directly), but yes for ESM CDNs like esm.sh (assuming that the old `dist/index.mjs` wouild have worked at all with esm.js — which I have not tested).

CommanderStorm: see https://maplibre.org/maplibre-gl-js/docs/#__tabbed_1_6

src/expression/expression.ts
ahocevar: I thought about giving each expression class a `name` property or something, but that would be a bigger change and would not solve the inheritance chain problem without additional changes.

test/build/style-spec.test.ts
ahocevar: Ok, module imports. That won't work with the new `dist/index.mjs` out-of-the-box. I'm not sure if there's a use case for this. maplibre-gl-js includes this package already as dependency.

But you're right, this is a breaking change and warrants a major release. However, many npm packages have been making the same kind of change in the last couple of years ("ES-only package").

HarelM:
@ahocevar I'm still confused a bit about this change. 
Can you explain why tree shaking doesn't work currently and needs this change? 
As far as I know, bundlers, for example rollup/down, are smart enough to pack only the relevant part of the library, is this not the case now?

ahocevar:
@HarelM The improvements I wanted to make with this pull request are two-fold.

First: webpack only tree-shakes at module granularity, so with a single-file `dist/index.mjs` it keeps everything. rollup does better (statement-level elimination) but still can't drop the operator registration code, since it runs as a top-level side effect. After splitting into one module per file, when importing e.g. `derefLayers` only, the import size goes down from 65,021 to 477 bytes (rollup), or 129,379 to 407 bytes (webpack 5).

Second — and this is something I had not thought through to the end: removing the `Within`, `Distance`, `CollatorExpression` and `GlobalState` dependencies from `src/expression/compound_expression.ts‎` does not give us an immediate gain, because these classes are still imported directly by the expressions registry in `definitions/index.ts`.

That said: thanks for keeping asking, I just removed the second commit.

HarelM:
I've published the following branch:
https://github.com/maplibre/maplibre-style-spec/compare/main...improve-export

Seems like only rollup is able to really tree-shake well with this change, but it doesn't affect rolldown and webpack, which is a bit disappointing... 

I'll continue to think about it as I feel this change is a bit too dramatic for this use case.
I wanted to delay the release of this package to include this, but I need more time to look into it, sorry...

ahocevar:
@HarelM I have removed the commit with the change to expressions, because the registry cannot be torn apart in a sensible way.

What really matters is the module-per-file exports, instead of a bundle for esm. Which is what's left in this pull request.

HarelM (approved):

HarelM:
This was released, let me know if it helped...

ahocevar:
Thanks a lot, @HarelM — it did help! Application bundles that import from ol-mapbox-style and use webpack as bundler are now ~21 kB smaller, which is almost 5%.

PR #1876, opened 2026-09-14
chore: improve types for diff command.

HarelM:
Improves types for diff command to allow better typings in maplibre-gl-js

## Launch Checklist


src/diff.ts
```diff
@@ -156,12 +175,41 @@ function diffSources(
...
+ * Adds a paint or layout property command for a single property of a layer.
+ * The property names are taken from a layer's paint/layout object, so they are known to be
+ * valid names for the given command, which a `for ... in` loop over that object can't express.
+ * @param after - The paint or layout object to take the property value from
+ * @param commands - The commands array to add to
+ * @param layerId - The id of the layer the property belongs to
+ * @param prop - The name of the property that changed
+ * @param command - The command to add
+ */
+function addPropertyCommand(
```
CommanderStorm, line 188: the typing here does not look ideal, but 🤷🏻

CommanderStorm (approved):

src/diff.ts
HarelM: Yeah, I'm not sure I have a better solution beside splitting this into several methods, but I'm not sure it would be a lot better... Let me know if you think I should follow this direction.

birkskyum (approved):

src/diff.ts
CommanderStorm: I don't really care, both are fine.

CommanderStorm: I think I am fairly transparent if there is something I care about ^^

PR #1878, opened 2026-09-14
Fix semiliteral parsing in array contexts

sargunv:
## Description

Fix `semiliteral` parsing when the surrounding property or expression specifies an expected type. The parser incorrectly applies that type to each element, which can cause a TypeError.

For example, `["semiliteral", [1, 2]]` should produce a valid `text-offset`, but each number is checked as though it must be an array. This also affects `text-variable-anchor-offset` and arrays passed to `at`.

Parse the elements independently, then check the completed array against the expected type. Invalid elements now produce a parse error at the offending element instead of crashing.

Found while adding semiliteral support in https://github.com/maplibre/maplibre-compose/pull/1414.

## Validation

Six regression tests fail before the fix and pass after it.

## Launch Checklist

- [x] No Mapbox backports.
- [x] Regression tests added.
- [x] Changelog updated.

## AI assistance

OpenAI Codex (GPT-6) assisted with the investigation, implementation, tests, and PR description.


src/expression/expression.test.ts
HarelM, line 1: I think most successful test are using json, don't they? It's more "cross platform" I believe. Do these unit tests add to coverage more than the json files?

src/expression/definitions/semiliteral.ts
```diff
@@ -41,7 +41,13 @@ export class Semiliteral implements Expression {
 
         if (type.kind === 'array') {
             const arr = value as Array<unknown>;
-            const parsed = arr.map((item) => context.parse(item, null, ValueType));
+            const arrayContext = context.concat(1);
```
HarelM, line 44: What does the concat 1 here means?

sargunv: It creates a parsing context for the array at index 1

```
  ["semiliteral", [element0, element1]]
                  ↑ index 1
```

src/expression/expression.test.ts
sargunv: done

HarelM (approved):

src/expression/definitions/semiliteral.ts
HarelM: Not sure I fully understand, but if the tests pass that's good enough.

HarelM:
I want to incorporate another fix before releasing a new version, so I'll release in a few hours, hopefully.

